PR #3, opened 2018-05-15
Unwrap serialisation code, take 2

alexdewar:
Hi Jamie,

I've incorporated your suggestions and it's looking much prettier now. Thanks!

Any other thoughts?

Alex


common/opencv_unwrap_360.h
```diff
@@ -106,11 +108,11 @@ class OpenCVUnwrap360
         cv::remap(input, output, m_UnwrapMapX, m_UnwrapMapY, cv::INTER_NEAREST);
     }
 
-    void writeFile()
+    /*
+     * Serialise this object.
+     */
+    void operator>>(cv::FileStorage &fs)
```
neworderofjamie: It's been a while since I did this but, I **think** this is one of the gross operator overloading cases where you actually have declare these as free functions for the operators to work as expected:
https://msdn.microsoft.com/en-us/library/1z2f6c2k.aspx
https://msdn.microsoft.com/en-us/library/x6aebccc.aspx

So in this case you'd add a friend for each operator to the class:
```c++
friend cv::FileStorage& operator << (cv::FileStorage &fs, const OpenCVUnwrap360 &unwrap);
friend cv::FileStorage& operator >> (cv::FileStorage &fs, OpenCVUnwrap360 &unwrap);
```
And then declare the operators at the bottom of the header file:

```c++
inline cv::FileStorage& operator << (cv::FileStorage &fs, const OpenCVUnwrap360 &unwrap)
{
    ...
}

inline cv::FileStorage& operator >> (cv::FileStorage &fs, OpenCVUnwrap360 &unwrap)
{
    ...
}
```

Yey C++ :/

alexdewar: That is pretty gross :-p. My tests do seem to work on gcc but I'm guessing MSVC will be less forgiving on this... Although the SO post says that it's so that the >> operator function can get at the object's private members, which isn't an issue here because all the members are public (this may or may not be good design but I do need to be able to get at these internals from my unwrap config program and didn't want to write a ton of setters and getters).

alexdewar: So I just tested this in VS 2017 and it seems to work ok. I'll merge this into master for now but we can revisit all this once I've got the unwrap config program working, when we can do another PR.

PR #8, opened 2018-05-26
Robot net api

alexdewar:
Hi @neworderofjamie , hope you are having/have had a good break!

Don't feel you have to look at this any time soon; it's not urgent. It's also pretty massive and we might want to have a chat about it first (I'm back on Weds). I'm mostly just sending this before I go for the satisfaction of having it (somewhat) finished :-). It's also not been tested on an actual robot yet.

Almost all of the code is the new networking API. I've put example server (robot) and client (computer) programs in robot_net_test. Essentially I was thinking of having an API that looks somewhat as it does in the examples and worked backwards from there -- the internal structure of some of these structures I'm sure can be improved though.

There are various limitations. For one, I'm currently doing the unwrapping on the client (computer) side, which probably isn't what we want. We'd need to write a little helper class (to turn an `OpenCVUnwrap360` into a `Video::Input`) before we could do it on the robot side. It might also be good to have the option to just send single frames over the network (e.g. currently-matching snapshot) rather than just streams.


net/socket.h
```diff
@@ -0,0 +1,259 @@
...
+    /*
+     * Get the last error message.
+     */
+    static std::string errorMessage()
+    {
+        return " (" + std::to_string(errno) + ": " + std::strerror(errno) + ")";
+    }
+}; // Socket
+} // Net
+} // GeNNRobotics
```
neworderofjamie: Thinking some more - I don't think a lot of this stuff belongs in socket - it should be moved up into ``Node``. When using the commands etc, the actual sockets themselves could then probably be hidden from the user entirely (which is good as you could easily fuck up the command stuff e.g. by reading too much) - all that they would then be presented with are your read and send functions etc.  I think that would also give you more scope for optimising the command system to reduce the number of runtime memory allocations if it proved necessary.

bebop/joystick.cc
```diff
@@ -1,73 +1,66 @@
...
 namespace GeNNRobotics {
 namespace Robots {
 BebopJoystick::BebopJoystick(Bebop *bebop)
   : m_Bebop(bebop)
 {
-    if (!m_Joystick.open()) {
-        throw std::runtime_error("Could not find joystick");
-    }
-    m_Joystick.startThread(EventCallback, this);
+    setCallback([=](HID::Event &js) { eventCallback(js); });
```
neworderofjamie: I think it's nice to be more specific in your capture expression i.e. the [] bit - in this case you only need to capture ``[this]``

video/input.h
```diff
@@ -70,26 +74,69 @@ class Input
...
 
-    virtual bool readFrame(cv::Mat &outFrame)
+    virtual cv::Size getOutputSize() const
     {
-        return false;
+        return cv::Size();
     }
 
-    virtual cv::Size getOutputSize() const
+    virtual bool needsUnwrapping()
```
neworderofjamie: I feel this method should be const

video/input.h
```diff
@@ -70,26 +74,69 @@ class Input
...
+
+    virtual void setOutputSize(const cv::Size &outSize)
+    {}
+
 protected:
     /*
      * Default constructor is protected to stop users directly creating objects
      * of this class.
      */
     Input()
```
neworderofjamie: You can also do this explicitly with ``Input() = delete;`` in C++11.

video/input.h
```diff
@@ -70,26 +74,69 @@ class Input
...
     {}
+
+private:
+    std::unique_ptr<std::thread> m_ImageThread;
```
neworderofjamie: You don't need to make pointers to ``std::thread``. Just default construct it (i.e. to an invalid thread) and then assign it i.e. ``m_ImageThread = std::thread([this](Net::Node *n){ runImageSink(n); }, &node));``

net/node.h
```diff
@@ -0,0 +1,104 @@
...
+class Node : public Threadable
+{
+public:
+    virtual Socket *getSocket() const = 0;
+
+    /*
+     * Add a handler for a specified type of command (e.g. if it's an IMG command,
+     * it should be handled by Video::NetSource).
+     */
+    void addHandler(Handler &handler)
```
neworderofjamie: The only limitation I see with this is that it only allows a single object to handle one type of event. An alternative would be to just define a handler type as ``typedef std::function<void(Command&)> CommandFn `` (or whatever the syntax is) and **that** is what is passed to ``addHandler(const std::string &command, CommandFn fn)`` rather than a handler object. Then handlers could just call ``addHandler`` to hook their member functions to specific commands. Not 100% sure what you'd do about ``onConnected`` though

alexdewar:
Thanks for all the helpful comments! Much appreciated.

PS -- I'm working from home today in case you or Andy want to chat about anything... But I'm busy with family tomorrow.

bebop/joystick.h
```diff
@@ -1,29 +1,33 @@
...
+#include "../hid/joystick.h"
+
+// local includes
 #include "bebop.h"
-#include "../joystick/joystick.h"
 
 namespace GeNNRobotics {
 namespace Robots {
-using ButtonEvent = bool (*)(Joystick::Event *js);
+using ButtonEvent = std::function<bool(HID::Event &js)>;
```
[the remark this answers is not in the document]
alexdewar: Yeah, turns out you can't use lambdas with function pointers for one thing....

hid/joystick_base.h
```diff
@@ -0,0 +1,174 @@
...
+        return read(m_JsEvent);
+    }
+
+    void run() override
+    {
+        if (!m_Callback) {
+            throw std::runtime_error("Callback function for joystick not set");
+        }
+
+        while (read()) {
```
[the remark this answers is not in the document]
alexdewar: This is a really good idea.

There are currently a bunch of race conditions where the program exits and I wasn't entirely sure what to do about it (I had an elaborate scheme with `std::shared_ptr`s in mind), but this is cleaner and simpler.

hid/joystick_windows.h
```diff
@@ -122,50 +127,50 @@ Joystick::read(Event &js)
                 int rThumbXState2 = Read().Gamepad.sThumbRX;
```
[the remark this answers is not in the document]
alexdewar: Re point 2: I think it is actually correct -- the code is trying to work out which buttons' states have changed since the last poll -- but will add comments.

I agree with the other points and will also give this file a bit of a tidy-up.

net/client.h
```diff
@@ -0,0 +1,81 @@
...
+    Socket *getSocket() const override;
+};
+
+/*
+ * Create client, connect to host over TCP
+ */
+Client::Client(const std::string host, int port)
+{
+    // Needed for Windows
+    WSAStartup();
```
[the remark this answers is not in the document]
alexdewar: Yeah, you're right. I stuck `WSAStartup()`/`WSACleanup()` in `Server`/`Client` ctors and dtors because they are only used once per process atm, but this is a lazy assumption.

bebop/joystick.cc
alexdewar: Agreed

video/input.h
alexdewar: Yep

video/input.h
alexdewar: :+1:

net/node.h
alexdewar: It's true that this would be a cleaner interface. OTOH all the places where we're handling `Net::Command`s will be encapsulated in classes anyway, so we don't necessarily need the flexibility. But it would be a bit nicer and I probably already have gone overboard with the multiple inheritance in places... We could have a separate set of event handlers for `onConnected` (à la Java).

neworderofjamie: Removing the ``getHandledCommandName()`` method and passing the name to ``addHandler`` would probably also be a simpler way to achieve the same flexibility.

PR #9, opened 2018-05-31
Path library

neworderofjamie:
I needed some more path processing stuff for the next bit of snapshot bot so, rather than add some reinvented wheels to filesystem.h, I dropped in the path libary we've been using for SpineML_To_GeNN. It aims to be dependency-free, cross platform and resemble ``std::filesystem`` so it can be removed painlessly when that becomes widely available. Additionally:

- Pretty sure, because extensions were transformed to lower case, the file extension matching in unwrapfile.cc would have failed.
- Pretty sure the path joining with "/" in input.cc would fail on Windows - the / operator in the filesystem library solves this.


alexdewar:
Looks good to me. This was on my todo list too.

PR #11, opened 2018-05-31
Grayscale

neworderofjamie:
This is hopefully the only extra bit of functionality required to implement your comments in https://github.com/BrainsOnBoard/snapshot_bot/pull/1. Also tidied up a few of the demos to use ``Video::getPanoramicCamera()``

### DIFF
--- a/camera_360_test/camera_360_test.cc
+++ b/camera_360_test/camera_360_test.cc
@@ -8,26 +8,21 @@
 // Common includes
 #include "../common/timer.h"
 #include "../imgproc/opencv_unwrap_360.h"
+#include "../video/panoramic.h"
 
 using namespace GeNNRobotics;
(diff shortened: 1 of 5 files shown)


alexdewar:
This all looks good, except I think the implementation of `readGrayscaleFrame()` in `OpenCVInput` should be in `Video::Input` instead, because this is how we'll want to implement it for all devices other than the Seecam (e.g. `Video::NetSource` in my PR #8).

All `Video::Input` devices should be implementing their own version of `readFrame()` so it ought to be ok. (I probably should have made it into a pure function to enforce this.)

neworderofjamie:
@alexdewar do you think you could check over my changes to this and #13 if you get a chance and merge them as well?

alexdewar:
Sure. Sorry, didn't realise you were waiting on me for this.

PR #13, opened 2018-05-31
Unwrap serialization

neworderofjamie:
In order to support serialising it as part of some larger settings yaml I needed to update the way the ``OpenCVUnwrap360`` class was serialized so it players nicer with the (somewhat gross) OpenCV serialisation system.  However this means you don't seem to be able to serialise objects into the root of a YAML document which meant updating the existing yaml files etc, but now means you can do stuff like this:

```c++
cv::FileStorage settingsFile("experiment_settings.yaml", cv::FileStorage::WRITE);
settingsFile << "localUnwrapper" << m_Unwrapper;
settingsFile << "fov" << m_FOV;
```

### DIFF
--- a/imgproc/opencv_unwrap_360.h
+++ b/imgproc/opencv_unwrap_360.h
@@ -7,11 +7,11 @@
 // OpenCV includes
 #include <opencv2/opencv.hpp>
 
+//----------------------------------------------------------------------------
+// GeNNRobotics::ImgProc::OpenCVUnwrap360
+//----------------------------------------------------------------------------
 namespace GeNNRobotics {
 namespace ImgProc {
-//----------------------------------------------------------------------------
-// OpenCVUnwrap360
-//----------------------------------------------------------------------------
 class OpenCVUnwrap360
 {
 public:
@@ -109,8 +109,9 @@ public:
     /*
      * Serialise this object.
      */
-    OpenCVUnwrap360 &operator>>(cv::FileStorage &fs)
+    void write(cv::FileStorage& fs) const
     {
+        fs << "{";
         // centre
         cv::Point2d centre = {
             (double) m_CentrePixel.x / (double) m_CameraResolution.width,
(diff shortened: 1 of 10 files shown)


alexdewar:
This looks much nicer; def a good thing to have it properly integrated with the existing serialisation code.

My only nit is that I don't think we should have the "{" and "}" bits inside the serialisation code -- you could leave that to the user and I think it would be more readable as it would then be in the same form as the other YAML maps that we'd be making:
```
cv::FileStorage settingsFile("experiment_settings.yaml", cv::FileStorage::WRITE);
settingsFile << "mySpecialParameters" << "{" << "height" << 10 << "sponginess" << 2 << "name" << "Steven" << "}";
settingsFile << "localUnwrapper" << "{" << m_Unwrapper << "}";
```
etc.

And then we could also leave the unwrap param files as is.

neworderofjamie:
I fear there is an extra layer of magic between you trying to serialise;
and the ``read`` and ``write`` function that asserts if you try and write
to the root of a document but I will try and figure out what the {}s
actually do tomorrow!

alexdewar:
Oh really? That's a bit weird.

I just understood "{" and "}" to mean "begin map" and "end map". If it's a faff to read/write at the root level, we could just keep the new unwrap param files as they are and then it won't be a problem (as we won't be doing this anywhere else). I think it would just be nice to be able to see at a glance whether you're writing a single entry (i.e. key/value pair) or a set of entries (i.e. for a map).

neworderofjamie:
I tried removing the { and } and it doesn't change the serialised output or allow you to write unnamed objects to file - it gives the following error:

>  No element name has been given in function operator<<

I'd vote to leave the {}s in because that's what the (scant) documentation suggests for serializing your own classes.

alexdewar:
Well if that's what the documentation suggests then let's do it that way.

Cheers for looking into it!

PR #14, opened 2018-06-06
Robot API updates v2

alexdewar:
Hi! Here's the new version of the API. The biggest change has been to turn it into an interface that can be polled.

There's a minor change to the `Input::readFrame()` API: now it returns true if there's a new frame available, available, false otherwise and throws an exception if something goes wrong. This is so that if we're reading images over the network we don't have to block waiting for a new image.

I've completely rewritten the joystick code to have an event-based API (with separate events for buttons and axes) and also a pollable API that works pretty much identically to @neworderofjamie 's with a few minor differences.


net/client.h
```diff
@@ -0,0 +1,77 @@
...
+#include "socket.h"
+
+namespace GeNNRobotics {
+namespace Net {
+class Client
+  : public Node
+  , Socket
+{
+public:
+    Client(const std::string host, int port = DefaultListenPort);
```
neworderofjamie: I feel this should be a reference rather than a copy?

net/server.h
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+#ifndef _WIN32
+    if (setsockopt(m_ListenSocket, SOL_SOCKET, SO_REUSEADDR, &on, sizeof(on)) <
+        0) {
+        goto error;
+    }
+#endif
+
+    memset(&addr, '0', sizeof(addr));
```
neworderofjamie: I'm pretty sure you don't want to do this - this filles the struct with the ASCII  code for 0 :smile:

net/socket.h
```diff
@@ -0,0 +1,259 @@
...
+    void setSocket(socket_t sock)
+    {
+        std::lock_guard<std::mutex> guard(m_ReadMutex);
+
+        m_Socket = sock;
+        checkSocket();
+    }
+
+private:
+    char m_Buffer[DefaultBufferSize];
```
neworderofjamie: Windows particularly has a very small stack size and fails in a singularly difficult to diagnose way when it overflows so so I think it would be better to use a `std::vector ` and resize if to DefaultBufferSize in the contructor

net/socket.h
```diff
@@ -0,0 +1,259 @@
...
+     */
+    size_t readOnce(char *buffer, size_t start, size_t maxlen)
+    {
+#ifdef _MSC_VER
+        int len = recv(m_Socket, (readbuff_t) &buffer[start], (bufflen_t) maxlen, 0);
+#else
+        int len;
+        while ((len = ::read(m_Socket, (readbuff_t) &buffer[start], (bufflen_t) maxlen)) == 0) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        }
```
neworderofjamie: How come you're using different socket functions on Windows and *NIX?

os/keycodes.h
```diff
@@ -1,9 +1,15 @@
...
 
+// C++ includes
+#include <limits>
+
 namespace GeNNRobotics {
 namespace OS {
-namespace KeyCodes {
+
 #ifdef _WIN32
+static const int KeyMask = std::numeric_limits<int>::max();
```
neworderofjamie: Shouldn't key codes be unsigned?

video/input.h
```diff
@@ -71,7 +85,12 @@ class Input
 
     virtual const std::string getCameraName() const
     {
-        return "unknown_camera";
+        return DefaultCameraName;
+    }
+
+    virtual cv::Size getOutputSize() const
```
neworderofjamie: Why not just make this pure virtual?

hid/joystick_base.h
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+    }
+
+    void updateAxis(JAxis axis, int16_t value, bool isInitial)
+    {
+        updateAxis(axis, axisToFloat(axis, value), isInitial);
+    }
+
+    void updateAxis(JAxis axisHorz, int16_t hvalue, int16_t vvalue, bool isInitial)
```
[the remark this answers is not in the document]
alexdewar: Possibly. But I think it is quite convenient to have it in there. It means that users don't have to save the axis values between polls, for example.

hid/joystick_linux.h
```diff
@@ -0,0 +1,193 @@
...
+    RB = 5,
+    Back = 6,
+    Start = 7,
+    Xbox = 8,
+    LeftStick = 9,
+    RightStick = 10,
+    LENGTH
+};
+
+class Joystick : public JoystickBase<JButton>
```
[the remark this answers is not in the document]
alexdewar: They aren't more generic, but on Windows the axes don't have numbers associated with them; you access them through e.g. `state.Gamepad.sThumbLY` etc. So I just used the axis numbers that Linux uses for both.

But I agree that there's something conceptually nicer about having both `JButton` and `JAxis` templated, so I'm open to that.

net/client.h
alexdewar: Good spot!

net/server.h
alexdewar: Haha. Oops.

net/socket.h
alexdewar: Good shout

net/socket.h
alexdewar: Erm, mostly because I tacked the Windows code on at the end and there's no `read()` function in Windows. But it looks like there is a `recv()` function in Linux and you're right that it would be nicer to just use one.

os/keycodes.h
alexdewar: Hmm I dunno. Intuitively yes, but the value you get from `waitKeyEx()` is a signed integer (so that it can return -1 in the case of no key pressed). This way does seem to work though.

video/input.h
alexdewar: Yeah that would be cleaner.

PR #15, opened 2018-06-07
Renaming

alexdewar:
This is just some small housekeeping changes. I've renamed the `createDefaultUnwrapper()` function to `createUnwrapper()` and I've renamed all the classes for the wheeled robots to drop the Motor* prefix, which didn't seem particularly helpful (apart from anything, we now have a `GeNNRobotics::Robots` namespace which makes the names less ambiguous). I've also renamed the header files and updated all the examples.

I renamed `Robots::MotorI2C` to `Robots::Norbot` as this seemed a bit clearer to me than `Robots::I2C` or `Robots::Arduino` etc. But it is a bit silly, so I'm happy to change to any other sensible alternative name.

I've also just reworked the examples in GeNN robotics to use the new joystick API in a separate PR. If you're happy with both of these PRs then I'll update the snapshot_bot code accordingly (and add in the display stuff) before merging them in.

### DIFF
--- a/camera_360_test/camera_360_test.cc
+++ b/camera_360_test/camera_360_test.cc
@@ -21,7 +21,7 @@ int main()
 
     // Create panoramic camera and suitable unwrapper
     auto cam = getPanoramicCamera();
-    auto unwrapper = cam->createDefaultUnwrapper(unwrapRes);
+    auto unwrapper = cam->createUnwrapper(unwrapRes);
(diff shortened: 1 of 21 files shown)


alexdewar:
Also part of the rationale for this is that I want to tidy up the Bebop code and put that in the robots folder too. Then we can have a directory structure like:
- robots/norbot.h
- robots/bebop.h
etc.

PR #18, opened 2018-06-07
Tweaks

neworderofjamie:
Mostly separating the network stuff from Input so you can a)send things other than camera frames and b)can control the rate that frames are sent at more carefully


neworderofjamie:
Is that better @alexdewar?

PR #21, opened 2018-06-12
extra i2c header required on my ubuntu 18 install

neworderofjamie:
This was required for the I2C code to build (after installing libi2c-dev) on my new Ubuntu 18 install (kernel 4.15ish) - does this work on your various machines (especially arch!)? If it does, don't merge - I need to test on jetsons n'stuff!

### DIFF
--- a/common/i2c_interface.h
+++ b/common/i2c_interface.h
@@ -15,6 +15,7 @@
 
 // I2C includes
 #include <linux/i2c-dev.h>
+#include <i2c/smbus.h>
 
 namespace GeNNRobotics {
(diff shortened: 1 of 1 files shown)


alexdewar:
Yep, this works for me too.

The i2c code has been broken for me on Arch for a while, so I've just been excluding that header with #ifdefs in my code.

alexdewar:
Yep

neworderofjamie:
Irritatingly this now **doesn't** compile on Ubuntu 16 - will need to figure out some #ifdef magic

PR #23, opened 2018-06-21
Libantworld work

neworderofjamie:
At this point this is something of an intermediate step towards integrating the 'antworld' 3D renderer more tightly with bob_robotics:
- Moved libantworld classes into ``BoBRobotics::AntWorld`` namespace
- Textured obj loading - this essentially lets you load in obj + mtl + textured images (typically these will need to be triangulated and decimated using, for example, Blender) as well as the binary files I created from Barbara Webb's matlab project 
- Less hard-coded rendering layout - you can now position the top down and ant-view wherever you want/size them etc
- Simple test application where you can use a HID to wander around the world
- Very thin ``BoBRobotics::Video::Input`` wrapper for reading pixels rendered by OpenGL - do you think this is necessary/useful?

As I see it the next steps (probably for another P.R.) is to create a ``BoBRobotics::Robots::Positionable`` (name?) interface which will encompass controlling robots which can be sent to a position + rotation:
- The gantry
- A tank + Vicon tracking + PID control
- The simulated world

As ever your thoughts would be much appreciated! I'll give you a demo with the pier model when you're back in the office as the model is hundreds of mb!


ant_world/ant_world.cc
```diff
@@ -241,12 +246,12 @@ int main(int argc, char *argv[])
             antHeading += Parameters::antTurnSpeed;
         }
         if(keybits.test(KeyUp)) {
-            antX += Parameters::antMoveSpeed * sin(antHeading * degreesToRadians);
-            antY += Parameters::antMoveSpeed * cos(antHeading * degreesToRadians);
+            antX += Parameters::antMoveSpeed * sin(antHeading * AntWorld::degreesToRadians);
+            antY += Parameters::antMoveSpeed * cos(antHeading * AntWorld::degreesToRadians);
```
alexdewar: We could also do this with the new units library at some point

ant_world_test/ant_world_test.cc
```diff
@@ -63,22 +74,31 @@ int main()
...
+    glEnable(GL_TEXTURE_2D);
 
-    HID::Joystick joystick;
+    // Create renderer - increasing cubemap size to improve quality in larger window
+    AntWorld::Renderer renderer(512);
+    renderer.getWorld().load("../libantworld/world5000_gray.bin",
+                             {0.0f, 1.0f, 0.0f}, {0.898f, 0.718f, 0.353f});
+
+    // Load world, keeping texture sizes below 4096 and compressing textures on upload
+    renderer.getWorld().loadObj("/home/j/jk/jk421/Documents/pier/pier_alex_smoothed_decimated_triangles.obj",
```
alexdewar: I feel like this shouldn't be a hardcoded path. You could just point it to `whatever.obj` in the current folder and then people can copy or symlink it in.

alexdewar: Come to think of it we could also put the file in the repo and use Git LFS, if it's not *too* huge (which I suspect it may be). You're not obliged to do a `git lfs install` on your repo so people could just download the data files if they want them.

obj_tools/obj_process.vcxproj
```diff
@@ -0,0 +1,84 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
alexdewar: Could also add this to BoBRobotics.sln

alexdewar: I mean the project file, not this one line of XML :-p

alexdewar:
PS -- I also agree that having a `Positionable` class is the way to go with the gantry, ant world etc.

ant_world_test/ant_world_test.cc
neworderofjamie: Yeah - I'll remove this hardcoded path. As the data in question was a sample from the drone surveying company we probably don't have the rights to publish it on our github though!

PR #24, opened 2018-06-21
Make dependencies

### LINKED ISSUE #19: Makefile dependencies
I agree that using makefiles over shellscripts is a good thing but currently they don't handle dependencies properly. I have got this working before (see libantworld) and it's pretty cool - I'll try and apply it to all makefiles when I feel that way inclined.

neworderofjamie:
This probably wins least exciting PR ever but hey - it fixes #19 :) This dependency stuff makes Make way less of a horrible build system to work with. The {} syntax you were using in some of the make clean directives didn't seem to work for me so I made it dumber

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@
 **/*.o
 **/*.obj
 
+# Ignore dependency files
+**/*.d
+
 # Ignore output folders from code generation
 **/*_CODE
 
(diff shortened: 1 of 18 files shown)


alexdewar:
This all looks good to me. And I'm excited by it at least ;-)

The only thing is in a couple of places in makefiles I use `-lstdc++` because you need that to compile on clang (which I use sometimes). But don't bother changing it back; that's just for future reference.

neworderofjamie:
Ahh shit I forgot about clang!

PR #25, opened 2018-06-22
Tweaks

neworderofjamie:
These changes are generally aiming to make applications that stream not block waiting for a client (I feel in many cases this is what you want). As an example I also made the PI demo stream what's happening in it's mighty brain


stone_cx/simulatorRobotDeadReckon.cc
```diff
@@ -55,8 +59,23 @@ void imuThreadFunc(std::atomic<bool> &shouldQuit, std::atomic<float> &heading, u
...
 {
+    // Simulation rendering parameters
+    const unsigned int activityImageWidth = 500;
+    const unsigned int activityImageHeight = 1000;
     const float velocityScale = 1.0f / 10.0f;
+
+    const bool streamActivity = (argc > 1);
+
+    // Create server and sink for sending activity image over network
+    Net::Server server(Net::Socket::DefaultListenPort);
```
alexdewar: You don't need to specify the default port here, but it obviously doesn't matter.

stone_cx/simulatorRobotDeadReckon.cc
```diff
@@ -4,11 +4,14 @@
 #include <fstream>
 #include <thread>
 
-// Common includes
-#include "../hid/joystick.h"
+// GeNN robotics includes
```
alexdewar: I guess it's BoB robotics now...

alexdewar:
Looks good to me.

PR #26, opened 2018-06-22
Third party library for units

alexdewar:
This is the library we were talking about, which is MIT-licensed so we can just bung it in wholesale. It has lots of cute features like compile-time unit checking etc. (It also has a constant for pi defined :+1:.)

There are a number of obvious places it could go in. E.g. for the Vicon system we could change the translation values to be `units::length::meter_t` and the rotation angles could be some kind of `units::angle`. This would def make sense for the future `Positionable` class you mentioned in #23. I haven't implemented any of this yet though because I don't want to break things.

I've also changed makefiles to build with `-std=c++14` and stuck a few random c++14-isms in (e.g. using `std::make_unique`).

And finally I've stuck the matplotlib stuff into third_party/ where it belongs.


neworderofjamie:
This looks excellent!

alexdewar:
Yeah it seems pretty nifty :-D

PR #29, opened 2018-07-20
Ant world tweaks

alexdewar:
Here are some small changes to the ant db code (e.g. output image size, file names, FOV etc.). Hopefully nothing too controversial!


alexdewar:
Ta!

PR #30, opened 2018-07-20
New interface for getting and setting pose

alexdewar:
I started off changing the ant world code to use physical unit types, with the aim of standardising things so we can have a ``Positionable`` class etc. and moved on to incorporating this into the Vicon code as well.

I've made a super-basic ``AntAgent`` class, which is just a thin wrapper for the existing ant world code, but gives an interface where you can get and set the agent's position and read frames.

I decided to use templating for the length and angle unit types, which means that e.g. you can read the coordinates from the Vicon system in mm and degrees if you prefer. I've used metres and radians as defaults though.


ant_world_db_creator/ant_world_db_creator.cc
```diff
@@ -233,13 +228,13 @@ int main(int argc, char *argv[])
...
             // If we've reached the X edge of the world, move to start of next Y
-            if((worldMin[0] + ((float)currentGridX * gridSpacingM)) > worldMax[0]) {
+            if(worldMin[0] + currentGridX * gridSpacing > worldMax[0]) {
                 currentGridY++;
                 currentGridX = 0;
             }
 
             // If we've reached the Y edge of the world, stop
-            if((worldMin[1] + ((float)currentGridY * gridSpacingM)) > worldMax[1]) {
+            if(worldMin[1] + currentGridY * gridSpacing > worldMax[1]) {
```
neworderofjamie: Why? I quite like making precedence explicit...

ant_world_test/ant_world_test.cc
```diff
@@ -136,25 +146,25 @@ int main()
...
-
         // Use right trigger to control forward movement speed
-        const float forwardMove = moveSpeed * deltaTime * joystick.getState(HID::JAxis::RightTrigger);
+        const meter_t forwardMove = moveSpeed * deltaTime * joystick.getState(HID::JAxis::RightTrigger);
 
         // Calculate movement delta in 3D space
-        x += forwardMove * sin(yaw * AntWorld::degreesToRadians) * cos(pitch * AntWorld::degreesToRadians);
-        y += forwardMove * cos(yaw * AntWorld::degreesToRadians) * cos(pitch * AntWorld::degreesToRadians);
-        z -= forwardMove * sin(pitch * AntWorld::degreesToRadians);
+        auto cosPitch = cos(pitch);
```
neworderofjamie: How does this know that cos expects radians?

common/pose.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
```
neworderofjamie: I think could be kinda overkill - not really sure what sort of agent wouldn't have all 4...

libantworld/agent.h
```diff
@@ -0,0 +1,105 @@
...
+private:
+    Vector<AngleUnit> m_Attitude{ 0_rad, 0_rad, 0_rad };
+    Vector<LengthUnit> m_Position{ 0_m, 0_m, 0_m };
+    Renderer &m_Renderer;
+    GLFWwindow *m_Window;
+}; // AntAgent
+} // AntWorld
+} // BoBRobotics
+
+namespace BoBRobotics {
```
neworderofjamie: Especially with templates, I find this inlining syntax really cumbersome and verbose. Is there any reason not to have the methods inline as in where the class is declared? I always felt that was one of the side-benefits of header-only :)

libantworld/agent.h
```diff
@@ -0,0 +1,105 @@
...
+
+// BoB robotics includes
+#include "../common/pose.h"
+#include "../video/opengl.h"
+
+using namespace BoBRobotics::Pose;
+
+namespace BoBRobotics {
+namespace AntWorld {
+template<class LengthUnit = meter_t, class AngleUnit = radian_t>
```
neworderofjamie: I have yet to really look into the units stuff, but is it possible to template the set/get methods on the units passed to each function instead? That would mean that ``AntAgent`` itself can use use whatever units it wants internally and users of it can mix and match to match their mood? Having to have **every** class that uses units templated seems both annoying and a little restrictive...

libantworld/renderer.cc
```diff
@@ -249,13 +249,13 @@ void Renderer::generateCubeFaceLookAtMatrices()
...
-void Renderer::applyFrame(float x, float y, float z,
-                          float yaw, float pitch, float roll)
+void Renderer::applyFrame(meter_t x, meter_t y, meter_t z,
+                          degree_t yaw, degree_t pitch, degree_t roll)
 {
-    glRotatef(roll, 0.0f, 1.0f, 0.0f);
-    glRotatef(pitch, 1.0f, 0.0f, 0.0f);
-    glRotatef(yaw, 0.0f, 0.0f, 1.0f);
-    glTranslatef(-x, -y, -z);
+    glRotated(roll.value(), 0.0, 1.0, 0.0);
```
neworderofjamie: In this case it's really not important as it's only done once a frame, but as this probably ends up being done on the GPU, using double here is probably not a great plan.

libantworld/route_continuous.cc
```diff
@@ -7,28 +7,32 @@
...
 {
-    return (x * x);
+    return hypot(v2[1] - v1[1], v2[0] - v1[0]);
 }
-//----------------------------------------------------------------------------
-float distanceSquared(float x1, float y1, float x2, float y2)
+
+template<class T>
+meter_t
+distance2(const T &v1, meter_t x2, meter_t y2)
```
neworderofjamie: I feel the clarity of ``distanceSquared`` was worth the few characters - ``distance2`` makes me think it's version 2 somehow.

libantworld/route_continuous.cc
```diff
@@ -225,13 +227,13 @@ bool RouteContinuous::load(const std::string &filename)
...
 //----------------------------------------------------------------------------
-void RouteContinuous::render(float antX, float antY, float antHeading) const
+void RouteContinuous::render(meter_t antX, meter_t antY, degree_t antHeading) const
 {
     // Bind route VAO
     glBindVertexArray(m_WaypointsVAO);
 
     glPushMatrix();
-    glTranslatef(0.0f, 0.0f, 0.1f);
+    glTranslated(0.0, 0.0, 0.1);
```
neworderofjamie: Especially here - why?! Doubles and GPUs don't mix!

vicon/udp.h
```diff
@@ -25,109 +24,145 @@
...
-        std::copy(std::begin(translation), std::end(translation), std::begin(m_Translation));
-        std::copy(std::begin(rotation), std::end(rotation), std::begin(m_Rotation));
+        m_Position[0] = x; m_Position[1] = y; m_Position[2] = z;
+        m_Attitude[0] = yaw; m_Attitude[1] = pitch; m_Attitude[2] = roll;
     }
 
     uint32_t getFrameNumber() const { return m_FrameNumber; }
-    const Vector &getTranslation() const{ return m_Translation; }
-    const Vector &getRotation() const{ return m_Rotation; }
+    Vector<LengthUnit> &getPosition() override { return m_Position; }
```
neworderofjamie: Once again, it may not be possible, but couldn't these functions be templated rather than ``ObjectData``? That way the Vicon code can use whatever units the system uses and only convert (or not) when you read stuff out.

vicon/udp.h
```diff
@@ -282,16 +314,16 @@ class UDPClient
...
+                    // Read object position
+                    Vector<millimeter_t> position;
+                    memcpy(&position[0], &buffer[itemOffset + 27], 3 * sizeof(double));
 
-                    // Read object rotation
-                    double rotation[3];
-                    memcpy(&rotation[0], &buffer[itemOffset + 51], 3 * sizeof(double));
+                    // Read object attitude
+                    Vector<radian_t> attitude;
+                    memcpy(&attitude[0], &buffer[itemOffset + 51], 3 * sizeof(double));
```
neworderofjamie: The reliance on ``millimeter_t`` etc using double as it's underlying store AND ``sizeof(millimeter_t) == sizeof(double)`` worries me a little here - I feel these variables could be passed to ``updateObjectData`` as raw doubles (the units are always the same after all) and then, in ``ObjectData`` they can be converted for the user.

common/opencv.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+/*
+ * This header is needed, because I'm getting compile errors when trying to include
+ * OpenCV because the KEY_UP and KEY_DOWN macros are already defined (on Ubuntu 18.04).
+ * 	- AD
+ */
```
neworderofjamie: YUCK :/ Any clue where?

video/opengl.h
```diff
@@ -57,9 +57,9 @@ class OpenGL : public Input
         return false;
     }
 
-private:
+protected:
```
neworderofjamie: Maybe add public getters rather than doing this? Protected member baaaad, public getters better (or something)

vicon/udp.h
```diff
@@ -228,14 +259,15 @@ class UDPClient
...
-        }
+        /*
+         * Update object data with position and attitude.
+         * 
+         * Note that we reorder the rotation angles we get from the Vicon system
+         * so that they are in the order of yaw, pitch and roll (which seems to
+         * be standard).
+         */
+        m_ObjectData[id].update(frameNumber, position[0], position[1], position[2],
+                                attitude[2], attitude[0], attitude[1]);
```
neworderofjamie: Did you come round to the slightly weird templated object class system rather than the callbacks then?

ant_world_db_creator/ant_world_db_creator.cc
alexdewar: It's more that I wanted to remove the cast to ``float``: I'll put the brackets back in.

ant_world_test/ant_world_test.cc
alexdewar: The units lib has wrapper functions for (I think) all of the cmath functions, so you get a version of ``cos`` which takes a ``radian_t`` parameter and if you pass it a ``degree_t`` object it'll be implicitly cast and converted to radians. It's a pretty cool library!

common/pose.h
alexdewar: Yeah I agree. I have been thinking about this and what you said below about using templates for ``AntAgent`` and ``ObjectData`` etc. and I'm thinking that a better strategy would be to make the "getter" functions templates rather than having these templated superclasses. Then we just use duck-typing for our agent classes.

libantworld/agent.h
alexdewar: Yeah you're right :+1:

libantworld/agent.h
alexdewar: Yes we could do that and I think we probably should. Good shout.

vicon/udp.h
alexdewar: This makes sense, especially as the Vicon system gives the rotation angles in a strange order.

ant_world_test/ant_world_test.cc
neworderofjamie: Damn that's cool!

common/opencv.h
alexdewar: No :-/. I might install an IDE on my Ubuntu VM and try to track it down though. I don't really like this..

video/opengl.h
alexdewar: That probably would be clearer...

vicon/udp.h
alexdewar: Yuh huh. I'm all about templating now as you may have noticed ;-)

neworderofjamie: My work here is done :smile:

common/opencv.h
neworderofjamie: From a quick search on my Ubuntu 18 install it looks like /usr/include/linux/input-event-codes.h is a likely candidate :/

neworderofjamie: Google suggests that adding -H to the g++ commandline might list how it gets included

libantworld/agent.h
```diff
@@ -18,88 +18,65 @@ using namespace BoBRobotics::Pose;
...
+    {
+        m_Attitude[0] = yaw;
+        m_Attitude[1] = pitch;
+        m_Attitude[2] = roll;
+    }
+
+    bool readFrame(cv::Mat &frame)
+    {
+        // Clear colour and depth buffer
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
```
neworderofjamie: I get the desire to do this but it does make the whole thing super inflexible - if you look at how antworld (which should be renamed) uses this stuff you wouldn't be able to use this class.

vicon/udp.h
```diff
@@ -39,98 +39,100 @@ namespace BoBRobotics {
...
     //----------------------------------------------------------------------------
-    void update(uint32_t frameNumber, LengthUnit x, LengthUnit y, LengthUnit z,
-                AngleUnit yaw, AngleUnit pitch, AngleUnit roll)
+    void update(uint32_t frameNumber, millimeter_t x, millimeter_t y, millimeter_t z,
+                radian_t yaw, radian_t pitch, radian_t roll)
     {
-        const Vector<LengthUnit> position {x, y, z};
-        constexpr second_t frameS = 10_ms;
-        constexpr second_t smoothingS = 30_ms;
+        const Triple<millimeter_t> position {x, y, z};
```
neworderofjamie: Why are you turning this into a tuple? All you do with it is extract it's x, y and z a few lines down...

common/pose.h
```diff
@@ -12,48 +15,43 @@ namespace Pose {
...
-/* 
- * A class for objects which can return their attitude (rotation).
- */
-template<class AngleUnit>
-class HasAttitude
-{
-public:
-    virtual Vector<AngleUnit> &getAttitude() = 0;
-};
+using Triple = std::tuple<T, T, T>;
```
neworderofjamie: Tuples **are** cool but, because they're immutable, they are also a pain in the ass - why not just use the array-based one throughout?

vicon/udp.h
```diff
@@ -184,6 +187,27 @@ class UDPClient
...
+        else {
+            throw std::runtime_error("Invalid object id: " + std::to_string(id));
+        }
+    }
+
+private:
+    //----------------------------------------------------------------------------
+    // Private API
+    //----------------------------------------------------------------------------
     bool connect(unsigned int port)
```
neworderofjamie: Why did you make this private? It makes the default constructor somewhat useless by meaning that you can't just create it and only connect if some runtime parameter is set (which I feel I do in the snapshot_bot). I get the horrible feeling that _I_ suggested it - if so it was dumb and I apologise :blush:

common/pose.h
alexdewar: Tuples are a pain in the ass, but the problem with using arrays is that we want templatable getters (so we can choose to get position as e.g. metres) and if we can't return an array declared on the stack. I suppose we could do:
``
template<class LengthUnit>
void getPosition(Array3<LengthUnit> &position);
``
But that would mean that users of the function would have to declare their own arrays before calling ``getPosition``.

Another possibility (for the future) is that we use [Eigen](http://eigen.tuxfamily.org)'s template vector classes (though I'm not certain that the templating in the units lib will necessarily place nicely with Eigen).

vicon/udp.h
alexdewar: Oops! I was the one being dumb. I forgot that ``connect`` wouldn't be called in the default constructor.

common/pose.h
neworderofjamie: How about ``std::array<LengthUnit>``?

neworderofjamie: e.g. something like this (pretty sure you can instantiate an array like this):

```c++
template<class OutputUnit, class InputType>
inline std::array<OutputUnit, 3> makeUnitTriple(const std::array<InputType, 3> &values)
{
    return {static_cast<OutputUnit>(values[0]), 
                static_cast<OutputUnit>(values[1]), 
                static_cast<OutputUnit>(values[2])};
}
```

alexdewar: Hmm this seems to work. It just seems not to work if you use C-style arrays :-/.

Ok well this is def preferable!

neworderofjamie: Not being able to return C-style arrays does seem really stupid - if you want to fill your stack with array by returning them from functions why shouldn't you be able to!

vicon/udp.h
```diff
@@ -126,13 +126,13 @@ class ObjectDataVelocity : public ObjectData
...
         const auto oldPosition = getPosition<>();
-        Array3<meters_per_second_t> instVelocity;
+        Vector3<meters_per_second_t> instVelocity;
         const auto calcVelocity = [deltaS](auto curr, auto prev) {
             return (curr - prev) / deltaS;
         };
-        instVelocity[0] = calcVelocity(getX(position), getX(oldPosition));
-        instVelocity[1] = calcVelocity(getY(position), getZ(oldPosition));
-        instVelocity[2] = calcVelocity(getZ(position), getZ(oldPosition));
+        std::transform(std::begin(position), std::end(position),
```
neworderofjamie: YEAH! That's the other advantage of std::array!

libantworld/agent.h
alexdewar: I agree that it does make the class inflexible, but I'm not sure what the best approach is here. I have just moved the rendering stuff out of that function, but in retrospect I think it should probably stay in. Essentially what I was trying to do here is make something a bit like our current (somewhat crappy) Matlab scripts for achieving the same thing, where you can set an agent's position and get its current panoramic view.

Then we could write functions like this silly example:
``
template<class T>
void getViewFromOrigin(T &agent, cv::Mat &frame)
{
    agent.setPosition(0_m, 0_m, 0_m);
    agent.readFrame(frame);
}
``

The problem is that if we need to do extra things in between setting the position and reading the view then it won't be easy to write e.g. tests of algorithms where we want all that to be abstracted away. In fact, I think we should probably wrap *all* the OpenGL stuff we do for ant_world_db_creator into ``AntAgent``, so that people like me who don't understand OpenGL don't have to worry about it :-).

I agree that it would be nice to have a generic way to access the cool visualisations that you have for ant_world and stone_cx, but we don't have to try to implement that now. It totally makes sense to have libantworld as a lower-level API that gives us the extra flexibility if we want it. Longer term we could write a generic interface for giving us the top-down view in parallel etc. (maybe by inheriting from ``AntAgent``).

neworderofjamie: No, that is fair enough!

alexdewar:
Hmm well the stupid OpenCV thing seems to have gone away by itself now (maybe an update to Ubuntu fixed it?).

@neworderofjamie  are you happy to merge?

PR #31, opened 2018-07-23
Ant world units

neworderofjamie:
Adding units to ant_world - definitely makes the ``Parameters`` namespace way more meaningful! In the process I got a little distracted moving ``using namespace`` directives out of the global namespace in headers. That's generally really nasty behaviour as there's no way to UNusing namespaces! As units is going to be standard in BoB Robotics I've fixed most of these by just moving the usings into one of our namespaces. In other smaller headers where it's less irritating/there's no literals I've just written them out longhand.


alexdewar:
This looks nice. I've been tidying up the makefiles on another branch and just noticed that I borked the ant_world code with the last set of changes I made!

neworderofjamie:
It builds through the GeNN 'user makefile' which may make it extra borky :)

ant_world/parameters.h
```diff
@@ -68,7 +75,7 @@ namespace Parameters
     constexpr unsigned int displayRenderHeight = intermediateSnapshotHeight * displayScale;
 
     // Ant parameters
-    // How fast does the ant move?
-    constexpr float antTurnSpeed = 4.0f;
-    constexpr float antMoveSpeed = 0.05f;
+    // How fast does the ant move each timestep?
+    constexpr degree_t antTurnSpeed = 4.0_deg;
+    constexpr meter_t antMoveSpeed = 0.05_m;
```
alexdewar: Just a minor nit: I'm not sure "speed" is the right word here (I thought you meant real velocity for a mo). How about ``antTurnStepSize`` and ``antStepSize`` (or something less clunky)?

imgproc/unwrapfile_src/unwrap.h
```diff
@@ -36,20 +33,20 @@ void getmaps(Mat &map_x, Mat &map_y, Size ssrc)
...
 {
     if (!im.data) {
-        cerr << "Error: Could not read data" << endl;
+        std::cerr << "Error: Could not read data" << std::endl;
         return false;
     }
     if (im.rows != im.cols) {
-        cerr << "File is not panoramic -- skipping " << filepath << endl;
+        std::cerr << "File is not panoramic -- skipping " << filepath << std::endl;
         return false;
```
alexdewar: Thanks for tidying up here. I have somewhat neglected this bit of code :-/

libantworld/agent.h
```diff
@@ -73,8 +75,8 @@ class AntAgent
     }
 
 private:
-    Vector3<degree_t> m_Attitude{ 0_deg, 0_deg, 0_deg };
-    Vector3<meter_t> m_Position{ 0_m, 0_m, 0_m };
+    Vector3<degree_t> m_Attitude{{ 0_deg, 0_deg, 0_deg }};
+    Vector3<meter_t> m_Position{{ 0_m, 0_m, 0_m }};
```
alexdewar: Do the extra set of curlies do anything? I'm guessing it's calling the array constructor with an array argument rather than passing the args directly, but does that change things?

libantworld/common.h
```diff
@@ -1,6 +1,30 @@
...
+}
+
+template<class T1, class T2>
+auto distance2(const T1 &v1, const T2 &v2)
+{
+    return hypot(v2[1] - v1[1], v2[0] - v1[0]);
+}
+
+template<class T>
+units::length::meter_t distance2(const T &v1, units::length::meter_t x2, units::length::meter_t y2)
```
alexdewar: I just remembered I meant to rename these to ``distance()`` because they aren't distance squared any more...

libantworld/renderer.cc
```diff
@@ -252,10 +252,10 @@ void Renderer::generateCubeFaceLookAtMatrices()
...
                           degree_t yaw, degree_t pitch, degree_t roll)
 {
-    glRotated(roll.value(), 0.0, 1.0, 0.0);
-    glRotated(pitch.value(), 1.0, 0.0, 0.0);
-    glRotated(yaw.value(), 0.0, 0.0, 1.0);
-    glTranslated(-x.value(), -y.value(), -z.value());
+    glRotatef(roll.value(), 0.0, 1.0f, 0.0);
+    glRotatef(pitch.value(), 1.0f, 0.0, 0.0);
+    glRotatef(yaw.value(), 0.0, 0.0, 1.0f);
+    glTranslatef(-x.value(), -y.value(), -z.value());
```
alexdewar: I also forgot to this. Sorry!

libantworld/route_ardin.cc
```diff
@@ -204,8 +189,8 @@ bool RouteArdin::load(const std::string &filename, bool realign)
             const radian_t heading = m_Headings[i - 1];
 
             // Realign segment to this angle
-            waypoint[0] = prevWaypoint[0] + (0.1f * cos(heading));
-            waypoint[1] = prevWaypoint[1] - (0.1f * sin(heading));
+            waypoint[0] = prevWaypoint[0] + (0.1f * units::math::cos(heading));
+            waypoint[1] = prevWaypoint[1] - (0.1f * units::math::sin(heading));
```
alexdewar: I suppose it should probably be ``0.1`` rather than ``0.1f`` here. I'm sure the units library would complain if it didn't like casting a float here, but it is using doubles internally.

libantworld/agent.h
neworderofjamie: It's required for more elderly compilers

ant_world/parameters.h
neworderofjamie: Hah yeah - the dubious physics of the antworld is exposed :grimacing:

libantworld/common.h
neworderofjamie: I figured that the removal of square roots at all costs optimisation was probably not really necessary these days....

neworderofjamie:
I've fixed those few issues and hidden the C++14 from GeNN if you wanna check it over and merge!

PR #32, opened 2018-07-24
Reorganise examples + use common build flags

alexdewar:
I've added a bunch of makefiles for projects and examples which weren't using them and tidied up the existing ones. I also added a small makefile to be included in the others with common parameters for CXXFLAGS and LINK_FLAGS and a couple of other little things. I have build-tested everything and fixed a few random warnings.

I've put all the examples into the Examples folder and there's a script in there which tries to build them all (this is partly with an eye to future CI....).


Examples/build_all_examples.sh
```diff
@@ -0,0 +1,28 @@
...
+    else
+        echo -e "\e[31m========== Building $project failed! ==========\e[39m"
+    fi
+    (( projectcount++ ))
+    cd ..
+done
+
+echo -e "\e[34m
+BUILD COMPLETED
+[$goodcount/$projectcount of the projects built successfully]\e[39m"
```
neworderofjamie: this would probably be nicer as a recursive make file - I think it can fairly easily be made to do this more tersely

Examples/ant_world_test/ant_world_test.cc
```diff
@@ -8,14 +8,14 @@
 #include <GLFW/glfw3.h>
 
 // BoB robotics includes
-#include "../hid/joystick.h"
+#include "hid/joystick.h"
```
neworderofjamie: It's much nicer not having ../ at the start of all the include!

Examples/camera_360_test/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+camera_360_test:
+	$(CXX) camera_360_test.cc $(CXXFLAGS) $(LINK_FLAGS) -o camera_360_test
+
+clean:
+	rm -f camera_360_test *.d
+
+%.d: ;
+
+-include camera_360_test.d
```
neworderofjamie: Does this definitely work? I thought the .d needed to be included before the target

Examples/camera_360_test/Makefile
```diff
@@ -0,0 +1,16 @@
+include ../../common/flags.mk
+CXXFLAGS += -I../..
+
+.PHONY: all clean
+
+all: camera_360_test
+
+camera_360_test:
```
neworderofjamie: Shouldn't this depend on the .cc and the .d?

common/flags.mk
```diff
@@ -0,0 +1,17 @@
+# This makefile is just an easy way to standardise the compiler flags we're using
+# between different samples/mini-libraries.
+#
+# Just put "-include ../common/flags.mk" in your makefile.
+
+CPP_STANDARD = c++14
```
neworderofjamie: I get super-confused  about the different sorts of equals in Makefiles but maybe this should be a ?= so you can set it and then include this. Also maybe add a comment that this is GeNN magic!

common/lm9ds1_imu.h
```diff
@@ -511,14 +512,14 @@ class LM9DS1
...
-                       [this](int16_t v, int16_t bias){ return v - bias; });
+                       [](int16_t v, int16_t bias){ return v - bias; });
     }
     
     void readAccel(int16_t (&data)[3])
     {
         readAccelGyroData(AccelGyroReg::OUT_X_L_XL, data);
         std::transform(std::begin(data), std::end(data), std::begin(m_AccelBias), std::begin(data),
-                       [this](int16_t v, int16_t bias){ return v - bias; });
+                       [](int16_t v, int16_t bias){ return v - bias; });
```
neworderofjamie: Good spot!

common/lm9ds1_imu.h
```diff
@@ -766,6 +767,8 @@ class LM9DS1
                 return 0.0175f;
             case GyroScale::DPS2000:
                 return 0.07f;
+            default:
```
neworderofjamie: As these are modern enums, surely this isn't necessary? Also if it is - it's definitely an error so should probably throw rather than return anything that could be ignored/cause maths horror

imgproc/opencv_optical_flow.h
```diff
@@ -68,16 +68,16 @@ class OpenCVOpticalFlow
 
     void render(cv::Mat &outputImage, unsigned int scale)
     {
-        assert(outputImage.cols == m_FlowX.cols * scale);
-        assert(outputImage.rows == m_FlowX.rows * scale);
+        assert(outputImage.cols == m_FlowX.cols * (int) scale);
```
neworderofjamie: maybe just change ``scale`` to an ``int`` and add ``assert(scale >= 1)`` - using ``unsigned int`` is a bad habit of mine

Examples/build_all_examples.sh
alexdewar: Good idea!

Examples/camera_360_test/Makefile
alexdewar: Good spot.

common/flags.mk
alexdewar: :+1:

common/lm9ds1_imu.h
alexdewar: I'd assume not, but gcc was giving me warnings (I think now that we have -Wpedantic enabled). It is a little odd. I suppose maybe theoretically one can cast a ``uint8_t`` to ``GyroScale`` even if it's not a valid value?

I'll change it to a throw instead.

imgproc/opencv_optical_flow.h
alexdewar: Ok, will do

Examples/build_all_examples.sh
alexdewar: I have just tried playing around with a makefile and it wasn't as easy as I thought... For some reason this:
```
%_test:
    $(MAKE) -C $@
```
...doesn't seem to find the folders and without pattern matching it ends up being quite clunky.

Also I deliberately made the script cleanbuild everything so that we can check if stuff has broken, so we don't want the makefile skipping any of the projects if they're already built.

common/lm9ds1_imu.h
neworderofjamie: Ahh ok - throw in that case! I feel newer gccs are getting _more_ pedantic - on Ubuntu 18 it raises some new warnings deep in a particularly agregious bit of GeNN

neworderofjamie:
How come we're not building everything with O2 (or even O3) if we're not debugging? In these glorious days of functioning open source compilers that doesn't **generally** break anything...

alexdewar:
It's more that we don't care about optimisation for the example programs, but we could equally go with -O2 across the board. I guess the difference in build time won't be much?

PR #37, opened 2018-07-27
Generic image database header

alexdewar:
I've taken the code for writing to the image database (i.e. saving an image and adding a line to the CSV file) and put it in its own header in a generic form.

Now you can do:
```C++
ImageDatabaseRecorder database("database_name", /*followRoute=*/ false);
...
database.saveImage(frame, 0.5_m, 0.5_m, 0_m, 0_deg);
```

The main impetus for this was is so that Norbert can save his image databases from the robot in the same form.

### DIFF
--- a/common/image_database_recorder.h
+++ b/common/image_database_recorder.h
@@ -1,0 +1,68 @@
+// Standard C includes
+#include <cstdio>
+
+// Standard C++ includes
+#include <fstream>
+#include <iostream>
+#include <stdexcept>
+#include <string>
+#include <tuple>
+
+// OpenCV
+#include <opencv2/opencv.hpp>
+
+// Third-party includes
+#include "../third_party/path.h"
+#include "../third_party/units.h"
+
+namespace BoBRobotics {
+using namespace units::length;
+using namespace units::angle;
+
+template<class AgentType>
+class ImageDatabaseRecorder
+{
+public:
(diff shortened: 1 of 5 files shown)


common/image_database_recorder.h
```diff
@@ -0,0 +1,64 @@
...
+
+        // Write image file info to CSV file
+        m_CSVStream << x.value() << ", " << y.value() << ", " << z.value() << ", "
+                    << heading.value() << ", " << filename << std::endl;
+    }
+
+private:
+    std::string m_DatabaseName;
+    std::ofstream m_CSVStream;
+    bool m_IsRoute;
```
neworderofjamie: This could be const as you set it in the initialiser list

video/input.h
```diff
@@ -98,6 +102,13 @@ class Input
         }
     }
 
+    virtual void readFrameSync(cv::Mat &outFrame)
```
neworderofjamie: I'm not really certain this should be generic method at this level. It only really seems relevant to the Network sources for which it might make more sense to add a 'blocking mode' as that's something that can be done in a much more efficient way.

alexdewar: Good point. I've reverted this (I didn't even end up using the function in the end anyway).

common/image_database_recorder.h
alexdewar: Done!

PR #38, opened 2018-07-31
Build fixes

neworderofjamie:
Having massaged the Jenkins script a little so it actually works, I fixed a few errors it threw up. There's still lots of linker failures on your machine for I2C stuff though and my machine doesn't have the ARSDK installed for all users - I think maybe we need to think carefully about what actually gets automatically tested and what counts as a failure

### DIFF
--- a/video/netsource.h
+++ b/video/netsource.h
@@ -32,12 +32,12 @@ public:
     NetSource(Net::Node &node)
     {
         // handle incoming IMG commands
-        node.addCommandHandler("IMG", [this](auto &node, auto &command) {
+        node.addCommandHandler("IMG", [this](Net::Node &node, const Net::Command &command) {
             onCommandReceived(node, command);
         });
 
         // when connected, send command to start streaming
-        node.addConnectedHandler([](auto &node) {
+        node.addConnectedHandler([](Net::Node &node) {
             node.getSocket()->send("IMG START\n");
         });
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -66,10 +66,7 @@ for(b = 0; b < builderNodes.size; b++) {
                         
                         // Build tests and set build status based on return code
                         def statusCode = sh script:"./build_all_examples.sh 1> \"" + uniqueMsg + "\" 2> \"" + uniqueMsg + "\"", returnStatus:true
-                        if(statusCode == 0) {
-                            setBuildStatus("Building examples", "SUCESSS");
-                        }
-                        else {
+                        if(statusCode != 0) {                            
                             setBuildStatus("Building examples", "FAILURE");
                         }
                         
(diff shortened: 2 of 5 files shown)


video/netsource.h
```diff
@@ -32,12 +32,12 @@ class NetSource : public Input
     NetSource(Net::Node &node)
     {
         // handle incoming IMG commands
-        node.addCommandHandler("IMG", [this](auto &node, auto &command) {
+        node.addCommandHandler("IMG", [this](Net::Node &node, const Net::Command &command) {
             onCommandReceived(node, command);
```
alexdewar: It's weird that this doesn't work on your compiler, because this is definitely part of the C++14 standard.

Obvs we don't *need* this here but maybe we're building something with C++11 by mistake?

alexdewar:
We def don't have to build everything on all nodes, but I think it would be good if we could fix the build failures. I'll have a tinker with the I2C stuff at some point and see if I can get it working.

I would like it if we could test the Bebop stuff on your machine -- it would be good to check it works with Ubuntu. Would you mind setting up the ARSDK so it works for the jenkins user too? You could just bung it in /opt/arsdk or something and set ARSDK_ROOT to point there.

neworderofjamie:
Sure - will do

alexdewar:
Sorry @neworderofjamie -- were you waiting for me to merge this? I am happy with it all if you're ready to merge.

neworderofjamie:
Did you fix the linker errors on your machine? I thought you were gonna investigate :smile:

alexdewar:
Didn't mean to close this. Oops!

I haven't fixed the linker errors on my machine, but I suspect it could be fiddly (it looks like there are some declarations in the header file without definitions in the .so for some reason). I was thinking we could just merge for now, given that master branch is also failing checks.

We could just exclude that test from jenkins for now (with an if-statement in the shell script) and worry about it later.

But I will have a look now and see if I can fix the linker error, which is the nicer solution.

neworderofjamie:
Weird - With I2C there is a kernel mode and a usermode interface (both of which look identical). I'm wondering if somehow on newer linuxes we're including the kernel mode one by mistake

alexdewar:
I just tried *not* including the kernel header (on Arch and Ubuntu 18.04) and then it won't compile at all, because there's a missing #define for I2C_SLAVE. This is all a bit weird.

As long as this example compiles ok on the Jetson I guess that's the main thing; maybe it's some weirdness of x86. In that case, I'd be tempted to just exclude lm9ds1_test from build-all-examples.sh. Is the Jetson running Ubuntu 18.04 btw?

neworderofjamie:
The Jetson currently runs 16.04 (as does my desktop on which it compiles fine). I'm suspicious there's an extra package required on 18/arch

alexdewar:
Ok so adding ``-li2c`` to the LINK_FLAGS *does* work, though I could swear I tried it before without success (maybe something to do with the order of arguments to gcc?).

The only thing is that we *don't* want this extra link flag for Ubuntu <=16.04. I'll have a play with the makefile and see if I can come up with an appropriate if statement

neworderofjamie:
Behold the green tick of righteousness!

alexdewar:
Cool! Well I'm happy to merge whenever you are...

PR #39, opened 2018-08-06
Joystick fixes

neworderofjamie:
I was trying to use the camera recorder and found a whole bunch of stuff had got broken:
- The Makefile wasn't building dependencies properly and the vicon target was named wrong
- Renamed the implementation of ``JoystickBase::updateAxis`` that actually raises events to aid clarity
- ``Tank::onJoystickEvent`` was really broken - ``m_X`` and ``m_Y`` never got updated


neworderofjamie:
How comes there's ``m_AxisStates`` with different types in ``JoystickBase`` and ``JoystickLinux``? That seems mega-confusing and I was convinced that was the problem for some time....

camera_recorder/Makefile
```diff
@@ -4,10 +4,14 @@ include ../common/flags.mk
...
+
+camera_recorder: camera_recorder.cc camera_recorder.d
 	$(CXX) camera_recorder.cc $(CXXFLAGS) -DI2C $(LINK_FLAGS) -o camera_recorder
 
-camera_recorder_vicon: camera_recorder.cc
-	$(CXX) camera_recorder.cc $(CXXFLAGS) -DVICON_CAPTURE $(LINK_FLAGS) -o camera_recorder_vicon_tracking
+-include camera_recorder_vicon.d
+
+camera_recorder_vicon: camera_recorder.cc camera_recorder_vicon.d
+	$(CXX) camera_recorder.cc $(CXXFLAGS) -DVICON_CAPTURE $(LINK_FLAGS) -o camera_recorder_vicon
```
alexdewar: Ugh sorry about that!

alexdewar:
Re ``m_AxisStates``, we could consider renaming it...

But the reason there is a separate ``int16_t`` array in ``JoystickLinux`` is because we want to keep a record of the x and y states of the thumbsticks, because we're using a circular deadzone and we get updates to the x and y states separately. So it could be that an update to the x state would mean that ``hypot(x,y) > deadZone``, but we need to know what y is to calculate this.

alexdewar:
And sorry for having broken stuff :-/

alexdewar:
How about we rename it to ``m_ThumbStickStates`` and stick a comment in there explaining this?

alexdewar:
It's worth adding that ``m_AxisStates`` in ``JoystickBase`` contains the axis values *after deadzoning*, partly so we don't repeatedly tell the joystick event handler that the axis value is zero

neworderofjamie:
As long as it works, renaming and commenting would totally work for me but couldn't we just have the platform-specific aspect deal with converting from platform-specific types to normalised float in m_AxisState and then leave it up to the base class to do deadzoning of whatever sort?

alexdewar:
That probably would a cleaner way of doing it - and we could get rid of the recurring template pattern then.

alexdewar:
But we would still need a separate array for the thumbsticks, because you need to keep track of their values a) so that you can check whether a change to another axis pushes both of them out of the deadzone and b) whether the "final" values for these axes (after deadzoning) have changed, so you can decide whether to raise an event or not. Essentially you need the undeadzoned values to use internally and the deadzoned values for event handling

alexdewar:
But I agree that we could just pass floats to ``JoystickBase`` and let ``JoystickLinux`` and ``JoystickWindows`` worry about how to turn axis values into floats

neworderofjamie:
Maybe just make it clearer with ``m_UnDeadzonedAxisState``

neworderofjamie:
Ok, I think this is quite a lot nicer and seems to work nicely on Linux and Windows, what do you think?

hid/joystick_base.h
```diff
@@ -179,58 +216,103 @@ class JoystickBase : public Threadable
...
+    //------------------------------------------------------------------------
+    void setPressed(JButton button, bool isInitial)
+    {
+        // Get state and set StateDown and StatePressed
+        uint8_t state = getState(button);
+        state |= (StateDown | StatePressed);
+
+        // Set new state
+        setState(button, state, isInitial);
+    }
```
alexdewar: I don't really understand what's going on in this function. Couldn't you just write:
```
setState(button, StateDown | StatePressed, isInitial);
```
I'm a bit worried that by or-ing state you could also have ``StateReleased`` set. Or am I missing something?

alexdewar: Ah wait I get it. I suppose you could have a down event and a release event in the same update...

hid/joystick_base.h
```diff
@@ -179,58 +216,103 @@ class JoystickBase : public Threadable
...
-            updateAxis(axisVert, 0.0f, isInitial);
-        } else {
-            updateAxis(axisHorz, x, isInitial);
-            updateAxis(axisVert, y, isInitial);
+        // If deadzone is enabled
+        if(m_DeadZone > 0.0f) {
+            // If axis and perpendicular axis are within circular deadzone, return 0
+            const float statePerpendicular = m_AxisState[toIndex(axisPerpendicular)];
+            if (sqrt((state * state) + (statePerpendicular * statePerpendicular)) < m_DeadZone) {
+                return 0.0f;
```
alexdewar: Could write ``hypot(state, statePerpendicular)`` instead

alexdewar:
This does look much cleaner. Cheers for doing this! I've just made a few small comments.

I'm guessing this is failing the check because the I2C fix hasn't been merged into this branch...?

alexdewar:
The joystick code builds ok on my machine, so I'm happy to merge if you are.

PR #40, opened 2018-08-06
I2C linking fix for lm9ds1_test

alexdewar:
Ok, so I *finally* have this working. What a faff. It's been tested on Arch and Ubuntu 18.04 and it *should* work on Ubuntu 16.04 too (it relies on a shell script which I have tested).

Let me know if it works for you too and maybe then we can finally merge #38.


neworderofjamie:
I think this is just hacking around the problem - from looking at /usr/linux/i2c-dev.h on my ubuntu 16 and 18 machines, they are totally different. The ubuntu 16 one is the user mode, header only library we're trying to use and the ubuntu 18 one is a bit of the kernel with everything declared as extern.  Extern c and linking might make it compile but I'm 99% sure it won't actually work. Either we just disable this example for automated tests or figure out how to install i2c-tools on newer linuxes. See https://stackoverflow.com/questions/24180545/install-libi2c-dev-on-arch-linux and https://stackoverflow.com/questions/25159064/why-are-i2c-smbus-function-not-available-i2c-embedded-linux for other (not particularly helpful) comments to that effect

neworderofjamie:
Ubuntu 16
```
jk421@inf900801:~/genn_robotics/examples/lm9ds1_test$ dpkg -S /usr/include/linux/i2c-dev.h
diversion by libi2c-dev from: /usr/include/linux/i2c-dev.h
diversion by libi2c-dev to: /usr/include/linux/i2c-dev.h.kernel
libi2c-dev, linux-libc-dev:amd64: /usr/include/linux/i2c-dev.h
```

neworderofjamie:
Ubuntu 18
```
jamie@jamie-Latitude-5490:~/GeNN_Robotics/examples/lm9ds1_test$ dpkg -S /usr/include/linux/i2c-dev.h
linux-libc-dev:amd64: /usr/include/linux/i2c-dev.h
```

alexdewar:
I'm not sure actually. It seems like the functions relating to smbus (e.g. ``i2c_smbus_write_byte()``) aren't available as part of the kernel's public API any more. I haven't been able to find out why/when these functions were removed, but can't find them after grepping through the Linux header files I have.

It's true that i2c-tools is a userspace library, but the smbus functions seem to be pretty small wrappers for ioctls (that should probably have just been inlined in a header somewhere):
https://git.kernel.org/pub/scm/utils/i2c-tools/i2c-tools.git/tree/lib/smbus.c

neworderofjamie:
Exactly - they **are** inlined in the i2c-dev header if you have that package installed correctly - it's nothing to do with the kernel.

alexdewar:
I've just had a look through the git history for i2c-tools and it says that they hived off the ``i2c_smbus_*()`` functions from the inlined ones in ``linux/i2c-dev.h``. So it seems that this is how you have to access these functions now.

https://git.kernel.org/pub/scm/utils/i2c-tools/i2c-tools.git/commit/?id=251d594fd0d338c7b85707a2e72153a404a25b10

neworderofjamie:
yep - my digging has found the same:

https://packages.ubuntu.com/xenial/all/libi2c-dev/filelist
vs
https://packages.ubuntu.com/bionic/amd64/libi2c-dev/filelist

neworderofjamie:
Apologies - it's been a trying (and warm) day!

common/is_ubuntu_1604_or_older.sh
```diff
@@ -0,0 +1,12 @@
...
+# Ubuntu 16.04 or older.
+
+# Check for lsb_release program
+command -v lsb_release > /dev/null 2>&1 || exit 1
+
+# Check we're running on Ubuntu
+if [ ! $(lsb_release -is) = Ubuntu ]; then exit 1; fi
+
+# Check we're running on Ubuntu 16.04 or earlier
+if [ $(lsb_release -rs | sed 's/\.//') -gt 1604 ]; then exit 1; fi
```
neworderofjamie: Ubuntu 17 also has the userland header-only libi2c-dev https://packages.ubuntu.com/artful/all/libi2c-dev/filelist

neworderofjamie: I feel that this will also effect Debian, Fedora et al so I feel what we **want** to test for is whether the version of i2c-tools is >= 4 but this is infuriatingly difficult - all the actual tools are so shitly written they don't have --version :/

neworderofjamie: Ok, after looking at the source for i2cdetect, -V **is** implemented just not documented. Grepping this would definitely be a better test:

On Ubuntu 16 (x86_64 and ARM64):
``i2cdetect version 3.1.1``
On Ubuntu 18 (x86_64):
``i2cdetect version 4.0``

neworderofjamie:
If you don't have time/have better things to do I'll have a go at a better version test based on the i2ctools version (``echo -e "$1\n$2" | sort -V`` seems like a good basis for this)

PR #41, opened 2018-08-07
I unwrapfile got somewhat left behind by bob_robotics progress

neworderofjamie:
* Removed mesh of header files
* Removed bespoke unwrapping implementation
* Added command line options for controlling unwrap resolution and camera config
* Added yaml config matching whatever hardcoded values were - @alexdewar - any ideas what this was for?
* Extension checking was broken


neworderofjamie:
OK so I moved the config-finding logic into the unwrapper (no idea why it wasn't there before - it makes way more sense)

PR #42, opened 2018-08-07
Reorganise various BoB robotics programs

alexdewar:
I've been thinking for a while that we should perhaps move some of the models/helper programs, in part so we have the option of adding them to our Jenkins tests if we so desire.

Commit summary:
* Rename ant_world to ardin_mb
* Put ardin_mb and stone_cx into projects/ folder
* Put the various helper programs into tools/ folder
* Add a script for building all the helper programs, like the one we have for the examples

We might not want to test-build all of these helper programs for CI, in which case we could do something like add a .test_ignore file to those folders and check for it when building.

I've build-tested everything and it all works for me.


projects/ardin_mb/README.md
```diff
@@ -1,4 +1,4 @@
-# Arden et al. Model of navigation
+# Ardin et al. Model of navigation
```
neworderofjamie: :blush:

common/build_all_function.sh
```diff
@@ -0,0 +1,30 @@
...
+
+build_all() {
+    types=$1; shift
+    goodcount=0
+    projectcount=0
+    for project in $@; do
+        if [ ! -d $project ]; then continue; fi
+        
+        echo -e "\e[34m========== BUILDING $project ==========\e[39m"
+        cd "$project"
```
neworderofjamie: With the GeNN projects you also need to generate the simulation code:

``genn-buildmodel -i $BOB_ROBOTICS_PATH model.cc`` for ardin_mb
``genn-buildmodel -c -i $BOB_ROBOTICS_PATH model.cc`` for stone_cx (cos it doesn't use GPU)

projects/ardin_mb/README.md
neworderofjamie: (that wasn't supposed to be a happy blush, more a shameful one but github didn't oblige)

common/build_all_function.sh
neworderofjamie: There's also the question of what version of GeNN it builds against...

alexdewar: Yeah this is why I haven't made a build_all.sh for the "projects" yet -- just for the examples and the tools.

I guess our options for the version of GeNN we use are:
* Whatever version is lurking on our machines already
* The latest version from master
* The latest release
In any case I agree it would be helpful to report this version (inc. the commit hash).

neworderofjamie: Ahh ok - I wasn't entirely clear on how integrated this is. Checking out GeNN master and building against that would definitely be the best thing to do. In GeNN's Jenkinsfile we use curl to download the tar gz of libraries we want to use from github, unzip them and point environment variables at them. I can have a go at that in another PR if you'd like?

alexdewar: Sure, that'd be awesome :-)

alexdewar:
I have written a build_all.sh for the projects now (with the GeNN stuff in) and it's all building for me.

alexdewar:
I also removed the _test bit of the example projects' folder names as it seems a bit redundant now

alexdewar:
@neworderofjamie are you happy if we merge this?

I think it's best if we do the Jenkins stuff as a different PR.

PR #44, opened 2018-08-20
Unwrapping videos to a series of JPGs

neworderofjamie:
As videos are kinda annoying to handle (and because OpenCV seems incapable of writing high quality ones) it seemed useful to have an option to unwrap MP4 files and save some subset of the frames to JPG. Also a bonus bugfix - previous we were assuming in several places that whatever was allocating ``ftype`` vector was zeroing the memory.

### DIFF
--- a/tools/unwrapfile/unwrapfile.cc
+++ b/tools/unwrapfile/unwrapfile.cc
@@ -29,8 +29,14 @@ enum class FileType {
     video
 };
 
+void saveMaxQualityJPG(const std::string &filename, const cv::Mat &image)
+{
+    std::vector<int> params{CV_IMWRITE_JPEG_QUALITY, 100};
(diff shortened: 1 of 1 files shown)


alexdewar:
This is def a good idea. And good spot with the bug!

The only suggestion I'd make is that we could think about having maybe another option to save the frames in the same format as we're using for other route databases (using the silly trivial functions in here: https://github.com/BrainsOnBoard/bob_robotics/blob/master/common/image_database.h). At the moment we're saving frames into a separate folder and calling them image_00000.png etc. (note the PNG format). This way it'll be easier to load in a recorded video into e.g. the perfect memory code as a route.

But we don't have to add this now and I'm happy to merge as is if you have other things to do... Otherwise it looks cool!

PR #45, opened 2018-09-03
Robot positioner norbert

tenxlenx:
all working, code cleaned up.


tenxlenx:
Hey Alex, Jamie

Yes the code is ready to be reviewed. just did a minor fix in main.cpp shortly after the pull request

alexdewar:
Cool, I'll take a look now.

robotPositioner/Makefile
```diff
@@ -0,0 +1,23 @@
+
```
neworderofjamie: There is a 'base' makefile in  ``common/flags.mk`` you can include to save some typing here -- see ``examples/see3cam/Makefile`` for an example

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
+#include <vector>
```
neworderofjamie: It's always good to put a ``#pragma once`` on the first line to prevent the compiler including header files (especially **large** header files more than once)

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+#include "../imgproc/opencv_unwrap_360.h"
+
+using namespace BoBRobotics::Vicon;
+using namespace BoBRobotics::ImgProc;
+using namespace BoBRobotics::Video;
+using namespace std::literals;
+
+#define PI 3.14159265359
+
+class robotPositioner {
```
neworderofjamie: I think this class does too much stuff! Things that definitely don't belong are:
- Vicon capture control
- Camera capture

neworderofjamie:
Damnit! Will fix tomorrow @alexdewar  - am reviewing as seperate comments so hopefully you can see and we can avoid duplication!

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+            cv::imwrite(fileName, unwrapped);
+            std::cout << " image " << m_imageNumber << " captured " << std::endl;
+        }
+    }
+
+//-----------------PUBLIC API---------------------------------------------------------------------
+public:
+
+    // constructor
+    robotPositioner(
```
neworderofjamie: I think it would be much nicer if this class just took references to a ``BoBRobotics::Robots::Tank`` and a ``Vicon::UDPClient<ObjectData>`` - this would:
- Reduce the number of parameters to the constructor
- Make it more flexible
- Make it easier to integrate into a larger program

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+            std::vector<double> temp_vec;
+            for (double field : row) {
+                temp_vec.push_back(field);         	
+            }
+            m_coordinates.push_back(temp_vec);
+        }
+    }
+
+    // drives the robot the desired pose (x,y,angle)
+    void driveRobotToGoalLocation() {
```
neworderofjamie: This API essentially blocks the program until the robot has physically got somewhere which I think is less than ideal! A nicer API might be:
- A function to set the goal position -- 
- A function to call every timestep to update the motors -- essentially the inner loop of this function
- A function to query whether the robot has reached it's goal (this could just be a return value of the update function)

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+        }
+    }
+
+    /**
+    *	starts the robot positioning program. The function expects a csv file with 4 columns |x|y|z|a|
+    *   where 'xyz' are the coordinates of the goal location and 'a' is the final angle.
+    *   the robot will visit all the coordinates in the file, then snap an image with the camera at each 
+    *   location if the camera is attached.
+    **/
+    void startSession(const std::string fileName) {
```
neworderofjamie: I think this functionality and the logic to read coordinates doesn't belong here - I think for now it's probably best to just move it into your example application.

tenxlenx:
Thanks for all the comments, they are really constructive! will do these fixes asap

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+    double m_pos_Z;                                           // robot's z position
+    double m_heading;                                         // heading (angle) of the robot
+    double m_goalPositionX;                                   // goal position x
+    double m_goalPositionY;                                   // goal position y
+    double m_goalPositionZ;                                   // goal position z
+    double m_bearingFromGoal;                                 // bearing (angle) from goal coordinate
+    double m_distanceFromGoal;                                // Euclidean distance from goal coordinate
+    double m_goalAngle;                                       // angle to turn after finding the correct location
+    double m_bearingFinal;                                    // heading converted to range <-180, 180> from <0, 360>
+
```
alexdewar: It would be great if you could use the units library we're using for all these values -- it gives you c++ types for metres, angles etc.: https://github.com/nholthaus/units

You'd just need to change the types of the positions from double to millimeter_t and the angle variables to degree_t or radian_t. See e.g. here for example: https://github.com/BrainsOnBoard/bob_robotics/blob/master/vicon/udp.h

alexdewar: ...and you can make liberal use of the ``auto`` keyword with it to reduce typing

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+    double m_k1;                                              // curveness of the path to the goal
+    double m_k2;                                              // speed of turning on the curves
+    double m_alpha;                                           // causes more sharply peaked curves 
+    double m_beta;                                            // causes to drop velocity if 'k'(curveness) increases
+    double m_max_velocity;                                    // max velocity
+
+    /**
+    * sets up the vicon system so it listens to the broadcast 
+    */
+    void setupVicon() {
```
alexdewar: As above: I feel like this code should live outside this class

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    template <class T, class S>
+    bool closeTo(const T num, const S threshold) {
+        if (num > -threshold && num < threshold) {
+            return true;
+        } else {
+            return false;
+        }
+    }
```
alexdewar: This function and the one above should be ``static`` because they don't change the object.

For ``closeTo()`` you could just write:
```
return (num > -threshold && num < threshold);
```

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+            std::string line;
+            while (getline(in, line)) {
+                std::stringstream sep(line);
+                std::string field;
+                fields.push_back(std::vector<double>());
+                while (getline(sep, field, ',')) {
+                    fields.back().push_back(stof(field));
+                }
+            }
+        }
```
alexdewar: I did write some generic image database code for this kind of situation: https://github.com/BrainsOnBoard/bob_robotics/blob/master/common/image_database_recorder.h

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+    // drives the robot wheels using v and w as velocities which get translated to 
+    // indiviual wheel velocities. Conversion from these velocities to individual 
+    // wheel velocities is done via solving simultaneous equations
+    // parameters:
+    //             v      : translational velocity
+    //             w      : rotational velocity
+    //             robot_r: wheel radius
+    //             robot_d: axis length
+    template <class T, class S>
+    void driveRobotWithVelocities(const T v, const T w, const S robot_r, const S robot_d) {
```
alexdewar: Not sure why this function is templated?

The units lib would be useful here so you can have unit types as parameters; as it is it's not clear what the units of velocity and distance etc. are

robotPositioner/robotPositioner.h
alexdewar: Agreed. I think the ``Tank`` object will have to be a pointer though rather than a reference because it's an abstract class and C++ is funny about that

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+    ~robotPositioner() {
+
+        if(!m_viconCaptureControl->stopRecording("test1")) {
+            std::cout << "failed to stop recording" << std::endl;
+        }
+
+        delete m_vicon;
+        delete m_viconCaptureControl;
+        delete m_bot;
+
```
alexdewar: As mentioned before, it would be best to move these objects outside the scope of the class, but aside from that you also don't really want to use ``new`` and ``delete`` in modern C++. For the most part you can avoid using pointers at all and just pass by reference, but if you have to it's much better to use some kind of smart pointer.

robotPositioner/robotPositioner.h
alexdewar: It would be great if you could also use this here so we have a standard way of reading and writing these things :-)

robotPositioner/robotPositioner.h
alexdewar: :+1:

robotPositioner/robotPositioner.h
alexdewar: I should probably clarify that I'm not saying regular pointers are always bad! It's just ``new`` and ``delete`` you want to avoid. It's fine to pass a regular pointer so long as some other scope "owns" it and will clean it up (e.g. because it's using a ``unique_ptr`` or something)

robotPositioner/robotPositioner.h
alexdewar: Yep.

On a related note: your example application could go into the examples/ folder and the header should live somewhere else with the other library code... maybe let's put it in common/ for now.

robotPositioner/robotPositioner.h
```diff
@@ -0,0 +1,382 @@
...
+        m_cam = new See3CAM_CU40(device, See3CAM_CU40::Resolution::_1280x720);
+        const cv::Size unwrapSize(180,50); // calculate camera input dimensions
+
+        const cv::Mat bubblescopeMask = See3CAM_CU40::createBubblescopeMask(m_cam->getSuperPixelSize());
+        m_cam->autoExposure(bubblescopeMask);
+        m_unwrapper = m_cam->createUnwrapper(unwrapSize);
+    }
+
+    // gets vicon data
+    double getViconData() {
```
alexdewar: This function should be ``void`` because it doesn't return anything. I think the name should probably be something like ``updateViconData()`` because otherwise it sounds like it should return something.

alexdewar:
Another thing I forgot to mention is that we have a function for loading a panoramic camera in video/panoramic.h. Use it like this:
```
auto cam = getPanoramicCamera();
```

It will try to find a See3Cam by default, then falls back on trying to find another kind of panoramic camera. It also does all the business of setting the parameters for the camera for you.

robotPositioner/robotPositioner.h
neworderofjamie: I **believe** references to abstract classes **should** work:

```c++
#include <iostream>

class Bar
{
public:
    virtual void update() = 0;
};

class Foo : public Bar
{
public:
    virtual void update() override
    {
        std::cout << "Updating foo" << std::endl;
    }
};

void updateBar(Bar &bar)
{
    bar.update();
}

int main()
{
    Foo a;
    updateBar(a);
    
    return EXIT_SUCCESS;
}
```

neworderofjamie:
@alexdewar - I just looked at the build errors and it **looks** like they're actually on your machine INF900802. Really non-memorable machine names is definitely one of the many weaknesses of centralized IT.....

alexdewar:
@neworderofjamie - Oops! Looks like you're right. It seems that I need to rebuild OpenCV because of some sofile change... I'll do that now.

alexdewar:
@neworderofjamie - all done and the examples are now building on my machine again

neworderofjamie:
Phew! I was worried what I could possibly have done to cause that :O

robotPositioner/robotPositioner.h
tenxlenx: there is a clear issue with this library when doing equations. I have w = k*v  where 'w' is angular velocity, 'k' is a constant and 'v' is velocity. It does not let me to calculate it because of type mismatch. I am not even surprised because k*v is still just a velocity which is not the type w needs to be

A possible workaround is that I am extracting the values and work with that, otherwise, I am getting all sorts of errors. I think this library is not able to deal with equations the way we expect it. Please double check and correct me if I am wrong.

neworderofjamie: Well one solution would be to give k a unit in angle per metre (which is a sensible unit of turning speed)

alexdewar: Yeah, Norbert just emailed me about this and I said the same thing. It does depend on whether w is a real physical unit though, rather than something like the turning speed passed into ``Tank.move()``.

But if v is a value in m/s and w is in deg/sec, then k will have a value in deg/m and you should be able to express it like this:
```
const auto k = 2_deg / 1_m; // for k==2
```

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+        meters_per_second_t v = m_max_velocity/scalar_t( (1+m_beta*pow(abs(k.value()),m_alpha)));
+        degrees_per_second_t w = k*v;
+
+        driveRobotWithVelocities(v,w, m_robot_r, m_robot_D);  
+    }
+
+    //! returns true if the robot reached the goal position
+    bool didReachGoal() {
+        return (m_distanceFromGoal < m_stopping_distance && 
+            closeTo(m_bearingFromGoal,m_allowed_heading_error));
```
neworderofjamie: as closeTo is only used once it doesn't make a lot of sense having it as a function - just stick the code in here

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+        // changing from <0,360> to <-180, 180>
+        m_bearingFromGoal = angleWrapAround(m_bearingFromGoal);
+        
+    } 
+
+//-----------------PUBLIC API---------------------------------------------------------------------
+public:
+
+    RobotPositioner(
+        BoBRobotics::Robots::Tank &tank,
```
neworderofjamie: This is a much nicer interface - my only thought is, how about you add virtuals for ``millimeter_t getWheelRadius() const`` and ``millimeter_t getAxisLength() const`` to ``BoBRobotics::Robots::Tank`` - I have a Surveyor on my desk if you wanna make whatever measurements are required

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+
+    
+    void driveRobotWithVelocities(const meters_per_second_t v,
+                                  const degrees_per_second_t w,
+                                  const millimeter_t robot_r,
+                                  const millimeter_t robot_d) {
+        // v = r * (Vl+Vr)/2
+        // w = r * (Vr-Vl)/D
+        double a =  (robot_r/2).value(); //convert to cm
+        double b =  (robot_r/2).value();
```
neworderofjamie: how come these two are the same?

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+        double c = ( robot_r / robot_d).value();
+        double d = (-(robot_r / robot_d).value());
```
neworderofjamie: d is also just -c  - there doesn't seem much point in having them seperate

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
+
+#pragma once
+
+// c++ standard includes
+#include <vector>
```
neworderofjamie: I'm pretty sure you're not using a lot of these headers any more!

common/robotPositioner.h
neworderofjamie: also it would be much better to use more meaningful variable names - for example I think ``radiusAxisLengthRatio`` would be much more useful that ``c``

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+
+        double e=(v.value()); // converting to millimeters/second 
+        double f=w.value();
+        double Vl,Vr;
+
+        // determinant
+        double det = (a*d - b*c);
+    
+        Vl = ((e*d - b*f)/det);
+        Vr = ((a*f - e*c)/det);
```
neworderofjamie: Finally, all these intermediate variables can be marked as ``const`` to make it clear that you don't change them once you calculate them

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+using namespace std::literals;
+using namespace units;
+using namespace units::length;
+using namespace units::time;
+using namespace units::velocity;
+using namespace units::literals;
+using namespace units::angle;
+using namespace units::constants;
+using namespace units::dimensionless;
+using namespace units::angular_velocity;
```
neworderofjamie: In general you should never ``using namespace`` in the global namespace as there's no way for other bits of code to 'undo' it - in this case the solution is to move ``RobotPositioner`` into the  ``BoBRobotics::Robots`` namespace (and move it into the robots folder)

alexdewar: Also I don't think you need this one:
```
using namespace units;
```
I think the other ``using``s cover this

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+        
+        // drive robot 
+        m_bot.tank(Vl, Vr);
+    }
+
+    
+    degree_t angleWrapAround(degree_t angle) {
+        // changing from <0,360> to <-180, 180>
+        if (angle <= degree_t(0)) { angle += degree_t(360); } 
+        if (angle > degree_t(180)) { angle -= degree_t(360); }
```
alexdewar: Just an observation for future reference: you can use the unit suffixes here. So you can write ``0_deg`` ``360_deg`` etc. But your way is also fine.

common/robotPositioner.h
alexdewar: That's an excellent idea!

common/robotPositioner.h
alexdewar: I think ``double det = (a*d - b*c);`` is equivalent to ``double det = -2 * a * c;``, right?

I would also be inclined not to bother converting all these values to doubles -- you could stick with the unit-typed variables, write ``auto det = [whatever]``, then the units will cancel out in the next calculation for ``Vl`` and ``Vr`` and I *think* you can then just treat them as doubles. But this is up to you.

common/robotPositioner.h
alexdewar: This function could also be marked ``static``...

common/robotPositioner.h
neworderofjamie: I did initially write a comment suggesting this as well but I'm semi-sure that Vl and Vr don't come out as reasonable units i.e. it's trying to do something like distance/time + angle/time

alexdewar: Ahh yes, my bad. You're totally right

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+                                  const millimeter_t robot_r,
+                                  const millimeter_t robot_d) {
+        // v = r * (Vl+Vr)/2
+        // w = r * (Vr-Vl)/D
+        double a =  (robot_r/2).value(); //convert to cm
+        double b =  (robot_r/2).value();
+        double c = ( robot_r / robot_d).value();
+        double d = (-(robot_r / robot_d).value());
+
+        double e=(v.value()); // converting to millimeters/second
```
alexdewar: I just noticed this: it looks like ``e`` will be a value in *metres* per second. @tenxlenx do you need it to be mm/s?

common/robotPositioner.h
```diff
@@ -0,0 +1,186 @@
...
+  
+
+    
+    void driveRobotWithVelocities(const meters_per_second_t v,
+                                  const degrees_per_second_t w,
+                                  const millimeter_t robot_r,
+                                  const millimeter_t robot_d) {
+        // v = r * (Vl+Vr)/2
+        // w = r * (Vr-Vl)/D
+        double a =  (robot_r/2).value(); //convert to cm
```
alexdewar: Also, this value is in mm not cm

neworderofjamie:
Hah don't worry - I won't merge until you're happy! That example application looks exactly the kind of thing I was thinking of though.

robots/robotPositioner.h
```diff
@@ -0,0 +1,181 @@
...
+
+        const millimeter_t robot_wheel_radius = bot.getRobotWheelRadius();
+        const millimeter_t robot_axis_length  = bot.getRobotAxisLength();
+        
+        const double a = (robot_wheel_radius/2).value(); 
+        const double b = (robot_wheel_radius/robot_axis_length).value();
+    
+        auto mmps = 1_mm/1_s;
+        mmps = v; // converting to mm/s
+        const double c = v.value();
```
alexdewar: I'm guessing this should read:
``
const double c = mmps.value();
``
...otherwise ``c`` will still be in m/s, rather than mm/s (which I guess is what you intended?).

alexdewar:
The other thing is that we're now getting test failures -- from looking at the log it looks like the robot positioner example won't build. @tenxlenx any thoughts? You can have a look at the build log if you click on "details" by the test failure message above...

@neworderofjamie your machine seems to have dropped from the CI for some reason?

alexdewar:
> Hey @tenxlenx. The robot simulator looks really cool. Good work! The only thing I'd say is for future reference that it could probably go in as its own PR, because it doesn't depend on the other work here. (You _could_ try moving the simulator code into its own branch and doing another PR if you want to practice your git skills -- but only if you want to.)

Btw I mean a new branch off master, not off this branch..

neworderofjamie:
I think my machine was just asleep cos I was working from home yesterday - it's back now!

Norbert - the CI errors are https://gen-ci.inf.sussex.ac.uk/job/BoB%20robotics/job/bob_robotics/job/robot_positioner_Norbert/24/artifact/msg_INF900801/*view*/

PR #46, opened 2018-09-05
Build updates

alexdewar:
This is just for some tiny changes to the build system.

- It turns out you can "opt in" to compiling different unit types. If we disable all except the ones we're using it massively improves build times for bits of code using units.h. If we ever decide we want any other unit types (e.g. acceleration) we can always adjust this.
- I've put the i2c makefile hacks into common/flags.mk, so that they can be used elsewhere. I've updated the robot versions of makefiles in stone_cx to use this, so they now compile for me


alexdewar:
Hmm, now my machine seems to be working again, but it looks like on your machine Jenkins went awry even before running the build test :-/

alexdewar:
It uses a named pipe rather than a temporary file at some point in the build (I forget which), so you're building it in RAM which is a bit faster.

Cheers for looking at this! :-)

PR #47, opened 2018-09-21
AntWorld rotater + other navigation code updates

alexdewar:
Hi Jamie,

The main change in this PR is that I've added a template argument so one can have an AntAgent rotate on the spot in AntWorld (using OpenGL) as opposed to the usual in-silico image rotation we do. (The tests I've done seem to show that the results aren't massively affected depending on which of these methods we use, but it's good to have the choice.)

I've also made a few, mostly superficial changes to the navigation code but I did also remove the option to save snapshots as one is going along. It made more sense to me to let the user manually save snapshots using the code in image_database.h if desired.

If you're ok with this PR, my next goal was to try to tweak the Ardin/Stone models so they're part of this navigation framework too.

### DIFF
--- a/navigation/insilico_rotater.h
+++ b/navigation/insilico_rotater.h
@@ -31,7 +31,7 @@ public:
     }
 
     template<class Func>
-    void operator()(Func func)
+    void rotate(Func func)
     {
         size_t i = 0;
         while (true) {
--- a/navigation/insilico_rotater.h
+++ b/navigation/insilico_rotater.h
@@ -9,6 +9,7 @@ class InSilicoRotater
 {
 public:
     InSilicoRotater(const cv::Size &unwrapRes,
+                    const cv::Mat &maskImage,
                     const cv::Mat &image,
                     const unsigned int scanStep = 1)
       : m_CurrentRotation(0)
@@ -19,11 +20,17 @@ public:
         assert(image.type() == CV_8UC1);
 
         image.copyTo(m_ScratchImage);
+        maskImage.copyTo(m_ScratchMaskImage);
     }
 
     cv::Mat &getImage()
     {
         return m_ScratchImage;
+    }
+
+    cv::Mat &getMaskImage()
+    {
+        return m_ScratchMaskImage;
     }
 
     bool next()
@@ -33,13 +40,9 @@ public:
             return false;
         }
 
-        // Loop through rows
-        for (int y = 0; y < m_ScratchImage.rows; y++) {
-            // Get pointer to start of row
-            uint8_t *rowPtr = m_ScratchImage.ptr(y);
-
-            // Rotate row to left by m_ScanStep pixels
-            std::rotate(rowPtr, rowPtr + m_ScanStep, rowPtr + m_ScratchImage.cols);
+        rollImage(m_ScratchImage);
+        if (!m_ScratchMaskImage.empty()) {
+            rollImage(m_ScratchMaskImage);
         }
         m_CurrentRotation = next;
 
@@ -59,7 +62,19 @@ public:
 private:
     int m_CurrentRotation;
     const unsigned int m_ScanStep;
-    cv::Mat m_ScratchImage;
+    cv::Mat m_ScratchImage, m_ScratchMaskImage;
+
+    void rollImage(cv::Mat &image)
+    {
+        // Loop through rows
+        for (int y = 0; y < image.rows; y++) {
+            // Get pointer to start of row
+            uint8_t *rowPtr = image.ptr(y);
+
+            // Rotate row to left by m_ScanStep pixels
+            std::rotate(rowPtr, rowPtr + m_ScanStep, rowPtr + image.cols);
+        }
+    }
 };
 } // Navigation
 } // BoBRobotics
--- a/navigation/insilico_rotater.h
+++ b/navigation/insilico_rotater.h
@@ -1,0 +1,65 @@
+#pragma once
+
+// OpenCV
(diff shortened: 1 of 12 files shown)


navigation/insilico_rotater.h
```diff
@@ -0,0 +1,74 @@
...
+
+    template<class Func>
+    void operator()(Func func)
+    {
+        size_t i = 0;
+        while (true) {
+            func(m_Image, m_MaskImage, i);
+
+            i += m_ScanStep;
+            if (i >= (size_t) m_Image.cols) {
```
neworderofjamie: Was there a particular reason this wasn't a ``for(size_t i = 0; i < (size_t)m_Image.cols; i += m_ScanStep)``? (It may well have come from my code originally  :smile: )

navigation/perfect_memory_base.h
```diff
@@ -57,65 +57,65 @@ class PerfectMemoryBase
...
         // Create vector to store RIDF values
         std::vector<std::vector<float>> differences(numSnapshots);
         for (auto &d : differences) {
-            d.resize(m_ScratchRollImage.cols);
+            d.resize(rotater.max());
         }
 
         // Scan across image columns
-        for (int col = 0; col < m_ScratchRollImage.cols; col += scanStep) {
+        rotater([this, &differences, numSnapshots](const cv::Mat &fr, const cv::Mat &mask, size_t i) {
```
neworderofjamie: Personally I prefer indenting this like:
```c++
rotater(
    [this, &differences, numSnapshots](const cv::Mat &fr, const cv::Mat &mask, size_t i)
    {
        ....
    });
```
```
So the lambda function is intended like a normal function but that is a **very** minor nit-pick

File: navigation/perfect_memory_base.h [hunk 3]
```diff
@@ -57,65 +57,65 @@ class PerfectMemoryBase
...
+     * The parameters are perfect-forwarded to the Rotater class, so e.g. for
+     * InSilicoRotater one passes in a cv::Mat and (optionally) an unsigned int
+     * for the scan step and for the AntWorldRotater, one passes in one or more
+     * angles.
+     */
+    template<class... Ts>
+    auto getHeading(Ts &&... args) const
     {
-        std::vector<std::vector<float>> differences = getImageDifferences(image);
+        auto differences = getImageDifferences(std::forward<Ts>(args)...);
```
neworderofjamie: allocating differences (potentially every frame) seems wasteful both in terms of memory (lots of snapshots * hundreds of columns * sizeof(float) isn't nothing and this is a lot of heap allocations) when all you're doing is finding the minimum - would it be possible (at the expense of some extra typing) to repeat the algorithm in this function and calculate the minimum inline?

navigation/perfect_memory_base.h
neworderofjamie: Also I don't think there's any need to use ``operator ()`` for this operation - I'd just give it a ``rotate`` method

navigation/plot.h
```diff
@@ -0,0 +1,49 @@
...
+
+//! Plot a line graph of a single RIDF
+void
+plotRIDF(const std::vector<float> &differences)
+{
+    // Vector to store modified RIDF
+    std::vector<float> ridf(differences.size() + 1);
+
+    // Rotate so data is in range -180 to 180
+    std::rotate_copy(differences.begin(),
```
neworderofjamie: You don't necessarily need to _resize_ ``ridf`` just _reserve_ it (i.e. you don't care what it's initial values are) so you could do:
```c++
std::vector<float> ridf;
ridf.reserve(differences.size() + 1);
std::rotate_copy(differences.begin(),
    differences.begin() + ceil((float) differences.size() / 2.0f),
    differences.end(),
    std::back_inserter(ridf));
```

navigation/perfect_memory_base.h
alexdewar: Makes sense

navigation/perfect_memory_base.h
alexdewar: You're right that it probs is a bit expensive. At least we should probably be reusing the vector if nothing else.

But the reason I went with storing the differences in vectors is so that we can get these values back out again for debugging and looking at the internal state. I did benchmark it and performance was basically the same as not having these big heap allocations, but I was only comparing a single frame. We might be able to get both options out by making a little iterator class for these differences though... I'll think about it.

Maybe in the meantime we could just settle for reusing the vector to save on the heap allocations though?

navigation/plot.h
alexdewar: Good spot!

navigation/insilico_rotater.h
neworderofjamie: :+1: past-somewhat was smort

navigation/perfect_memory_base.h
neworderofjamie: Yeah - caching that vector or vectors as a member of the class would indeed be an excellent middle ground.

navigation/perfect_memory_base.h
```diff
@@ -79,30 +79,8 @@ class PerfectMemoryBase
...
-                for (size_t s = 0; s < numSnapshots; s++) {
-                    // Calculate difference
-                    const auto diff = calcSnapshotDifference(fr, mask, s);
-                    differences[s][i] = diff;
-                }
-            });
-
-        return differences;
+        calcImageDifferences(std::forward<Ts>(args)...);
+        return m_Differences;
```
neworderofjamie: This function should probably now return a const reference - this will still copy as it won't move a member's data out from under it

PR #49, opened 2018-09-28
Switched OpenCV unwrapper to using units

neworderofjamie:
It wasn't compiling properly because units::maths had been injected into its namespace so I fixed it by switching it over to using units


alexdewar:
Looks good to me!

PR #50, opened 2018-10-01
Basic interface for robot gantry

alexdewar:
This is a basic version of code for accessing the robot gantry. At the moment, essentially you can just get/set its position and return to home. I've also added unwrapping params for the camera, which seems to work.

There's a whole bunch of extra things you can do with the gantry (e.g. continuous motion) but this was as much faffing with the API as I could be bothered to do for the time being!

I've also made some small (hopefully non-breaking) changes for a few functions where I'd gone a bit templating-mad.


examples/gantry/gantry.cc
```diff
@@ -0,0 +1,54 @@
+/*
+ * Note that there is currently an annoying bug with the gantry's camera.
+ * You have to open a TV tuner program *before* this example, so that the
+ * TV tuner card is set to PAL B mode (cf. NTSC) otherwise the display
+ * will be mangled. There is definitely a way to do this in one of the
+ * Windows video APIs (though seemingly not in OpenCV) and I'll add this
+ * in when I can be bothered. -- AD
+ */
```
neworderofjamie: OpenCV video does somehow seem to be singularly terrible in many ways...Maybe add this as a bug

robots/gantry.h
```diff
@@ -0,0 +1,243 @@
...
+// Standard C++ includes
+#include <algorithm>
+#include <chrono>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <thread>
+
+// Gantry-specifc includes
+#include "C:\Program Files\Advantech\Motion\PCI-1240\Examples\Include\Ads1240.h"
```
neworderofjamie: p.s. I googled this to see what the hell you're driving and it is described as being "designed for general-purpose extreme motion applications" - that is awesome

video/input.h
```diff
@@ -27,13 +27,10 @@ class Input
     /*!
      * \brief Create an ImgProc::OpenCVUnwrap360 object for this video stream
      * 
-     * @param args The resolution of the unwrapped image, as cv::Size or two ints
+     * @param args The resolution of the unwrapped image
      */
-    template<typename... Ts>
-    ImgProc::OpenCVUnwrap360 createUnwrapper(Ts &&... args)
+    ImgProc::OpenCVUnwrap360 createUnwrapper(const cv::Size &unwrapRes)
```
neworderofjamie: Hah I would agree that this was overkill :smile:

alexdewar:
Thanks for reviewing :-). I'll happily give you a demo when I've got it doing something a bit more interesting :-p

PR #53, opened 2018-10-03
Turn asserts() into exceptions

alexdewar:
I've added a BOB_ASSERT() macro which basically does the same thing as assert() but throws an exception rather than terminating the program. It also reports the file and line number.

The main reason to do this is so when we're using robots we can catch these exceptions, which will call the objects' destructors and stop the robot moving rather than continuing on into a wall.


alexdewar:
I guess we could print the error to ``std::cerr`` only on Windows then?

neworderofjamie:
I think that would be the simplest solution

PR #54, opened 2018-10-04
Fix calculation of headings in route_continuous.cc

alexdewar:
At the moment the calculation seems to give incorrect headings (not sure
exactly why). With this fix the headings look correct and when a route
is displayed on screen the views look sensible too (the agent seems to
face the right way).

### DIFF
--- a/libantworld/route_continuous.cc
+++ b/libantworld/route_continuous.cc
@@ -173,7 +173,7 @@ bool RouteContinuous::load(const std::string &filename)
 
         // Add segment heading to array
         m_Headings.push_back(atan2(makeM(segmentStart[1] - segmentEnd[1]),
-                                   makeM(segmentEnd[0] - segmentStart[0])));
+                                   makeM(segmentStart[0] - segmentEnd[0])));
 
         // Calculate segment length and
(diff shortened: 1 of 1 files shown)


alexdewar:
I'm not sure about this either (I haven't entirely wrapped my head around the various different frames of reference we're using here). But the headings definitely look wrong if I plot them and they look right with this patch, so I feel there's something going on here that's not quite right.

neworderofjamie:
But do the renders from ant world db creator still look right?

alexdewar:
The renders kinda look right in both versions :-/

neworderofjamie:
Dumb question, but are you definitely rebuilding libantworld? :smile:

neworderofjamie:
Thinking about this a little more - what you're doing is swapping whether the angle is taken from the positive or the negative x axis. As you're turning it back into x and y anyway you should just be able correct for it when you plot by changing the sign of your cos term?

alexdewar:
Finally figured what I was doing wrong here so I'll close without merging :-p

neworderofjamie:
Do tell!

PR #55, opened 2018-10-04
InfoMax implementation in Eigen

alexdewar:
Right, so I think this is just about ready to merge (except for some annoying minor compile error on Ubuntu which I'm tracking down now). You need Eigen 3 (libeigen3-dev) and OpenMP for this to work.

There is an example called ``infomax_route_example`` which lets you train and recapitulate an image database (e.g. one of the route databases output by ``ant_world_db_creator``) and outputs a (now working!) quiver plot.

I also updated the image database code so there's now a single class for reading and writing image databases. You can use the class like an STL-like container to get position etc. and load images.


third_party/matplotlibcpp.h
```diff
@@ -52,6 +52,8 @@ struct _interpreter {
     PyObject *s_python_function_axis;
     PyObject *s_python_function_xlabel;
     PyObject *s_python_function_ylabel;
+    PyObject *s_python_function_xticks;
```
neworderofjamie: We should probably be nice OSSers and pull request some of these changes into their repo.

navigation/image_database.h
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        cv::Mat loadGreyscale() const
+        {
+            return cv::imread(path, cv::IMREAD_GRAYSCALE);;
+        }
+    };
+
+    template<class T>
+    ImageDatabase(T &&databaseName)
```
neworderofjamie: Is this templating really necessary? Can't you just take a const reference to a path and assume that filesystem::path provides sensible implicit constructors?

navigation/image_database.h
```diff
@@ -0,0 +1,173 @@
...
+#include <iostream>
+#include <string>
+#include <vector>
+
+namespace BoBRobotics {
+namespace Navigation {
+using namespace units::length;
+using namespace units::angle;
+
+class ImageDatabase
```
neworderofjamie: This class is an excellent idea

navigation/image_database.h
```diff
@@ -0,0 +1,173 @@
...
+            filesystem::create_directory(m_Path);
+            return;
+        }
+
+        // Otherwise parse metadata file
+        std::ifstream file(metadataPath.str());
+        std::string line;
+        std::vector<std::string> fields;
+        std::getline(file, line); // skip first line
+        while (!file.eof()) {
```
neworderofjamie: You can use ``std::getline`` to do the csv parsing bit too:
```c++
std::string lineString;
std::string fieldString;
while(std::getline(file, lineString)) {
    std::stringstream lineStream(lineString);
    fields.clear();
    while(std::getline(lineStream, fieldString ',')) {
        fields.push_back(fieldString);
    }
    
    //**TODO** turn fields into Entry
}
```

navigation/image_database.h
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    const filesystem::path &getPath() const { return m_Path; }
+
+    const Entry &operator[](size_t i) const { return m_Database[i]; }
+    auto begin() const { return m_Database.cbegin(); }
+    auto end() const { return m_Database.cend(); }
+    size_t size() const { return m_Database.size(); }
+
+    void saveImage(cv::Mat &frame, millimeter_t x, millimeter_t y, millimeter_t z, degree_t heading, bool isRoute)
```
neworderofjamie: Maybe this would be better named ``addImage``?

navigation/infomax.h
```diff
@@ -0,0 +1,187 @@
...
+
+        // const auto newsd = matrixSD(weights);
+        // std::cout << "New SD" << std::endl
+        //           << newsd << std::endl;
+
+        return weights.transpose();
+    }
+
+private:
+    size_t m_SnapshotCount = 0;
```
neworderofjamie: It doesn't **look** like ``m_SnapshotCount`` is being used

navigation/infomax.h
```diff
@@ -0,0 +1,187 @@
...
+        });
+
+        const auto el = std::min_element(outputs.begin(), outputs.end());
+        size_t bestIndex = std::distance(outputs.begin(), el);
+        if (bestIndex > outputs.size() / 2) {
+            bestIndex -= outputs.size();
+        }
+        const radian_t heading = units::make_unit<turn_t>((double) bestIndex / (double) outputs.size());
+
+        return std::make_tuple(heading, *el, std::move(outputs));
```
neworderofjamie: I don't think you need to explicitly move here - you're returning outputs and never using it again so I feel the compiler should figure it out

alexdewar:
Ugh now I'm getting linking errors for OpenCV and python on Ubuntu. Dunno why.

navigation/infomax.h
```diff
@@ -0,0 +1,187 @@
...
+    {
+        assert(image.type() == CV_8UC1);
+
+        const cv::Size &unwrapRes = getUnwrapResolution();
+        assert(image.cols == unwrapRes.width);
+        assert(image.rows == unwrapRes.height);
+
+        // Convert image to vector of floats
+        const auto u = m_Weights * getFloatVector(image);
+        const auto y = tanh(u.array());
```
neworderofjamie: I think Eigen does some cunning stuff to reduce temporary allocations, but I'm guessing there's still going to be some sizable temporary allocations here. It might be good to make ``u``and ``y`` members.

navigation/visual_navigation_base.h
```diff
@@ -40,25 +40,25 @@ class VisualNavigationBase
...
-        for (size_t i = 0;; i++) {
-            const auto filename = routePath / getRouteDatabaseFilename(i);
-            if (!tryTrain(filename, resizeImages)) {
-                return;
+        cv::Mat image;
+        if (resizeImages) {
+            cv::Mat imageResized;
+            for (const auto &e : imdb) {
+                image = e.loadGreyscale();
+                assert(image.type() == CV_8UC1);
```
neworderofjamie: I think these asserts are unnecessary - the load greyscale function should return a greyscale image!

examples/infomax/infomax_route_example.cc
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    // Number of times each image is presented to network
+    int numReps = (argc > 2) ? std::stoi(argv[2]) : 1;
+    const filesystem::path netPath = "trained_network_"s + routePath.filename()
+            + "_reps"s + std::to_string(numReps) + ".bin"s;
+
+    // If we already have a network for these params, load from disk
+    if (netPath.exists()) {
+        std::cout << "Loading weights from " << netPath << std::endl;
```
neworderofjamie: I think this loading and saving logic belongs in the infomax class

examples/infomax/read_data.h
```diff
@@ -0,0 +1,43 @@
...
+#include <stdexcept>
+
+namespace {
+using namespace Eigen;
+using namespace std::literals;
+
+// Read a matrix of type T from a binary file
+template<typename T = double>
+auto
+readData(const filesystem::path &filepath)
```
neworderofjamie: As I said above - I feel this should just be an infomax constructor

neworderofjamie:
I fixed that for you - G++ arguments in the wrong order - I'm guessing
GNU made their argument parsing less terrible in later versions

alexdewar:
Thanks :-)

third_party/matplotlibcpp.h
alexdewar: I've put in a PR

neworderofjamie: :+1:

navigation/image_database.h
alexdewar: Yep, we could do. But I put the templating in so that one can do:
```
void myfunc(const ImageDatabase &db) { ... }

myfunc("mydatabase/");
```

I could/should just have string + const char * ctors for this though to avoid excessive templating (which is one of my many vices!)

navigation/image_database.h
alexdewar: Oh cool! That's good to know.

navigation/image_database.h
alexdewar: Good shout.

navigation/infomax.h
alexdewar: Possibly. But it can't hurt and we def don't want to copy construct the vector again:
https://stackoverflow.com/questions/39514150/function-return-a-tuple-made-of-vectors

navigation/infomax.h
alexdewar: Good idea

navigation/visual_navigation_base.h
alexdewar: True!

examples/infomax/infomax_route_example.cc
alexdewar: Ok good idea

navigation/visual_navigation_base.h
alexdewar: ...although on checking it, it's not the greyscale-ness we're checking here, but whether or not the image is of type uint8_t

examples/infomax/infomax_route_example.cc
alexdewar: Although actually I am using this function to do things other than load weights. I use it in infomax_matlab_comparison to check the values of u and y etc.

navigation/visual_navigation_base.h
neworderofjamie: Fine point - ``cv::imread`` looks like it could theoretically return some HDR weirdness

navigation/infomax.h
neworderofjamie: damnit! after reading https://shaharmike.com/cpp/move-semantics/ I thought I better understood move semantics but that stack overflow suggests that moving is indeed necessary

PR #57, opened 2018-10-05
A very basic generic robot class

alexdewar:
I've made a basic generic ``Robot`` class for ``Tank`` and ``UAV`` to inherit from. It provides the functions ``moveForward()``, ``turnOnTheSpot()`` and ``stopMoving()``, with the idea that we can use it for when we want a model to control a robot in a simple way. (We could also have virtual agents with similar functionality, but as it's only applied to robots for now I called it ``Robot``.)

I've also added some ``BOB_ASSERT()``s to check the input parameters for various ``tank()`` functions. The reason I did this instead of clamping the values between -1 and 1 is because I think it would be useful to know when we're giving bad parameters, because it probably means we've done something wrong somewhere!

Whaddya think?

### DIFF
--- a/projects/stone_cx/robotCommon.cc
+++ b/projects/stone_cx/robotCommon.cc
@@ -1,6 +1,7 @@
 #include "robotCommon.h"
 
 // Standard C++ includes
+#include <algorithm>
 #include <iostream>
 #include <numeric>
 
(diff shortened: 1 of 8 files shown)


neworderofjamie:
I think that makes sense - making the user code do the clamping if it's required is perfectly fair. Could you update the Stone CX code (I think just https://github.com/BrainsOnBoard/bob_robotics/blob/master/projects/stone_cx/robotCommon.cc#L24) to do this cos otherwise I'll try and use it in many months time and it'll crash and I'll have forgotten why :smile:

alexdewar:
Sure!

PR #58, opened 2018-10-08
Integrating Ardin MB example with BoB robotics Navigation

neworderofjamie:
Rather than using its own classes to abstract away the differences between perfect memory and the mushroom body, this refactor wraps the mushroom body model in a class derives from ``BoBRobotics::Navigation::VisualNavigationBase``. Additionally this PR:
* Tidies up the logic using ``BoBRobotics::Common::FSM``
* Tweaked the MB model slightly
* Added functionality to (slowly) render an (ugly) vector field
* Tidied up the bound-calculation in ``BoBRobotics::AntWorld::World`` so it's all done in ``units::length::meter_t``


libantworld/route_ardin.cc
```diff
@@ -191,6 +192,21 @@ bool RouteArdin::load(const std::string &filename, bool realign)
         }
     }
 
+    // Initialise bounds to limits of underlying data types
+    std::fill_n(&m_MinBound[0], 2, meter_t(std::numeric_limits<meter_t::underlying_type>::max()));
+    std::fill_n(&m_MaxBound[0], 2, meter_t(std::numeric_limits<meter_t::underlying_type>::min()));
```
alexdewar: I think you might be able to just write:
```
std::fill_n(&m_MinBound[0], 2, std::numeric_limits<meter_t>::max());
```
etc. The units lib has overloads for some numeric_limits functions

libantworld/world.cc
```diff
@@ -124,33 +127,28 @@ bool World::load(const std::string &filename, const GLfloat (&worldColour)[3],
...
 
-        // Convert min/max bounds to metre types (this is admittedly bit gross...)
-        m_MinBoundM[0] = makeM(m_MinBound[0]);
-        m_MinBoundM[1] = makeM(m_MinBound[1]);
-        m_MinBoundM[2] = makeM(m_MinBound[2]);
-        m_MaxBoundM[0] = makeM(m_MaxBound[0]);
-        m_MaxBoundM[1] = makeM(m_MaxBound[1]);
-        m_MaxBoundM[2] = makeM(m_MaxBound[2]);
+        std::cout << "Min: (" << m_MinBound[0] << ", " << m_MinBound[1] << ", " << m_MinBound[2] << ")" << std::endl;
+        std::cout << "Max: (" << m_MaxBound[0] << ", " << m_MaxBound[1] << ", " << m_MaxBound[2] << ")" << std::endl;
```
alexdewar: This is all much tidier!

projects/ardin_mb/vector_field.cc
```diff
@@ -0,0 +1,121 @@
+#include "vector_field.h"
```
alexdewar: Another (possibly easier but less flexible) option to doing this in OpenGL would be to use matplotlib-cpp's ``quiver()`` function

alexdewar:
This is all looking good and much tidier :-)

projects/ardin_mb/vector_field.cc
neworderofjamie: Yeah - the problem is that I want to overlay the result on the OpenGL map view.

I guess theoretically we could draw that whole view using matplotlib but I can see that turning into a shit show.

alexdewar: That's true. Plus I suspect you'd have to add your own wrapper function for rendering the triangles anyway.

PR #59, opened 2018-10-09
Random, hopefully uncontroversial changes

alexdewar:
These are a bunch of mostly robot-related changes I've made while hacking away at other stuff -- I've been trying to keep the generically useful and uncontroversial stuff on its own branch.

Summary of main commits:
* Small feature additions to Bebop module
* A basic image database collector for the gantry
* Stop the ImageDatabase overriding existing files by default
* Add a ``clearMemory()`` function to the navigation algos
* Option to have ``Video::Display`` as fullscreen or not, defaulting to not
* Change minimum version of gcc to 5.x (things weren't compiling for me with 4.9 and I can't be bothered to fix it)


navigation/image_database.h
```diff
@@ -121,7 +121,14 @@ class ImageDatabase
         m_Database.push_back(Entry{ { x, y, z }, heading, filename });
 
         // Write current view to file
-        cv::imwrite((filesystem::path(m_Path) / filename).str(), frame);
+        const auto imagePath = m_Path / filename;
+        assert(!imagePath.exists()); // We don't want to overwrite data by default!
```
neworderofjamie: BOB_ASSERT?

robots/gantry.h
```diff
@@ -199,7 +215,8 @@ class Gantry
 private:
     BYTE m_BoardId;
     bool m_IsMovingLine = false;
-    static constexpr std::array<double, 3> PulsesPerMillimetre = { 7.49625, 8.19672, 13.15789 };
+    static constexpr Vector3<double> PulsesPerMillimetre = { 7.49625, 8.19672, 13.15789 };
```
neworderofjamie: Could you maybe add a comment explaining the source of these magic numbers incase someone else ever has to assume your role as chief gantry-wrangler?

neworderofjamie:
I think GCC 4.9 only supports a very small subset of C++14 so not supporting it makes total sense

navigation/image_database.h
alexdewar: Good spot!

robots/gantry.h
alexdewar: Sure, I'll stick a comment in.

The source is Chris Johnson's Matlab script. I have to assume he measured these values empirically, but they do seem to be relatively accurate.

alexdewar:
Cheers for looking at this :-)

PR #60, opened 2018-10-09
Vicon plotting and API changes

alexdewar:
The main point of this PR is to add a basic (2D) matplotlib plotter for the Vicon system.

But as I was working on it I decided that we might be better off changing the ``getObjectData()`` function to return the objects by reference rather than value. ``getPosition()`` and ``getAttitude()`` already copy the data from the object and it seems a bit unnecessary to copy everything twice. I've also stuck a mutex in ``ObjectData`` to stop data races (and this meant I had to add a copy constructor). As we're now returning a reference to the object rather than just its data I thought I'd made sense to rename all the ``ObjectData``s etc. to just ``Object``s.

### DIFF
--- a/common/plot_agent.h
+++ b/common/plot_agent.h
@@ -1,0 +1,28 @@
+#pragma once
+
+// Third-party includes
+#include "../third_party/matplotlibcpp.h"
+#include "../third_party/units.h"
+
+// Standard C++ includes
+#include <array>
+
+namespace BoBRobotics {
+    template<typename AgentType>
+    void plotAgent(const AgentType &agent, std::array<double, 2> &&xlim, std::array<double, 2> &&ylim)
+    {
+        // Update plot
+        const auto position = agent.template getPosition<>();
+        const auto attitude = agent.template getAttitude<>();
+        const std::vector<double> vx{ position[0].value() };
+        const std::vector<double> vy{ position[1].value() };
+        const std::vector<double> vu{ units::math::cos(attitude[0]) };
+        const std::vector<double> vv{ units::math::sin(attitude[0]) };
+        const std::vector<int> x0 { 0 }, y0 { 0 };
+        matplotlibcpp::clf();
+        matplotlibcpp::plot(x0, y0, "r+");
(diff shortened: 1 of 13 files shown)


alexdewar:
I take what you're saying about the matplotlib interface -- I don't particularly mind if we do the plotting with an OpenCV window or something instead.

The problem with copying the ``ObjectData`` classes every time we want data though is that there could (presumably?) be a data race where the ``m_Position`` or ``m_Attitude`` arrays are being written to with incoming data while they're being copied (regardless of whether that's because we're copying the entire ``ObjectData`` object or just reading out these arrays).

I did put most of the actual plotting code in ``plot_agent.h``. The only purpose of the ``ObjectPlotter`` class is so that we can set a flag when there is incoming data and then only refresh the plot when we need to (which I think would still be useful if we were plotting in OpenCV).

What do you think?

alexdewar:
...although I've just checked and the Vicon does seem to be giving us data at a much faster rate than we're plotting it any way. So we could just drop the ``ObjectPlotter`` class.

neworderofjamie:
Unless I'm missing something the ``Vicon::UDP::m_ObjectDataMutex`` means that object data is either being copied (using ``Vicon::UDP::getObjectData()``) or being updated (using ``Vicon::UDP::updateObjectData()``)

neworderofjamie:
Yeah - I think the Vicon is always going to give us data faster than we could do very much with it (especially if that is being done via python)

alexdewar:
Ah yeah I guess so. Hmm. How about we drop everything except plot_agent.h then? I know using matplotlib probably isn't ideal but it could be useful as a stopgap. Plus it does give us scope for easily doing things like plotting trajectories etc.

neworderofjamie:
Sure! Sorry for shitting on that somewhat :blush:

neworderofjamie:
(Damnit github - that wasn't supposed to be a **happy** blush)

alexdewar:
Haha nw :-p. You were def right about it

PR #61, opened 2018-10-10
Minor changes and fixes

alexdewar:
Hi Jamie,

The main point of this PR is to fix a small bug I found in the ``Tank::drive()`` function: the value of ``r`` should be divided by ``sqrt(2)``, otherwise we get out-of-range motor values and now with the new ``BOB_ASSERT()``s this breaks things :-)

I've also included some plotting code and a fix for the problem where ``KEY_UP`` and ``KEY_DOWN`` macros were breaking OpenCV as well as some other small bits.

### DIFF
--- a/common/plot_agent.h
+++ b/common/plot_agent.h
@@ -1,0 +1,28 @@
+#pragma once
+
+// Third-party includes
+#include "../third_party/matplotlibcpp.h"
+#include "../third_party/units.h"
+
+// Standard C++ includes
+#include <array>
+
+namespace BoBRobotics {
+    template<typename AgentType>
+    void plotAgent(const AgentType &agent, std::array<double, 2> &&xlim, std::array<double, 2> &&ylim)
+    {
+        // Update plot
+        const auto position = agent.template getPosition<>();
+        const auto attitude = agent.template getAttitude<>();
+        const std::vector<double> vx{ position[0].value() };
+        const std::vector<double> vy{ position[1].value() };
+        const std::vector<double> vu{ units::math::cos(attitude[0]) };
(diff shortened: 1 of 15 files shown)


alexdewar:
I've decided it makes more sense to just cap the value of ``r`` at one. Otherwise when you push the joystick straight up, the motors will only go at ``1/sqrt(2)`` of maximum speed.

common/plot_agent.h
```diff
@@ -0,0 +1,28 @@
...
+// Third-party includes
+#include "../third_party/matplotlibcpp.h"
+#include "../third_party/units.h"
+
+// Standard C++ includes
+#include <array>
+
+namespace BoBRobotics {
+    template<typename AgentType>
+    void plotAgent(const AgentType &agent, std::array<double, 2> &&xlim, std::array<double, 2> &&ylim)
```
neworderofjamie: How come you're passing these as an rvalue reference? isn't this just a case for ``const std::array<double, 2> &xlim``?

hid/joystick_linux.h
```diff
@@ -1,24 +1,35 @@
...
+/*
+ * We need to undef these macros (defined in linux/joystick.h) because they
+ * break OpenCV.
+ */
+#ifdef KEY_UP
+#undef KEY_UP
+#endif
+#ifdef KEY_DOWN
+#undef KEY_DOWN
+#endif
```
neworderofjamie: Nice work figuring out where this was coming from!

common/plot_agent.h
alexdewar: Hmm yeah it probably should be :-/

PR #62, opened 2018-10-12
More random updates

alexdewar:
Hi Jamie,

Here's another set of random cleanups and fixes (I really need to start working on one thing, dammit!).

Commit summary:
* Threading tidy-ups and fixes
* Tidy-up and commenting of the cross platform networking code in os/net.h
* Extra minor features for Bebop
* Other minor fixes

### DIFF
--- a/common/threadable.h
+++ b/common/threadable.h
@@ -1,6 +1,7 @@
 #pragma once
 
 // Standard C++ includes
+#include <atomic>
 #include <memory>
 #include <thread>
 
@@ -38,8 +39,7 @@ public:
     //! Stop the background thread
     virtual void stop()
     {
-        m_DoRun = false;
-        if (m_ThreadRunning) {
+        if (m_DoRun.exchange(false) && m_ThreadRunning) {
             if (m_Thread.joinable()) {
                 m_Thread.join();
             } else {
@@ -54,6 +54,6 @@ private:
     bool m_ThreadRunning = false;
 
 protected:
-    bool m_DoRun = true;
+    std::atomic<bool> m_DoRun{ true };
 };
 }
--- a/common/threadable.h
+++ b/common/threadable.h
@@ -32,26 +32,23 @@ public:
     //! Run the process on a background thread
     virtual void runInBackground()
     {
-        m_Thread = std::thread([this] { run(); });
-        m_ThreadRunning = true;
+        m_Thread = std::make_unique<std::thread>([this] { run(); });
     }
 
     //! Stop the background thread
     virtual void stop()
     {
-        if (m_DoRun.exchange(false) && m_ThreadRunning) {
-            if (m_Thread.joinable()) {
-                m_Thread.join();
+        if (m_DoRun.exchange(false) && m_Thread) {
+            if (m_Thread->joinable()) {
+                m_Thread->join();
             } else {
-                m_Thread.detach();
+                m_Thread->detach();
             }
-            m_ThreadRunning = false;
         }
     }
 
 private:
-    std::thread m_Thread;
-    bool m_ThreadRunning = false;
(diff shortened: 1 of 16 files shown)


common/threadable.h
```diff
@@ -28,32 +29,39 @@ class Threadable
...
+    {
+        return m_DoRun;
+    }
+
     //! Run the process on a background thread
     virtual void runInBackground()
     {
-        m_Thread = std::thread([this] { run(); });
-        m_ThreadRunning = true;
+        m_Thread = std::make_unique<std::thread>([this] { run(); });
```
neworderofjamie: What is the purpose of moving ``m_Thread`` to the heap?

common/threadable.h
```diff
@@ -28,32 +29,39 @@ class Threadable
...
 
     //! Stop the background thread
     virtual void stop()
     {
         m_DoRun = false;
-        if (m_ThreadRunning) {
-            if (m_Thread.joinable()) {
-                m_Thread.join();
+        if (m_Thread) {
+            if (m_Thread->joinable()) {
```
neworderofjamie: ``m_Thread->joinable()`` "Checks if the thread object identifies an active thread of execution." - if it doesn't shouldn't you just leave it alone rather than detach it? (which presumably only makes sense if it's also an active thread of execution - in which case you probably don't want to detach it anyway)

examples/bebop/bebop_test.cc
```diff
@@ -1,14 +1,14 @@
...
-#include <iostream>
-#include <stdexcept>
-#include <thread>
-
 // BoB robotics includes
 #include "hid/joystick.h"
 #include "libbebop/bebop.h"
 #include "video/display.h"
 
+// Standard C++ includes
```
neworderofjamie: How come you re-ordered this? If it's to overcome some inclusion-order horror maybe add a comment.

common/threadable.h
```diff
@@ -28,32 +29,39 @@ class Threadable
     //! Run on the current thread, blocking until process ends
     virtual void run() = 0;
 
+    //! Check if the run() function has been called
+    virtual bool isRunning()
```
neworderofjamie: Would it maybe be better to make the virtual function protected and make ``run()`` set the flag and call the internal virtual? The current approach seems a little error-prone as I can easily imagine forgetting to call ``runStart()`` if I was writing a Threadable.

common/threadable.h
alexdewar: It's not that I wanted ``m_Thread`` on the heap especially. I just made it a pointer so it could be ``nullptr`` and then we don't need a separate flag variable

common/threadable.h
alexdewar: I *believe* this is what you're supposed to do with threads -- either join them or detach them when you're finished with them. Maybe it's so you can free the resources associated with it? Not sure: https://en.cppreference.com/w/cpp/thread/thread/detach

It's what my C++ book suggests in any case and I was just blindly following it ;-)

examples/bebop/bebop_test.cc
alexdewar: Because I read somewhere that it's good to have your own includes at the top, so that if e.g. hid/joystick.h implicitly depends on one of the standard C++ includes you have underneath you'll get an error. Not sure if it's really a good idea or not

common/threadable.h
alexdewar: That's a much better idea!

alexdewar: Heh. In fixing this up I already found somewhere where I forgot to call ``runStart()`` :-p

common/threadable.h
alexdewar: It looks like you're right. I double-checked in the book and it turns out I was misremembering it :-/

PR #63, opened 2018-10-13
Random video input device + tidy-ups for Video::Input code

alexdewar:
This is a tiny PR that mostly just adds a ``Video::Input`` device that just returns random white noise with a specified image size, for debugging etc. (I was having trouble using my crappy old webcam with OpenCV and adding this was easier!)

I've also included some random misc tidy-ups for ``Video::Input`` classes.

### DIFF
--- a/video/opengl.h
+++ b/video/opengl.h
@@ -20,10 +20,10 @@ class OpenGL : public Input
 public:
     /*!
      * \brief Create a Video::Input for reading from an OpenGL window
-     * 
+     *
      * @param readX, readY The starting coordinates to read from (from bottom left of screen)
      * @param readWidth, readHeight Size of image
-     * 
+     *
      * **NOTE** intentionally NOT using a cv::Rect as OpenGL coordinates are (typically)
      * from bottom left of screen which would require window size etc. to convert
      */
@@ -34,7 +34,7 @@ public:
     //----------------------------------------------------------------------------
     // Input virtuals
     //----------------------------------------------------------------------------
-    virtual const std::string getCameraName() const override
+    virtual std::string getCameraName() const override
     {
         return "opengl";
     }
@@ -55,7 +55,7 @@ public:
         // **TODO** it should be theoretically possible to go directly from frame buffer to GpuMat
         glReadPixels(m_ReadX, m_ReadY, m_ReadWidth, m_ReadHeight,
                      GL_BGR, GL_UNSIGNED_BYTE, outFrame.data);
-        
+
         // Flip image vertically
         cv::flip(outFrame, outFrame, 0);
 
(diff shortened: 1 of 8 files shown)


video/randominput.h
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+private:
+    cv::Size m_Size;
+    std::string m_CameraName;
+    std::random_device m_RandomDevice;
+    std::mt19937 m_Generator;
+    std::uniform_int_distribution<uchar> m_Distribution;
+
+    void fillRandom(uchar *start, uchar *end)
```
neworderofjamie: Perhaps unnecessary C++ary but:
``std::generate(start, end, [this](){ return m_Distribution(m_Generator); });``

video/randominput.h
```diff
@@ -0,0 +1,67 @@
...
+// OpenCV
+#include <opencv2/opencv.hpp>
+
+// Standard C++ includes
+#include <random>
+
+namespace BoBRobotics {
+namespace Video {
+//! A Video::Input which generates white noise, for testing purposes
+class RandomInput : public Input {
```
neworderofjamie: As it's header-only anyway, you **could** template on the type of generator.

video/opengl.h
```diff
@@ -46,16 +46,14 @@ class OpenGL : public Input
 
     virtual bool readFrame(cv::Mat &outFrame) override
     {
-        // If outFrame is not of the expected size or format, recreate it
-        if (outFrame.cols != m_ReadWidth || outFrame.rows != m_ReadHeight || outFrame.type() != CV_8UC3) {
-            outFrame.create(m_ReadHeight, m_ReadWidth, CV_8UC3);
-        }
+        // Make sure frame is of right size and type
+        outFrame.create(m_ReadWidth, m_ReadHeight, CV_8UC3);
```
neworderofjamie: I was initially paranoid/suspicious of this, but it's clear from the docs that this is a good idea as create has exactly this behaviour :smile:

video/randominput.h
alexdewar: Hey, finding an opportunity to use STL algorithms is what C++ is about! This is def nicer

video/randominput.h
alexdewar: That's true. I did think about this, but the only thing is that I can't seem to find any other generators that would work here (i.e. return ``uchar``s in the right range). I guess you could provide a custom generator though?

video/opengl.h
alexdewar: Yeah, I only just discovered this :-)

video/randominput.h
neworderofjamie: ``std::uniform_int_distribution<uchar>`` is totally the only distribution that makes sense but different RNGs could plausible be useful (basically cos MT is slow)

video/randominput.h
neworderofjamie: That is totally my understanding of Herb Sutter's intentions :smile:

PR #64, opened 2018-10-15
Networking fixes

alexdewar:
I was trying to use the networking API the other day to do some work with a Norbot and found that it seemed to be a lot more fragile than I remembered it being :confused:. So I've now done a big tidy-up mostly for the sake of doing better locking and fixing up the clean-up code so that we can shut down gracefully, whether there are errors or not.

I've added an optional interface for catching exceptions on background threads so they can be rethrown on the main one. This is mostly just to make things cleaner, but could also help stop runaway robots :-). The other reason is because I've added a new exception (``SocketClosedError``) which is thrown when trying to use a socket that we've closed on purpose, so we can ignore it and then do clean up.

It's taken a depressingly long time to fix this (partly because some of the segfaults I was getting turned out to be because of a bug in my version of OpenCV :disappointed:) but hopefully it's vaguely useful!

### DIFF
--- a/common/background_exception.h
+++ b/common/background_exception.h
@@ -39,6 +39,13 @@ private:
     static std::exception_ptr m_Exception;
 }; // BackgroundException
 
+/*
+ * We want to *not* include definitions when we're linking against another
+ * object file. This currently only seems to be a problem when linking with
+ * libbebop.
+ */
+#ifndef NO_BACKGROUND_EXCEPTION_DEFINITIONS
 bool BackgroundException::m_DoCatch = false;
 std::exception_ptr BackgroundException::m_Exception;
+#endif
 } // BoBRobotics
--- a/common/background_exception.h
+++ b/common/background_exception.h
@@ -5,7 +5,7 @@
 
 namespace BoBRobotics {
 //! A wrapper for passing exceptions between threads (i.e. a background thread to the main one)
-class GlobalException {
+class BackgroundException {
 public:
     //! Sets the global exception
     static void set(const std::exception_ptr &error)
@@ -23,7 +23,7 @@ public:
 
 private:
     static std::exception_ptr m_Exception;
-}; // GlobalException
+}; // BackgroundException
 
-std::exception_ptr GlobalException::m_Exception;
+std::exception_ptr BackgroundException::m_Exception;
 } // BoBRobotics
--- a/common/background_exception.h
+++ b/common/background_exception.h
@@ -4,13 +4,26 @@
 #include <stdexcept>
 
 namespace BoBRobotics {
+//----------------------------------------------------------------------------
+// BoBRobotics::BackgroundException
+//----------------------------------------------------------------------------
 //! A wrapper for passing exceptions between threads (i.e. a background thread to the main one)
 class BackgroundException {
 public:
+    //! Start catching exceptions on background threads
+    static void enableCatching()
+    {
+        m_DoCatch = true;
+    }
+
     //! Sets the global exception
     static void set(const std::exception_ptr &error)
     {
-        m_Exception = error;
+        if (m_DoCatch) {
+            m_Exception = error;
+        } else {
+            std::rethrow_exception(error);
+        }
     }
 
     //! Checks if the global exception is set and rethrows it if so
@@ -22,8 +35,10 @@ public:
     }
 
 private:
+    static bool m_DoCatch;
     static std::exception_ptr m_Exception;
 }; // BackgroundException
 
+bool BackgroundException::m_DoCatch = false;
 std::exception_ptr BackgroundException::m_Exception;
 } // BoBRobotics
--- a/examples/robot_net/computer.cc
+++ b/examples/robot_net/computer.cc
@@ -56,18 +56,20 @@ main(int argc, char **argv)
         Video::NetSource video(client);
 
         // transmit motor commands over network
-        Robots::TankNetSink motor(client);
+        Robots::TankNetSink tank(client);
 
         // add joystick for controlling Tank
         HID::Joystick joystick;
-        motor.addJoystick(joystick); // send joystick events to motor
+        tank.addJoystick(joystick); // send joystick events to robot
 
         // display video stream
         Video::Display display(video, { 1240, 500 });
 
         // poll joystick and video stream repeatedly
         do {
-            if (!joystick.update() && !display.update()) {
+            bool joystickUpdate = joystick.update();
+            bool displayUpdate = display.update();
+            if (!joystickUpdate && !displayUpdate) {
                 std::this_thread::sleep_for(50ms);
             }
 
--- a/examples/robot_net/computer.cc
+++ b/examples/robot_net/computer.cc
@@ -50,6 +50,9 @@ main(int argc, char **argv)
 
         // Make connection to robot on default port
         Net::Client client(robotIP);
+
+        // Run client on background thread, catching any exceptions for rethrowing
+        BackgroundException::enableCatching();
         client.runInBackground();
 
         // Read video stream from network
@@ -65,7 +68,7 @@ main(int argc, char **argv)
         // Display video stream on screen
         Video::Display display(video, { 1240, 500 });
         while(display.isOpen()) {
-            // Check for exceptions on background thread
+            // Rethrow any exceptions caught on background thread
             BackgroundException::check();
 
             // Poll joystick and camera for updates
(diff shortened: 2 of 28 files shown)


os/net.h
```diff
@@ -17,10 +17,21 @@
 #include <winsock2.h>
 #include <ws2tcpip.h>
 
+// Standard C includes
+#include <cwchar>
+
 // Macros defined in Linux but not Windows
 #define MSG_NOSIGNAL 0
```
neworderofjamie: I think defining something that is an actual flag on one platform as 0 on another is a little confusing. As you only use it once maybe this would be nicer:
```c++
// Turn of broken pipe signalling (which abruptly kills application) on POSIX
#ifdef _WIN32
const int sendFlags = 0;
#else
const int sendFlags = MSG_NOSIGNAL;
#endif
::send(...);
```

robots/tank_netsink.h
```diff
@@ -31,16 +31,19 @@ class TankNetSink : public Tank
         }
 
         // send steering command
-        m_Node->getSocket()->send("TNK " + std::to_string(left) + " " +
-                                  std::to_string(right) + "\n");
+        {
+            std::lock_guard<Net::Node> guard(m_Node);
```
neworderofjamie: I think leaving the locking of this up to users of ``Net::Node`` is a bit error-prone. It would be nicer if you made ``Node::getSocket()`` private and added wrappers with ``std::lock_guard`` around the ``Net::Socket`` functions that users of ``Net::Node`` actually should be using (i.e. send and receive) to ``Net::Node``

video/display.h
```diff
@@ -156,6 +144,14 @@ class Display : public Threadable
             }
         }
     }
+
+private:
+    cv::Mat m_Frame, m_Unwrapped;
+    bool m_ShowUnwrapped = false;
+    std::unique_ptr<ImgProc::OpenCVUnwrap360> m_Unwrapper;
+    Input &m_VideoInput;
+    bool m_FullScreen;
```
neworderofjamie: It doesn't look like ``m_FullScreen`` is actually used.

video/display.h
```diff
@@ -86,20 +91,11 @@ class Display : public Threadable
...
-            // set opencv window to display full screen
-            cv::namedWindow(WINDOW_NAME, CV_WINDOW_NORMAL);
-            if (m_FullScreen) {
-                setWindowProperty(WINDOW_NAME, cv::WND_PROP_FULLSCREEN, cv::WINDOW_FULLSCREEN);
-            }
-            m_Open = true;
-        }
-
         cv::Mat frame;
         bool newFrame = readFrame(frame);
```
neworderofjamie: Pretty sure this implementation of ``Display::readFrame`` will result in an allocation and copy every frame. Maybe reimplement as:
```c++
virtual bool readFrame(cv::Mat &frame)
{
    // unwrap frame if required
    if (m_Unwrapper && m_ShowUnwrapped) {
        if (!m_VideoInput.readFrame(m_IntermediateFrame)) {
            return false;
        }
        m_Unwrapper->unwrap(m_IntermediateFrame, frame);
    } 
    else {
        if (!m_VideoInput.readFrame(frame)) {
            return false;
        }
    }
    return true;
}
```
Where you always call it with ``readFrame(m_Frame)``

video/randominput.h
```diff
@@ -65,11 +66,11 @@ class RandomInput : public Input {
     cv::Size m_Size;
     std::string m_CameraName;
     GeneratorType m_Generator;
-    std::uniform_int_distribution<uchar> m_Distribution;
+    std::uniform_int_distribution<int> m_Distribution;
```
neworderofjamie: What was wrong with ``std::uniform_distribution<uchar>``?

common/background_exception.h
```diff
@@ -0,0 +1,51 @@
...
+
+// Standard C++ includes
+#include <stdexcept>
+
+namespace BoBRobotics {
+//----------------------------------------------------------------------------
+// BoBRobotics::BackgroundException
+//----------------------------------------------------------------------------
+//! A wrapper for passing exceptions between threads (i.e. a background thread to the main one)
+class BackgroundException {
```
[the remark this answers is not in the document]
alexdewar: It was only because I also use this class in ``Video::NetSink`` which isn't a ``Threadable``. I guess also because it's a general thing that we might want to use for other threaded code in future.

common/background_exception.h
```diff
@@ -0,0 +1,51 @@
...
+    static bool m_DoCatch;
+    static std::exception_ptr m_Exception;
+}; // BackgroundException
+
+/*
+ * We want to *not* include definitions when we're linking against another
+ * object file. This currently only seems to be a problem when linking with
+ * libbebop.
+ */
+#ifndef NO_BACKGROUND_EXCEPTION_DEFINITIONS
```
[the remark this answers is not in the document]
alexdewar: Yep! I know initialising static variables in headers is a no-no, but I thought it would be annoying to have to include these definitions in almost every .cc file we have. So I've just changed libbebop to make with ``-DNO_BACKGROUND_EXCEPTION_DEFINITIONS`` and this kinda gets round it. We'll just have to remember to do this for any other lib*s that we do in future which also use this class.

examples/robot_net/computer.cc
```diff
@@ -47,29 +48,36 @@ main(int argc, char **argv)
...
+            // Rethrow any exceptions caught on background thread
+            BackgroundException::check();
 
-        // poll joystick and video stream repeatedly
-        do {
-            if (!joystick.update() && !display.update()) {
+            // Poll joystick and camera for updates
+            bool joystickUpdate = joystick.update();
+            bool displayUpdate = display.update();
+            if (!joystickUpdate && !displayUpdate) {
```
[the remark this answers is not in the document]
alexdewar: Yeah, it also meant that the display often wasn't opening, so the loop would exit immediately and it would quit. I assumed it was some other networking bug for ages

robots/tank_netsink.h
alexdewar: I haven't made this particularly apparent, but I've changed the code so that we only do locking for sending now, as reading always takes place on the same thread and some of the handlers for incoming commands need to send stuff.

I did have it the way you suggested before, but the problem is that when we're sending and receiving frames we need to do two reads/writes in quick succession (one for the ASCII command, one for the raw JPEG bytes that come after) and sometimes we could end up with another read or write in the middle which would screw things up. I did think of a more C++-y way of forcing users to lock it: return a little object called ``LockedSocket`` or something, which just provides a ``get()`` method to get at the ``Socket*`` and locks and unlocks the ``Node``'s mutex on construction/destruction. What do you think?

video/display.h
alexdewar: Good thinking

video/randominput.h
alexdewar: MSVC won't let you have a ``std::uniform_distribution<uchar>`` and says it violates some standard or other (which is a bit odd because it seems useful)

video/display.h
alexdewar: Actually, on second thoughts, I don't think it will. I just googled it and it seems like if you assign to a matrix it'll only do a shallow copy (which is what I wanted): you have to explicitly call ``copyTo()`` or ``clone()`` if you want to duplicate the data.

common/background_exception.h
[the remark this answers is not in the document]
neworderofjamie: Ahh I didn't notice that - in which case this is the correct solution

robots/tank_netsink.h
alexdewar: I suppose the downside to this approach is that we'll probably end up with a bunch of unnecessary locking and unlocking of the mutex

neworderofjamie: Ahh yeah of course - in that case I think that what we have now is fine. Although it would be kinda cool if you hid the socket like I said previous and added a method which returned one of these as a kinda socket writing transaction object:

```c++
class SocketWriter
{
public:
    SocketWriter(Socket &socket, std::mutex &mutex) : m_Socket(socket), m_Guard(mutex)
    {
    }
    SocketWriter(const SocketWriter&) = delete;

    void write(...){ m_Socket.write(...); }

private:
    Socket &m_Socket;
    std::lock_guard<std::mutex> m_Guard;
};
```

video/display.h
neworderofjamie: Good to know :+1:

video/randominput.h
neworderofjamie: How irritating - https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution does agree that uchar is within the dread realm of undefined behaviour

PR #65, opened 2018-10-18
More image database work

alexdewar:
Hi Jamie,

I've just been doing some work on the image database code, changing a few things:
1. Recording extra metadata in a ``database_metadata.yaml`` file (database entries are still in a CSV file)
2. Changing to use ``GridRecorder`` and ``RouteRecorder`` classes for writing to image databases
3. Saving grid coordinates in addition to physical positions for grid databases

This is partly for making databases with the gantry (and transferring over old Matlab ones) and it would be useful to have a standardised way of recording the metadata for them, including parameters that we might want to be able to read in (e.g. unwrap parameters used etc.). But I'm hoping that this will also be useful for e.g. the library square data. If you're happy with the format I'll add an example of how to read this in in python to the repo for students and things.

The reason for saving the details about the grid size and spacing separately from the physical positions is because I found there were tiny floating point rounding errors when calculating the difference between e.g. the x positions, so it's much nicer to just have this parameter saved explicitly somewhere.

I reorganised ant_world_db_creator to use the new interface and I think it's fair to say that that part of the PR is a fairly dubious improvement :-/. There's a tidier example for the gantry:
https://github.com/BrainsOnBoard/bob_robotics/blob/more_image_database/tools/gantry_db_creator/gantry_db_creator.cc

### DIFF
--- a/video/opengl.h
+++ b/video/opengl.h
@@ -47,7 +47,7 @@ public:
     virtual bool readFrame(cv::Mat &outFrame) override
     {
         // Make sure frame is of right size and type
-        outFrame.create(m_ReadWidth, m_ReadHeight, CV_8UC3);
+        outFrame.create(m_ReadHeight, m_ReadWidth, CV_8UC3);
 
         // Read pixels from framebuffer into outFrame
         // **TODO** it should be theoretically possible to go directly from frame buffer to GpuMat
--- a/navigation/image_database.h
+++ b/navigation/image_database.h
@@ -57,7 +57,7 @@ struct Range
 
     size_t size() const
     {
-        return (separation == 0_mm) ? 1 : ((end - begin) / separation).to<size_t>();
+        return (separation == 0_mm) ? 1 : (1 + ((end - begin) / separation).to<size_t>());
     }
 };
 
--- a/navigation/image_database.h
+++ b/navigation/image_database.h
@@ -30,6 +30,7 @@ using namespace units::length;
 using namespace units::literals;
 using namespace units::math;
 
+//! A range of values in millimetres
 struct Range
 {
     millimeter_t begin;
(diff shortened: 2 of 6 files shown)


video/opengl.h
```diff
@@ -47,7 +47,7 @@ class OpenGL : public Input
     virtual bool readFrame(cv::Mat &outFrame) override
     {
         // Make sure frame is of right size and type
-        outFrame.create(m_ReadWidth, m_ReadHeight, CV_8UC3);
+        outFrame.create(m_ReadHeight, m_ReadWidth, CV_8UC3);
```
neworderofjamie: Good spot - I hate this about OpenCV

neworderofjamie:
First of all sorry for taking so long! I made a more comprehensive review but github is being shit so:
1) I think ``Range`` would be better if all its members were const and you added  a range(start, end, seperation) constructor containing the ``check()`` logic - that manual checking is a bit gross
2) In ``getFilename`` where you use a mixture of lambda functions, C string and C++ std::string - I think it would be nicer (and terser) to either use a single snprintf into a single buffer or std::stringstream (I **think** ``std::setw(7)`` and ``std::setfill('0')`` are the two incantations you need)
3) There seems to be a mix of ``size_t`` and ``int`` being used for indexing - it would be better to be consistent

All looks real cool though!

neworderofjamie:
First of all sorry for taking so long! I made a more comprehensive review but github is being shit so:
1) I think ``Range`` would be better if all its members were const and you added  a range(start, end, seperation) constructor containing the ``check()`` logic - that manual checking is a bit gross
2) In ``getFilename`` where you use a mixture of lambda functions, C string and C++ std::string - I think it would be nicer (and terser) to either use a single snprintf into a single buffer or std::stringstream (I **think** ``std::setw(7)`` and ``std::setfill('0')`` are the two incantations you need)
3) There seems to be a mix of ``size_t`` and ``int`` being used for indexing - it would be better to be consistent

All looks real cool though!

alexdewar:
Cheers for this!

PR #66, opened 2018-10-22
A very boring PR

alexdewar:
Hi Jamie,

There are a few small fixes here and some extra Bebop things, but this PR is mostly due to me getting distracted by warnings in my new (probably overzealous) IDE.

It did pick up a few places where we're using ``unsigned int`` and should be using ``int`` etc. though, so it's probably worth cleaning these places up. I've also done a small bit of rearranging of files.

Sorry this isn't more exciting!


neworderofjamie:
Oh no! I hope this isn't the prelude to a QT GUI for bob robotics..

libantworld/renderer.h
```diff
@@ -28,7 +28,7 @@ using namespace units::length;
 class Renderer
 {
 public:
-    Renderer(unsigned int cubemapSize = 256, double nearClip = 0.001, double farClip = 1000.0,
+    Renderer(GLsizei cubemapSize = 256, double nearClip = 0.001, double farClip = 1000.0,
```
neworderofjamie: I feel using ``unsigned int`` is a bad habit of mine (and one that Thomas shares so my chances of getting rid of it are slim)

libantworld/snapshot_processor_ardin.cc
```diff
@@ -30,38 +30,38 @@ void SnapshotProcessorArdin::process(const cv::Mat &snapshot)
...
+            int sum = 0;
+            for(int i = kernelStartY; i < kernelEndY; i++) {
+                for(int j = kernelStartX; j < kernelEndX; j++) {
                     sum += snapshot.at<cv::Vec3b>(i, j)[1];
                 }
             }
 
             // Divide sum by number of pixels in kernel to compute average and write to intermediate snapshot
-            m_IntermediateSnapshotGreyscale.at<uint8_t>(y, x) = sum / kernelPixels;
+            m_IntermediateSnapshotGreyscale.at<uint8_t>(y, x) = static_cast<uchar>(sum / kernelPixels);
```
neworderofjamie: It seems a bit weird we're sticking ``uint8_t`` in one template and ``uchar`` in another - maybe make them match (personally I would go for ``uint8_t`` as it's standard C++ rather than OpenCV(?) but up to you)

alexdewar:
> Oh no! I hope this isn't the prelude to a QT GUI for bob robotics..
Haha, no dw ;-)

Ta for looking through this.

PR #68, opened 2018-11-02
Error handling

alexdewar:
This PR is mostly to add optional handling of signals (esp. ctrl-C presses and segfaults) to our error checking code, which is handy for stopping robots when you terminate a program. I also added a little wrapper so that you can declare a ``bob_main()`` function instead of ``main()`` to get this error catching without sticking in your own ``try...catch`` block, because it's something we're doing commonly.

I also changed the error handling in ``Vicon::UDPClient`` to throw ``OS::Net::NetworkError`` (which includes the OS' error message) rather than writing to stderr.


common/background_exception_catcher.h
```diff
@@ -0,0 +1,117 @@
...
+#include <stdexcept>
+#include <string>
+#include <unordered_set>
+
+namespace BoBRobotics {
+//----------------------------------------------------------------------------
+// BoBRobotics::BackgroundExceptionCatcher
+//----------------------------------------------------------------------------
+//! A wrapper for passing exceptions between threads (i.e. a background thread to the main one)
+class BackgroundExceptionCatcher
```
neworderofjamie: You could use the classic Singleton (anti-)design pattern here rather than having so many statics. Also MAYBE the templatedness could even prevent the need for the #define to prevent multiple symbol declaration.
```c++
template<typename T>
class Singleton
{
public:
    static T *getInstance()
    {
        if(s_Instance == nullptr) {
            s_Instance = new T;
        }
        return s_Instance;
    }
private:
    
    static T *s_Instance;
};
Singleton<T>::s_Instance = nullptr;
```

Then you'd use it with the curiously recurring template pattern so:
```c++
class BackgroundExceptionCatcher : public Singleton<BackgroundExceptionCatcher>
{
....
}
```

And access the background exception catcher with ``BackgroundExceptionCatcher::getInstance()``

common/background_exception_catcher.h
```diff
@@ -0,0 +1,117 @@
...
+    {
+        // Unregister signal handlers
+        for (auto sig : m_Signals) {
+            std::signal(sig, SIG_DFL);
+        }
+
+        CatcherExists = false;
+    }
+
+#ifdef DEBUG
```
neworderofjamie: Having function signatures in #ifdefs never seems the best plan...How about?
```suggestion
#ifdef DEBUG
#define DEFAULT_TRAPPED_SIGNALS { SIGINT }
#else
#define DEFAULT_TRAPPED_SIGNALS { SIGSEGV, SIGINT, SIGFPE }
#endif
void trapSignals(const std::unordered_set<int> &signals = DEFAULT_TRAPPED_SIGNALS)
{
}
```

common/background_exception_catcher.h
alexdewar: Good idea. I've found a slightly more C++11-y version of this design pattern to use: https://stackoverflow.com/questions/1008019/c-singleton-design-pattern

common/background_exception_catcher.h
alexdewar: This is def less gross than my way :-)

PR #69, opened 2018-11-05
Exceptions and RAII-ification

alexdewar:
Hey,

I noticed that we weren't always getting exceptions thrown when the I2C or see3cam broke on the robot, so  I went on a bit of a mission to change the interfaces to indicate errors by throwing exceptions rather than returning an error value. This means less error-checking boiler-plate is needed and means we don't have to remember to do these checks everywhere.

I also changed the networking code API so that we use a more RAII-style ``Connection`` object representing a single connection, rather than having the ``Client`` and ``Server`` both be ``Node``s (where the server may or may not currently have a valid connection on the go).

All in all this shaves a couple of hundred lines of code off and hopefully makes the interfaces a bit cleaner and more robust.


examples/see3cam/see3cam_test.cc
```diff
@@ -95,29 +93,30 @@ int main(int argc, char *argv[])
...
-            }
+                    cam.captureSuperPixelGreyscale(output);
+                }
 
-            if(success) {
                 cv::imshow("Raw", output);
 
                 unwrapper.unwrap(output, unwrapped);
                 cv::imshow("Unwrapped", unwrapped);
+            } catch (...) {
```
neworderofjamie: Is it not possible to be more specific here?

alexdewar: Hmm. Yeah it probably should only catch our ``std::runtime_error``s

alexdewar:
Thanks :-). I've just had a thought that some of the higher-level parts of ``Socket`` (like buffering) probably more properly belong in ``Connection`` now...

neworderofjamie:
I would be very supportive of that change!

net/connection.h
```diff
@@ -114,14 +133,17 @@ class Connection : public Threadable
...
+            command = readCommand();
         } while (isRunning() && parseCommand(command));
     }
 
 private:
     std::map<std::string, CommandHandler> m_CommandHandlers;
-    std::unique_ptr<std::mutex> m_SendMutex;
+    std::vector<char> m_Buffer;
     Socket m_Socket;
+    std::unique_ptr<std::mutex> m_SendMutex;
```
neworderofjamie: How come this is a pointer?

alexdewar: Sorry, merged without giving you a chance to look at it!

I made the mutex into a pointer so that the class is moveable, which we need so that we can return a ``Connection`` object from a function in the ``Server`` class (although the move might get elided anyway). It turns out mutexes aren't copyable or moveable.

PR #72, opened 2018-11-09
Robot simulator tweaks

alexdewar:
Hi Norbert,

I had a play with your robot simulator code last night and got an example running. It's looking good! I made some fairly minor changes in hacking around with it:
* I added my own ``car.bmp`` because yours wasn't included (but feel free to change back if yours looks cooler!)
* There were some missing ``breaks`` in a switch statement
* The ``quit`` variable didn't have its value initially set so it was causing segfaults on my machine
* I've tidied up the use of the units lib and I think I found some units-related problems (for ``v = wr`` the ``w`` needs to be in radians/sec)
* Use constants cf. ``#define``s -- this threw up a scaling issue (there was a multiple where there should have been a divide)

I'm hoping I haven't broken anything, but the parameter values you're using might have to change with the scaling fixes I've done.


alexdewar:
PS -- If you do have code to test the robot positioner code with the simulator I wouldn't mind having a play with it... Could you email it to me or push it on to a new branch on Github? Ta!

PR #73, opened 2018-11-12
More fixes for simulator

alexdewar:
Hi Norbert,

I hit up against the same compiler error you were getting -- turns out I needed extra separate definitions for the ``static consexpr`` class members.

I've also merged in master and changed the example's makefile so it works with changes I've made to makefile stuff on master.

For future reference, we now have an example makefile for BoB robotics projects etc.: https://github.com/BrainsOnBoard/bob_robotics/blob/master/make_common/Makefile.example


alexdewar:
Actually, on second thoughts, let's put off putting in a PR for this for a while. I've been making some other changes in the process of trying to get it to work with the positioner code.

PR #76, opened 2018-11-21
Throw exceptions in a few places in libantworld

alexdewar:
Hi. This PR just changes libantworld in a few places to throw exceptions when files can't be loaded rather than returning false, in keeping with the style we've used elsewhere.

### DIFF
--- a/tools/ant_world_db_creator/ant_world_db_creator.cc
+++ b/tools/ant_world_db_creator/ant_world_db_creator.cc
@@ -229,9 +229,7 @@ main(int argc, char **argv)
     if (argc > 1) {
         // Create route object and load route file specified by command line
         AntWorld::RouteContinuous route(0.2f, 800);
-        if (!route.load(argv[1])) {
-            return 1;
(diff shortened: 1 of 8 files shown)


tools/ant_world_db_creator/ant_world_db_creator.cc
```diff
@@ -229,9 +229,7 @@ main(int argc, char **argv)
     if (argc > 1) {
         // Create route object and load route file specified by command line
         AntWorld::RouteContinuous route(0.2f, 800);
-        if (!route.load(argv[1])) {
-            return 1;
-        }
+        route.load(argv[1]);
```
neworderofjamie: Could you fix ardin_mb as well? It tests the return value of this at https://github.com/BrainsOnBoard/bob_robotics/blob/master/projects/ardin_mb/state_handler.cc#L33

alexdewar:
Cheers for looking at this! ``ardin_mb`` was already fixed up.

PR #77, opened 2018-11-21
Add Parrot's SDK as git submodule

alexdewar:
This PR adds the SDK as a submodule and automatically builds it when building libbebop. It'll only do this if the ``ARSDK_ROOT`` environment variable isn't set though, so we can keep on using it as before. I just thought it might be an idea to bundle it for when other people want to use the Bebop code.

### DIFF
--- a/third_party/arsdk3/Makefile
+++ b/third_party/arsdk3/Makefile
@@ -1,17 +1,25 @@
 .PHONY: checkout build
 
 checkout:
+	@# Check out git submodule
 	@git submodule update --init arsdk_manifests
+
+	@# If build.sh doesn't exist, then checkout repo
 ifeq ("$(wildcard build.sh)","")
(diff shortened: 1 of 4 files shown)


third_party/arsdk3/Makefile
```diff
@@ -0,0 +1,17 @@
+.PHONY: checkout build
+
+checkout:
+	@git submodule update --init arsdk_manifests
+ifeq ("$(wildcard build.sh)","")
```
neworderofjamie: It seems that, due to Make being super intuitive and all, ``$(wildcard XX)`` **is** the best way of doing this, but it might be nice to comment what it's doing here cos I had to look it up

alexdewar: Sure, that's a good point

PR #78, opened 2018-11-24
Add support for OpenCV 4

alexdewar:
So OpenCV 4 is out and Arch has already upgraded, which predictably means that BoB robotics stuff won't compile for me any more.

Thankfully it seems like they haven't introduced any breaking changes that I can tell, but they have removed large chunks of the old C API, which notably means that all the ``CV_*`` constants we're using don't exist any more and you're supposed to use their enums instead (which also exist in OpenCV 3).

I've made these changes and now everything seems to work for me. Hopefully it shouldn't break anything for OpenCV 3 either.


alexdewar:
No worries! Thanks for looking at it.

PR #80, opened 2018-12-06
Keep track of when Vicon packets were received (fixes #79)

### LINKED ISSUE #79: VICON module doesn't handle object loss
Occasionally the VICON system will drop an object. If this occurs the VICON module currently (from my reading of the code) would continue to pass on the old object position continuously. This is dangerous as the code using this data has no way of knowing that a dropout has occurred and handling it. Two options exist - reset all object data to zero before parsing a new VICON packet - thus the end code can detect zero and act accordingly. Conversely drop the object from the list when the VICON data drops out - this is obviously a larger fix, but has the benefit of being clearer to the end code - if an object no longer exists in the VICON class then it has suffered dropout.

alexdewar:
This adds a ``getElapsedTime()`` method to the ``ObjectData`` class so you can e.g. stop a robot if the positional data is too old.

I've updated one of the examples to show this feature.

### DIFF
--- a/common/stopwatch.h
+++ b/common/stopwatch.h
@@ -1,0 +1,39 @@
+#pragma once
+
+// Standard C++ includes
+#include <chrono>
+
+namespace BoBRobotics {
+class Stopwatch
+{
+public:
+    using TimePoint = std::chrono::high_resolution_clock::time_point;
+    using Duration = std::chrono::high_resolution_clock::duration;
+
+    void start()
+    {
+        m_StartTime = now();
+    }
+
+    Duration elapsed() const
+    {
+        return now() - m_StartTime;
+    }
+
+    Duration lap()
+    {
+        const TimePoint currentTime = now();
+        const Duration elapsed = currentTime - m_StartTime;
+        m_StartTime = currentTime;
+        return elapsed;
+    }
+
+    static TimePoint now()
+    {
+        return std::chrono::high_resolution_clock::now();
+    }
+
+private:
+    TimePoint m_StartTime;
+}; // Stopwatch
+} // BoBRobotics
--- a/examples/vicon/vicon_plot_test.cc
+++ b/examples/vicon/vicon_plot_test.cc
@@ -35,9 +35,19 @@ main()
         return EXIT_FAILURE;
     }
 
+    bool warningGiven = false;
     do {
         plt::figure(1);
-        plotAgent(vicon.getObjectData(0), { -2500, 2500 }, { -2500, 2500 });
+        const auto data = vicon.getObjectData(0);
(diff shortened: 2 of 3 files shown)


neworderofjamie:
This seems a bit heavyweight! As object data already contains frame number, can't you just compare that the the latest frame number (which you need to expose)

neworderofjamie:
Agreed - irritatingly the VICON system doesn't tell you what framerate it's transmitting at it just throws data at you

alexdewar:
Yeah. The other problem is that the user would also have to keep track of e.g. the first frame number and the time it was received so that they can work out how long ago in time a given frame number refers to.

I do agree that my implementation is a bit bloated though. One way we could slim it down is to just expose the functionality of the objects' ``Stopwatch``es directly from the ``UDPClient`` class like so:
```
if (vicon.timeSinceLastPacket(/*objectNum=*/ 0) > 500ms) {
...
```
That would mean that we wouldn't need to calculate the duration for every call of ``getObjectData()`` (when it might not be needed).

It also occurs to me that this won't work with multiple objects in any case, because we're not keeping track of the string IDs of objects, so the first set of data in the packet is always taken to refer to object 0, even though (I'm assuming) if the Vicon system has lost track of object 0 but not object 1, then this data will actually be the data for object 1. Longer term we should probably address this, but short term it does mean we could get away with one ``Stopwatch`` for all objects, because things will break for multiple objects anyway.

ajc158:
Yep, I think that this should be wrapped with improvements including tracking string IDs and getting objects by string ID before adding to master - for now I can merge in the changes to my branch to give a degree of safety (since I'm only considering one drone at the moment).

neworderofjamie:
I think that is a good API and only requires adding a timestamp to each object.

I don't quite see what you mean about multiple objects. We're indexing into ``m_ObjectData`` based on the Vicon object ID which I'm **assuming** doesn't change if an object dissapears.

ajc158:
Hmm.. that would need to be checked - I'm not sure that is the case

alexdewar:
My bad, I misremembered the code! So presumably the right option then is to just go with the ``vicon.timeSinceLastPacket(/*objectNum=*/ 0)`` type interface.

alexdewar:
...though being able to use the string IDs would be a nice addition for the future.

neworderofjamie:
Yeah - once you do this I'll implement string IDs in a separate P.R.

alexdewar:
Done -- though I've only build tested this. Probably best if we try it on the Vicon system before merging.

alexdewar:
Good point!

neworderofjamie:
:+1:

alexdewar:
Right -- seems to work so I'll merge it now. Cheers for the input guys!

PR #84, opened 2018-12-20
Add pose template classes + other misc updates

alexdewar:
Hey,

I thought it might be time to try to merge some of the changes I've made on my various other branches because it's getting a bit faffy to keep them up to date with master.

The biggest change in this PR is that I've added some template classes for pose: you now have ``Vector2`` and ``Vector3`` for describing positions as well as ``Pose2`` and ``Pose3`` classes. There's a fair amount of boilerplate in there, but hopefully using these classes should make things clearer and simpler (e.g. you can cast between vectors and poses of different unit types).

Besides that, I've also:
* updated the ``AntAgent`` class with some extra features for simulations (e.g. you can now set a velocity for it and set it moving)
* added a generic tool to run on a Norbot for simply controlling it remotely
* and added a few other small fixes in other places


common/pose.h
```diff
@@ -1,23 +1,209 @@
...
+      : Vector2(array[0], array[1])
+    {}
+
+    operator Vector3<LengthUnit>() const { return Vector3<LengthUnit>(x(), y(), z()); }
+
+    LengthUnit &x() { return (*this)[0]; }
+    const LengthUnit &x() const { return (*this)[0]; }
+    LengthUnit &y() { return (*this)[1]; }
+    const LengthUnit &y() const { return (*this)[1]; }
+    static constexpr LengthUnit z() { return LengthUnit(0); }
```
neworderofjamie: I don't really like ``Vector2`` having z - inspired by some shader language or another - I think some Vector3 constructors like:

```c++
Vector3(const Vector2 &xy, LengthUnit z)
{
}

Vector3(LengthUnit x, const Vector2 &yz)
{
}
```

would be a nicer and, hopefully, clearer way to provide Vector2->Vector3 extension

common/pose.h
```diff
@@ -1,23 +1,209 @@
...
+    LengthUnit &z() { return (*this)[2]; }
+    const LengthUnit &z() const { return (*this)[2]; }
+};
+
+template<typename LengthUnit, typename AngleUnit>
+class Pose3;
+
+//! A two-dimensional pose
+template<typename LengthUnit, typename AngleUnit>
+class Pose2
```
neworderofjamie: I think deriving off anything std::anything is generally viewed as a bad idea and, in this case, there's particularly little point as according to [this](https://stackoverflow.com/questions/37186443/confusion-while-deriving-from-stdtuple-can-not-handle-stdget) the cruddy ``std::get<n>(x)`` stuff doesn't even survive inheritance. Why not just have the length and angles as members? I also think ``return m_Position.x()`` looks way nicer than ``return std::get<0>(*this)[0];``

common/pose.h
```diff
@@ -1,23 +1,209 @@
...
+    const LengthUnit &x() const { return std::get<0>(*this)[0]; }
+    LengthUnit &y() { return std::get<0>(*this)[1]; }
+    const LengthUnit &y() const { return std::get<0>(*this)[1]; }
+    static constexpr LengthUnit z() { return LengthUnit(0); }
+
+    Array3<AngleUnit> attitude() const { return { yaw(), AngleUnit(0), AngleUnit(0) }; }
+    AngleUnit &yaw() { return std::get<1>(*this); }
+    const AngleUnit &yaw() const { return std::get<1>(*this); }
+    static constexpr AngleUnit pitch() { return AngleUnit(0); }
+    static constexpr AngleUnit roll() { return AngleUnit(0); }
```
neworderofjamie: Again I don't really like this way of making different types of pose interchangable. How about just doing away with ``Pose2`` and use ``Pose3`` for everything? Those extra bytes almost certainly don't matter and 2D stuff can just leave alone the components they don't use (also quite a lot of 2Dish stuff like the antworld are essentially 2D but z isn't at zero so they hover above the ground a bit...

libantworld/agent.h
```diff
@@ -21,67 +35,279 @@ using namespace units::literals;
...
+        updatePose();
+        return m_Pose;
+    }
+
+    GLFWwindow *getWindow() const
+    {
+        return m_Window;
+    }
+
+    radians_per_second_t getMaximumTurnSpeed()
```
neworderofjamie: const?

libantworld/agent.h
```diff
@@ -21,67 +35,279 @@ using namespace units::literals;
...
+                });
     }
 
-    template<class AngleUnit = degree_t>
-    Vector3<AngleUnit> getAttitude() const
+    template<typename LengthUnit = meter_t>
+    Vector3<LengthUnit> getPosition()
     {
-        return convertUnitArray<AngleUnit>(m_Attitude);
+        std::lock_guard<std::mutex> guard(m_PoseMutex);
```
neworderofjamie: maybe make ``m_PostMutex`` mutable so you can make this API const

libantworld/agent.h
```diff
@@ -21,67 +35,279 @@ using namespace units::literals;
...
+    {
+        throw std::runtime_error("GLFW error number: " + std::to_string(errorNumber) + ", message:" + message);
+    }
+
+    static void handleGLError(GLenum, GLenum, GLuint, GLenum, GLsizei, const GLchar *message, const void *)
+    {
+        throw std::runtime_error(message);
+    }
+
+    static TimeType now()
```
neworderofjamie: What's this for? It looks like you're using the stop watch for calculating elapsed time between updates

libantworld/agent.h
```diff
@@ -21,67 +35,279 @@ using namespace units::literals;
...
+
+    virtual bool readFrame(cv::Mat &frame) override
+    {
+        update();
+
         // Read frame
         return Video::OpenGL::readFrame(frame);
     }
 
+    virtual void stopMoving() override
```
neworderofjamie: Is there any particular reason for the move mode stuff? It seems overly complicated in this instance (maybe it makes more sense for something else?) - why can't you move and turn at the same time and why do you have to not be moving in order to set position?

libantworld/agent.h
```diff
@@ -21,67 +35,279 @@ using namespace units::literals;
...
+
+        using namespace units::math;
+        switch (m_MoveMode) {
+        case MoveMode::MovingForward: {
+            const meter_t dist = m_MoveSpeed * m_Velocity * elapsed;
+            m_Pose.x() += dist * cos(m_Pose.yaw());
+            m_Pose.y() += dist * sin(m_Pose.yaw());
+        } break;
+        case MoveMode::Turning:
+            m_Pose.yaw() -= m_MoveSpeed * m_TurnSpeed * elapsed;
```
neworderofjamie: How come you're multiplying by both ``m_MoveSpeed`` and ``m_TurnSpeed``?

navigation/image_database.h
```diff
@@ -205,7 +205,7 @@ class ImageDatabase
         }
 
         //! Get the physical position represented by grid coordinates
-        Vector3<millimeter_t> getPosition(const Vector3<size_t> &gridPosition)
+        Vector3<millimeter_t> getPosition(const Array3<size_t> &gridPosition)
```
neworderofjamie: const?

vicon/udp.h
```diff
@@ -171,6 +173,47 @@ template<typename ObjectDataType = ObjectData>
 class UDPClient
 {
 public:
+    class Object
```
neworderofjamie: I think this is a nice idea but maybe call it ``ObjectReference`` or ``ObjectRef`` or something to save confusion? Not totally sure it needs to be a child class of ``UDPClient`` either.

video/opengl.h
```diff
@@ -21,14 +21,12 @@ class OpenGL : public Input
...
-     * @param readWidth, readHeight Size of image
-     *
-     * **NOTE** intentionally NOT using a cv::Rect as OpenGL coordinates are (typically)
-     * from bottom left of screen which would require window size etc. to convert
+     * @param size Size of image
+     * @param bottomLeft The starting coordinates to read from (from bottom left of screen)
      */
-    OpenGL(GLint readX,  GLint readY,  GLsizei readWidth,  GLsizei readHeight)
-      : m_ReadX(readX), m_ReadY(readY), m_ReadWidth(readWidth), m_ReadHeight(readHeight)
+    OpenGL(const cv::Size &size, const cv::Point &bottomLeft = { 0, 0 })
```
neworderofjamie: I think this is a good change but can you check that ArdinMB and the various examples are updated to handle it?

common/default_get_pose.h
```diff
@@ -0,0 +1,19 @@
...
+namespace BoBRobotics {
+template<typename Derived, typename DefaultLengthUnit, typename DefaultAngleUnit>
+class DefaultGetPose
+{
+public:
+    template<typename LengthUnit = DefaultLengthUnit, typename AngleUnit = DefaultAngleUnit>
+    auto getPose()
+    {
+        auto derived = reinterpret_cast<Derived *>(this);
+        return Pose3<LengthUnit, AngleUnit>(derived->template getPosition<LengthUnit>(),
```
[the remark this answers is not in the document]
alexdewar: Yeah -- not pretty :-/

common/default_get_pose.h
```diff
@@ -0,0 +1,19 @@
...
+
+namespace BoBRobotics {
+template<typename Derived, typename DefaultLengthUnit, typename DefaultAngleUnit>
+class DefaultGetPose
+{
+public:
+    template<typename LengthUnit = DefaultLengthUnit, typename AngleUnit = DefaultAngleUnit>
+    auto getPose()
+    {
+        auto derived = reinterpret_cast<Derived *>(this);
```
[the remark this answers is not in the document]
alexdewar: Ahh ok, cool.

alexdewar:
Yeah I agree we should make sure we're happy with it given how much we use this code.

Annoyingly Eigen doesn't play nicely with the units library, because it can't work out all the conversions it needs to do between units. We could def add a function to access the vector types as Eigen vectors of doubles though. And probably we could use Eigen arrays in the places where we're currently using ``Array2`` and ``Array3``.

common/pose.h
```diff
@@ -1,23 +1,209 @@
...
+
+public:
+    VectorBase() = default;
+
+    template<typename... Ts>
+    VectorBase(Ts &&... args)
+      : m_Array({ std::forward<Ts>(args)... })
+    {}
+
+    operator const std::array<LengthUnit, N> &() const
```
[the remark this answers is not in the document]
alexdewar: Yeah that makes sense. Alternatively we could swap them out for Eigen arrays/vectors in some places where it would simplify things.

common/pose.h
```diff
@@ -1,23 +1,209 @@
...
+
+    Vector3(LengthUnit x, LengthUnit y, LengthUnit z)
+      : VectorBase<LengthUnit, 3>(x, y, z)
+    {}
+
+    Vector3(const std::array<LengthUnit, 3> &array)
+      : Vector3(array[0], array[1], array[2])
+    {}
+
+    operator Vector2<LengthUnit>() const { return Vector2<LengthUnit>(x(), y()); }
```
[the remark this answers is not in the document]
alexdewar: Sure, that sounds good. I'm guessing you mean we could return ``std::array``s from this rather than a bunch of new classes though....?

libantworld/agent.h
alexdewar: Oops, must've forgotten to take this out

libantworld/agent.h
alexdewar: I was using it for some simple simulations elsewhere but I'm not using any of the moving stuff in this branch atm.

You can only turn on the spot or move forwards at the moment, but we could change it to take an angular velocity and a forward velocity (I've got code for this with the robot positioner stuff). I didn't bother adding the code to update pose while the agent is moving, but it wouldn't be hard to do and maybe I should.

libantworld/agent.h
alexdewar: They're just badly named variables. ``m_MoveSpeed`` is a proportion and ``m_TurnSpeed`` is the max turning speed. I'll make it more explicit

vicon/udp.h
alexdewar: Yeah good point. It's not a great name.

video/opengl.h
alexdewar: Ok, will do!

common/pose.h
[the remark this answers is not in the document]
neworderofjamie: Does it make any difference? those classes are only a very thin wrapper over ``std::array`` anyway - it shouldn't make any difference performance-wise

libantworld/agent.h
alexdewar: ...although it turns out it calls ``updatePose()`` which sets some other vars. Although it is a little odd to have a non-const getter, I think it does actually make sense here (because the pose is being updated based on elapsed time every time it's called).

libantworld/agent.h
alexdewar: Maybe let's just leave this as-is for the mo -- we can maybe tidy it up once the other simulation/robot positioner stuff is merged.

libantworld/agent.h
neworderofjamie: That seems like bad getter behaviour to me! I feel it would be nicer if``updatePost`` just get called once a frame from ``update`` and ``getPosition`` returns that frame's state? I've definitely lost track of what this does since 20th December :smile: so that may make no sense at all

alexdewar: Haha, yes, sorry for leaving it for so long!

That does make sense. I'm thinking actually that having the ``Stopwatch`` inside this class is a mistake. Might be better to let the user call ``updatePose()`` and pass in the elapsed time as an arg, so we don't always have to run the simulation in real time. Whaddya think?

neworderofjamie: Agreed - that sounds much better

common/pose.h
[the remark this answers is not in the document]
alexdewar: I wasn't thinking about performance: it's just that a ``Vector2`` treats the two members of its array as x and y. So it's probably easier to return ``std::array``s rather than making some new classes for it.

common/pose.h
alexdewar: We could use ``Pose3`` in most places, but there are places where we are only saving x, y and angle where ``Pose2`` would fit better. Performance-wise it won't matter if we have one ``Pose3`` object, but it might do for e.g. a really big array of Vicon data and if we don't care about the other dimensions it doesn't seem like there's much point in saving them... I'm not totally against the idea, but I feel like it makes sense to have a separate class for this.

I agree it's a bit weird to have e.g. a ``z()`` for a ``Pose2``, but it's mostly for some of the simulation bits and pieces I've done on other branches: the idea was that you can have template functions which take a ``PoseType`` as an argument and then just write a generic function using ``x()``, ``y()`` and ``z()`` etc.

I've used some variation on this code for the robot positioner stuff and for other simulation bits I've been doing, but I thought I'd do it as a separate PR because it touches a lot of code and I thought it might be the most controversial bit ;-). I don't mind hacking away away at it some more but it would be great if we could get it merged in some form soon-ish just so I can rebase the new robot positioner/visual navigation bits off it :-). Obvs I don't want you to merge anything that's too obviously nasty though!

PR #85, opened 2019-01-17
Bits and pieces to support antworld work

neworderofjamie:
1. Some third party libraries:
   - Nice simple header-only CSV parser (https://github.com/ben-strasser/fast-cpp-csv-parser)
   - Slightly less nice and simple Immediate Mode GUI library (https://github.com/ocornut/imgui)
2. Make file flags to set up the GUI library (and Eigen)
3. More excitingly units-based map coordinate handling library - I'd very much appreciate an actual review on this one
4. ``RenderTarget`` system for Libantworld
   - Allows proper off-screen rendering (have added ``RenderTargetInput`` to integrate this with ``Video::Input``)
   - Allows integration with ImGUI windows
5. Fixed various corner cases in ``World`` handling of obj files produced by different programs


common/map_coordinate.h
```diff
@@ -0,0 +1,231 @@
...
+template<typename D>
+struct Cartesian
+{
+    typedef D Datum;
+
+    const units::length::meter_t x;
+    const units::length::meter_t y;
+    const units::length::meter_t z;
+};
+
```
alexdewar: I like the templating on the datum here -- definitely a nice way to do it

common/map_coordinate.h
```diff
@@ -0,0 +1,231 @@
...
+{
+    static const Transform transform;
+    static const Ellipsoid ellipsoid;
+};
+const Transform OSGB36::transform{units::length::meter_t(-446.448), units::length::meter_t(125.157), units::length::meter_t(-542.060),
+                                  20.4894, units::angle::arcsecond_t(-0.1502), units::angle::arcsecond_t(-0.2470), units::angle::arcsecond_t(-0.8421)};
+const Ellipsoid OSGB36::ellipsoid{units::length::meter_t(6377563.396), units::length::meter_t(6356256.909), 1.0 / 299.3249646};
+
+// An OS coordinate consisting of an 'easting' and 'northing' distance
+typedef std::tuple<units::length::meter_t, units::length::meter_t> OSCoordinate;
```
alexdewar: Maybe ``OSCoordinate`` could be a struct instead? I'm always worried I've got northing and easting the wrong way round when I do these things...

common/map_coordinate.h
```diff
@@ -0,0 +1,231 @@
...
+    using namespace units::length;
+    using namespace units::literals;
+
+    const meter_t h = 0.0_m; // height above ellipsoid - not currently used
+    const double sinPhi = sin(latLon.lat);
+    const double cosPhi = cos(latLon.lat);
+    const double sinLambda = sin(latLon.lon);
+    const double cosLambda = cos(latLon.lon);
+
+    const double eSq = 2.0 * Datum::ellipsoid.f - Datum::ellipsoid.f * Datum::ellipsoid.f;  // 1st eccentricity squared ≡ (a²-b²)/a²
```
alexdewar: Is this definitely right? It's just it's different from the equation in the comment (I can see how eccentricity => ``a`` and ``b``, but you're not doing a division). Maybe another comment is in order if you're doing something non-obvious here.

common/map_coordinate.h
```diff
@@ -0,0 +1,231 @@
...
+
+    return Cartesian<Datum>{(nu + h) * cosPhi * cosLambda,
+                            (nu + h) * cosPhi * sinLambda,
+                            (nu * (1.0 - eSq) + h) * sinPhi};
+}
+
+// Transform cartesian coordinates from WGS84 into desired space
+// Ported from Javascript Ordnance Survey Grid Reference functions (c) Chris Veness 2005-2017 (MIT Licence)
+template<typename Datum>
+Cartesian<Datum> transformCartesian(const Cartesian<WGS84> &c)
```
alexdewar: Is this function specific to the WGS84 datum? Or could we just write it as a template functoin with ``Datum1`` and ``Datum2``?

common/map_coordinate.h
```diff
@@ -0,0 +1,231 @@
...
+// Ported from Javascript Ordnance Survey Grid Reference functions (c) Chris Veness 2005-2017 (MIT Licence)
+template<typename Datum>
+LatLon<Datum> cartesianToLatLon(const Cartesian<Datum> &c)
+{
+    using namespace units::angle;
+    using namespace units::length;
+    using namespace units::math;
+    using namespace units::literals;
+
+    const double e2 = 2.0 * Datum::ellipsoid.f - Datum::ellipsoid.f * Datum::ellipsoid.f;    // 1st eccentricity squared ≡ (a²-b²)/a²
```
alexdewar: See comment above

common/map_coordinate.h
```diff
@@ -0,0 +1,231 @@
...
+    const meter_t r = sqrt(p * p + c.z * c.z);      // polar radius
+
+    // parametric latitude (Bowring eqn 17, replacing tanβ = z·a / p·b)
+    const double tanBeta = (Datum::ellipsoid.b * c.z) / (Datum::ellipsoid.a * p) * (1.0 + epsilon2 * Datum::ellipsoid.b / r);
+    const double sinBeta = tanBeta / std::sqrt(1.0 + tanBeta * tanBeta);
+    const double cosBeta = sinBeta / tanBeta;
+
+    // geodetic latitude (Bowring eqn 18: tanφ = z+ε²bsin³β / p−e²cos³β)
+    const degree_t phi = std::isnan(cosBeta) ? 0.0_rad : atan2(c.z + epsilon2 * Datum::ellipsoid.b * sinBeta * sinBeta * sinBeta,
+                                                               p - e2 * Datum::ellipsoid.a * cosBeta * cosBeta * cosBeta);
```
alexdewar: Am I right in thinking ``std::isnan(cosBeta)`` is the same as ``tanBeta <= 0`` here? If so, that might be a bit clearer, rather than having to try to remember which floating point operations can result in NaN.

common/map_coordinate.h
```diff
@@ -0,0 +1,231 @@
...
+    // height above ellipsoid (Bowring eqn 7) [not currently used]
+    /*const double sinPhi = sin(phi), cosPhi = cos(phi);
+    const meter_t nu = std::get<0>(ellipsoid) / std::sqrt(1.0 - e2 * sinPhi * sinPhi); // length of the normal terminated by the minor axis
+    var h = p*cosφ + cartesian[2]*sinφ - (osgb36EllipseA*osgb36EllipseA/ν);*/
+
+    return LatLon<Datum>{phi, lambda};
+}
+
+// Convert latitude and longitude in the OSGB36 space to OS grid coordinates
+// Ported from Javascript Ordnance Survey Grid Reference functions (c) Chris Veness 2005-2017 (MIT Licence)
```
alexdewar: Would you mind doing these comments in doxygen style? I.e.:
```
/**!
 * \brief Convert latitude and longitude in the OSGB36 space to OS grid coordinates
 * Ported from Javascript Ordnance Survey Grid Reference functions (c) Chris Veness 2005-2017 (MIT Licence)
 */

File: libantworld/render_target_input.cc [hunk 8]
```diff
@@ -0,0 +1,38 @@
...
+//----------------------------------------------------------------------------
+// BoBRobotics::AntWorld::RenderTargetInput
+//----------------------------------------------------------------------------
+namespace BoBRobotics
+{
+namespace AntWorld
+{
+cv::Size RenderTargetInput::getOutputSize() const
+{
+    return cv::Size(m_RenderTarget.getWidth(), m_RenderTarget.getHeight());
```
alexdewar: Maybe it makes sense just to have ``RenderTarget`` use ``cv::Size``?

libantworld/render_target_input.cc
```diff
@@ -0,0 +1,38 @@
...
+    // Read texture into outframe
+    glGetTexImage(GL_TEXTURE_2D, 0, GL_BGR, GL_UNSIGNED_BYTE, outFrame.data);
+
+    // Flip image vertically
+    cv::flip(outFrame, outFrame, 0);
+
+    // Unbind texture
+    glBindTexture(GL_TEXTURE_2D, 0);
+
+    return true;
```
alexdewar: Does this implementation differ from that in ``Video::OpenGL``? Could they be amalgamated (e.g. with inheritance)?

libantworld/render_target_input.h
```diff
@@ -0,0 +1,63 @@
...
+{
+public:
+    /*!
+     * \brief Create a Video::Input for reading from an OpenGL window
+     *
+     * @param readX, readY The starting coordinates to read from (from bottom left of screen)
+     * @param readWidth, readHeight Size of image
+     *
+     * **NOTE** intentionally NOT using a cv::Rect as OpenGL coordinates are (typically)
+     * from bottom left of screen which would require window size etc. to convert
```
alexdewar: Looks like we could drop this comment.

libantworld/renderer.cc
```diff
@@ -168,6 +203,30 @@ void Renderer::renderFirstPersonView(meter_t x, meter_t y, meter_t z,
     m_World.render();
 }
 //----------------------------------------------------------------------------
+void Renderer::renderFirstPersonView(meter_t x, meter_t y, meter_t z,
+                                     degree_t yaw, degree_t pitch, degree_t roll,
+                                     RenderTarget &renderTarget, bool bind, bool clear)
```
alexdewar: I think we could do without this function and the one above, as the user could just call ``.bind()`` and ``.clear()`` manually -- but up to you.

libantworld/renderable.h
```diff
@@ -0,0 +1,21 @@
...
+{
+class Renderable
+{
+public:
+    virtual ~Renderable(){}
+
+    //----------------------------------------------------------------------------
+    // Declared virtuals
+    //----------------------------------------------------------------------------
+    virtual void render() const = 0;
```
alexdewar: It seems a bit weird that this is ``const``. Surely the internal state of ``Renderable``s changes when you render them?

libantworld/world.cc
```diff
@@ -34,7 +34,9 @@ void readVector(std::istringstream &stream, std::vector<GLfloat> &vector, float
     }
 
     // Check this is the end of the linestream i.e. there aren't extra components
-    BOB_ASSERT(stream.eof());
+    if(assertIfNotEOF) {
+        BOB_ASSERT(stream.eof());
+    }
```
alexdewar: Adding an extra arg just for this assert statement seems like a bit of overkill. How about we just drop the ``BOB_ASSERT`` from here and the caller can check for EOF at the call site if needed?

libantworld/world.cc
```diff
@@ -509,8 +544,22 @@ void World::Surface::bind() const
...
+void World::Surface::unbind() const
+{
+    // If surface has a texture, bind it
+    if(m_Texture != nullptr) {
+        glDisable(GL_TEXTURE_2D);
+        m_Texture->unbind();
+    }
+
+    // Unbind vertex array
+    glBindVertexArray(0);
```
alexdewar: Ditto above

libantworld/world.cc
```diff
@@ -451,6 +479,12 @@ void World::Texture::bind() const
     glBindTexture(GL_TEXTURE_2D, m_Texture);
 }
 //----------------------------------------------------------------------------
+void World::Texture::unbind() const
+{
+    // Unbind texture
+    glBindTexture(GL_TEXTURE_2D, 0);
```
alexdewar: Is the ``0`` a ``nullptr``? If so, could be explicit about that

make_common/bob_robotics.mk
```diff
@@ -107,3 +111,8 @@ ifdef WITH_I2C
 	LINK_FLAGS += -li2c
 	endif
 endif
+
+ifdef WITH_EIGEN
+	CXXFLAGS += `pkg-config --cflags eigen3`
+	LINK_FLAGS += -fopenmp
+endif
```
alexdewar: It's a good idea to have this.

third_party/csv.h
```diff
@@ -0,0 +1,1268 @@
+// Copyright: (2012-2015) Ben Strasser <code@ben-strasser.net>
```
alexdewar: This looks handy!

third_party/imgui/.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
+## How to create an Issue
```
alexdewar: This looks like a cool library, but would you mind adding it as a git submodule? It's what we're doing elsewhere (e.g. for ARSDK and msp) and does make things a bit tidier IMHO.

libantworld/world.cc
alexdewar: above == below, obviously

common/map_coordinate.h
neworderofjamie: Yeah - I was pleased with that bit of C++ inspiration

common/map_coordinate.h
neworderofjamie: Yeah - I started off with all tuples but then had that worry and changed several things to structs - I should be consistent and do the same here.

common/map_coordinate.h
neworderofjamie: Basically yes - all the transforms are transforming from WGS84 (maybe I should make that clearer in the name of the transforms)

common/map_coordinate.h
neworderofjamie: It is, but I'm not sure of the purpose of f at all. It's called the flattening but can be calculated as ``f = (a -b) / a`` https://en.wikipedia.org/wiki/Flattening so I'm not sure why the original javascript had it manually calculated and used inconsistently - I will double check the maths and probably remove f

libantworld/render_target_input.cc
neworderofjamie: I think there is one as it uses it for loading textures but I don't reaaaally want to introduce too much coupling between libantworld and opencv

libantworld/render_target_input.cc
neworderofjamie: Yes - ``Video::OpenGL`` reads data from the 'display framebuffer' which is basically the same data the OS uses to draw to the screen. ``RenderTarget`` lets you not render to that and instead render to an offscreen buffer and ``RenderTargetInput`` lets you read _that_ back.

libantworld/renderable.h
neworderofjamie: I'm not 100% sure about this bit of design at all tbh but render really shouldn't be changing any state - it should just making some opengl draw calls.

libantworld/world.cc
neworderofjamie: good point

libantworld/world.cc
neworderofjamie: no - OpenGL handles are integers

libantworld/render_target_input.h
neworderofjamie: good spot

common/map_coordinate.h
neworderofjamie: Ok so, from Wikipedia:
eq1
Also from Wikipedia:
codecogseqn
And via the power of my rusty algebra:
codecogseqn 1

So they're equivalent! What do you think I should do:
1. Leave it but update the comment
2. Get rid of the manually calculated ``f`` term and put everything in terms of ``a`` and ``b``
3. Get rid of the manually calculated ``f`` term and add a method to the struct which calculates it from ``a`` and ``b``?

neworderofjamie: I kinda think 2 as ``latLonToOS`` already does that

alexdewar: Yeah I think either 1 or 2. It was just the equation in the comment which threw me a bit.

common/map_coordinate.h
alexdewar: But what I meant was would this function also work for tranforming e.g. from something else to WGS84? In which case it might be worth having the other datum also as a template parameter just to indiciate that you're not doing any weird WGS84-specific things, and also in case we decide to convert from something other than WGS84 in future.

common/map_coordinate.h
alexdewar: Yeah it does, but I think ``//! Comments`` are for single-liners: i.e. just a brief description. The example I gave is one line of brief description (i.e. in the doxygen index) + extra info on the main page.

libantworld/render_target_input.cc
alexdewar: We are using OpenCV basically everywhere though... but it's up to you.

libantworld/render_target_input.cc
alexdewar: Ahh ok

common/map_coordinate.h
alexdewar: ...I don't have a preference style-wise though.

libantworld/world.cc
alexdewar: Ahh yeah of course.

third_party/imgui/.github/CONTRIBUTING.md
alexdewar: Cool. I do have a makefile for the ARSDK that checks out the submodule and builds it if you'd like an example.

common/map_coordinate.h
neworderofjamie: No, it only transforms from cartesian coordinates in the WGS84 datum to another datum of your choice

alexdewar: ...another thought. You could have a ``.transform()`` member function of ``Cartesian`` instead of having it as a separate function. Just a thought though.

alexdewar: Ahh fair enough. Dw about it then.

third_party/imgui/.github/CONTRIBUTING.md
neworderofjamie: I have attempted to set this up - could you check it works for you (building the new rothamsted_reconstruction repro is probably a good test)

alexdewar: Yep, it builds fine for me.

You could also retrieve the git submodule with the makefile:
``
submodule:
    $(shell git submodule update --init imgui)
``

Then make ``%.o`` depend on ``submodule``.

alexdewar: ...sorry, misread your comment. The submodule builds ok for me. I'll test the rothamsted_reconstruction repo now.

alexdewar: Right, so it did build for me, but I had to change the includes to read:
``
// BoB robotics third-party includes
#include "third_party/imgui/imgui/imgui.h"
#include "third_party/imgui/imgui/imgui_internal.h"
#include "third_party/imgui/imgui/examples/imgui_impl_glfw.h"
#include "third_party/imgui/imgui/examples/imgui_impl_opengl2.h"
``

neworderofjamie: The includes in the submodule or those in the rothamsted reconstruction repo (which I definitely failed to push...)?

alexdewar:
The Rothamsted reconstruction repo. The submodule built fine for me.

⁣Sent from BlueMail ​

third_party/imgui/.github/CONTRIBUTING.md
alexdewar: The submodule wasn't being checked out first for me (my mistake!). I've pushed a fix. Other than that I think we're done!

PR #86, opened 2019-01-25
Fix for gantry code + misc tweaks

alexdewar:
It turns out there was a small bug in the example gantry code we have there which I fixed up.

While I was at it I made some other small fixes and added a ``remove_all()`` function to ``path.h`` (after C++17), which uses the tinydir code. Annoyingly Visual Studio also reformatted all the code in ``path.h``.


alexdewar:
Plus I added an option to overwrite existing image databases...

PR #87, opened 2019-01-25
Norbot control program + fixes

alexdewar:
Hey,

I thought it would be handy to have a generic program for remotely controlling Norbots with a joystick and reading from the camera etc. There is one already in the examples folder, but I thought it would be better if we have a separate program that actually probes to see if there is a camera or joystick etc. without terminating. Essentially I got bored of rewriting little stubs for this kind of thing and thought we could do with a generic tool.

I also fixed some minor bugs to do with threading etc. In the process I ended up with destructors hanging a bunch of times and because ``BackgroundExceptionCatcher`` had added a signal handler for ctrl+C it was quite annoying to quit programs. So I changed it so that the signal handlers are unset again when the ``BackgroundExceptionCatcher`` object gets destroyed and then you can at least quit the program. It does mean the syntax looks a little odd (I think we had something similar before and dropped it) but it seems a bit cleaner this way.


alexdewar:
Hey @neworderofjamie , any chance you could take a quick look at this?

I'm guessing it's fallen through the cracks :-p

PR #90, opened 2019-02-15
Small bug fix for InfoMax

neworderofjamie:
1. ``size_t`` is unsigned meaning that ``bestIndex`` logic was broken
2. ``InfoMaxRotater`` didn't implement ``getImageDifferences``


navigation/infomax.h
```diff
@@ -203,27 +203,56 @@ class InfoMaxRotater : public InfoMax<FloatType>
...
     //------------------------------------------------------------------------
+    template<class... Ts>
+    const std::vector<FloatType> &getImageDifferences(Ts &&... args) const
+    {
+        calcImageDifferences(std::forward<Ts>(args)...);
+        return m_RotatedDifferences;
+    }
+
     template<class... Ts>
     auto getHeading(Ts &&... args) const
```
alexdewar: Does this need a ``&`` to make sure we get a reference and not a copy?

neworderofjamie: Where exactly? possible places to put ``&`` is almost as fun a C++ quiz question as possible places to put ``const``...

alexdewar: Oh yeah, sorry :-p

I *think* the ``auto`` in the return type might be resolved to a value type cf. a reference.

I meant like this:
``auto &getHeading(Ts &&... args) const``

It also looks like you can use ``decltype(auto)`` for this:
https://en.cppreference.com/w/cpp/language/auto

neworderofjamie: Ahh ok - I don't **think** ``getHeading`` should return a reference though - it's returning a tuple which, amongst other things, include a reference to ``m_RotatedDifferences``.

alexdewar: Ah, my bad. I guess in that case you def won't want to return by reference.

PR #92, opened 2019-02-22
Add basic support for Lego Mindstorms robots

alexdewar:
I finally got fed up listening to the screeching of Norbots' servos, so I've had a bit of a play with a Lego Mindstorms robot -- and the control is much nicer!

Part of the motivation is just to have a robot I can actually control so we can finally merge the robot positioner stuff, but I think it might also be a better platform for doing robot experiments in the drone zone in the short term (the floor is now so buckled that Norbots can't even drive across it :-(), possibly without actually strapping a Jetson on to it.


alexdewar:
Cool. It's with Debian on it -- you can just stick it on an SD card and run it from that without reflashing, which is quite nice.

PR #93, opened 2019-02-25
Insilico rotation range

neworderofjamie:
Restricting the range you insilico rotate images by, not only improved performance, but also seems to help reduce visual aliasing. I'm currently using it like https://github.com/BrainsOnBoard/snapshot_bot/blob/bob_robotics_update/memory.cc#L91-L139 which is a bit ugly. Ideally the logic would be moved into ``PerfectMemoryRotater`` and ``InfoMaxRotater`` but, especially when you want to scan a range not centred about zero, I don't **think** it's trivial as there's potentially something like 4 discontinuous sections to the scan.

Any thoughts on what the API should look like/whe


navigation/insilico_rotater.h
```diff
@@ -36,40 +39,52 @@ class InSilicoRotater
...
+                rollImage(m_MaskImage, m_ScanStep);
             }
         }
     }
 
     size_t max() const
     {
-        return static_cast<size_t>(m_Image.cols) / m_ScanStep;
+        return (m_EndRoll - m_BeginRoll) / m_ScanStep;
     }
```
alexdewar: I feel like this function should be renamed to ``size()`` or ``numRotations()`` or something now, because ``max()`` kinda sounds like it should just return ``m_EndRoll``.

alexdewar:
For the case when we have discontinuous ranges, we could change the API so ``InSilicoRotater`` uses either ``size_t``s for the start and end of the range of indexes, or a pair of iterators (so one could e.g. have a ``std::vector`` with a bunch of arbitrary indexes in).

I'm happy to have a go at this if you like?

navigation/insilico_rotater.h
neworderofjamie: Excellent point!

neworderofjamie:
I think maybe that's maybe overkill as, as well as having to figure out what sequence of continuous ranges a ``startAngle``, ``scanAngle`` pair of angles represents, that would mean that ``PerfectMemoryRotater`` and ``InfoMaxRotater`` would also have to repeatedly ``std::iota`` some vectors.

alexdewar:
Not necessarily -- you could get round this with a bit of templating. All the operations we're doing to ``beginRoll`` and ``endRoll`` also work on iterators. The only difference between the iterator-type and ``size_t`` type is that the iterators would need to be dereferenced to get the roll index, so you could define a pair of functions for this:

``size_t toIndex(size_t index) { return index; }``
and
```
template<type Iter>
size_t toIndex(Iter it) { return *it; }
```

neworderofjamie:
That is true but while I am nearly always in favour of a bit of templating, I just don't really see a use case for fully non-continuous column indices (scanStep already lets you change the stride) unless I'm missing some cunning here and that's not what you're trying to achieve. Thinking more, this logic is very ``InSilicoRotater`` dependent (i.e. this is trivial for ``AntWorldRotater``) though so probably does belong here rather than in  ```PerfectMemoryRotater`` and ``InfoMaxRotater`` but, in my mind, the solution is probably just a derived version of ``InSilicoRotater`` with logic to figure out how many times to call ``rotate`` based on it's forwarded constructor parameters though...

alexdewar:
We might be talking at cross purposes here...

I agree it would be weird to compare images at a bunch of out-of-order rotations (e.g. ``{ 1, 3, 2, 0 }``) but it seems like passing in a ``std::vector<size_t>`` of rotations to ``getHeading()`` would a bit cleaner than calling ``getImageDifferences()`` several times for multiple ranges and then doing extra work to find the minimum across the different ranges tested. Or have I got the wrong end of the stick -- was that not what you were trying to do?

It might be easier if I just do a PR to show you what I had in mind ;-)

neworderofjamie:
Yeah that does make sense actually - go for it / just add to this one as it's somewhat half-formed atm

neworderofjamie:
Did this branch add anything? I'm a little confused  - I thought it got replaced by #94

alexdewar:
Uh no -- it's just that I did the PR into this branch rather than master, to make it easier for you to look at the diff.

neworderofjamie:
Ahh of course - I'd forgotten the chronology!

PR #95, opened 2019-02-26
Robot positioner

alexdewar:
Hallelujah! It's actually working now. It turns out that having proper servos makes a big difference.

Unfortunately this is now a fairly huge PR given all the various bits and pieces that have gone into it, although it's mostly been a case of adding bits in rather than mangling existing APIs.

Commit summary:
* The tidied version of Norbert's ``RobotPositioner`` class
* Norbert's SDL-based GUI for displaying a robot's position
* A ``SimulatedTank`` class for testing
* A tool for getting the positions of unlabelled markers from the Vicon system (intended for e.g. masking out the area with an obstacle in)
* A simpler PID system than the ``RobotPositioner`` that only cares about position and not heading (and actually works on Norbots)
* Some data + tools from my basic systems ID work with Norbots
* Various tweaks and fixes all over

@neworderofjamie Don't feel you have to look at everything -- some of these aren't particularly important changes. But if you could give it a quick once-over that'd be ace.

@tenxlenx I've included you in case you want to take a look or have any other thoughts. And slowing the robot down as it approached the goal actually worked :-)


alexdewar:
PS -- @neworderofjamie The build seems to be failing on your machine because you don't have SDL2 installed.

examples/simulated_positioner/simulated_positioner.cc
```diff
@@ -0,0 +1,115 @@
...
+                    std::cout << "Reached goal" << std::endl;
+                    reachedGoalAnnounced = true;
+                }
+            } else {
+                reachedGoalAnnounced = false;
+            }
+        }
+
+        // A small delay, so we don't eat all the CPU
+        std::this_thread::sleep_for(2ms);
```
neworderofjamie: You shouldn't really need these sleeps - if you make sure SDL is vsyncing (google suggests you just add ``SDL_RENDERER_PRESENTVSYNC`` to the flags when you create the rendered) it should be ok

examples/tank_pid/tank_pid.cc
```diff
@@ -0,0 +1,277 @@
...
+using namespace units::length;
+using namespace units::literals;
+using namespace units::angle;
+using namespace units::angular_velocity;
+using namespace units::velocity;
+using namespace std::literals;
+namespace plt = matplotlibcpp;
+
+// For sound
+#define PLAY_PATH "/usr/bin/play"
```
neworderofjamie: hahah, is this really necessary? I'm thoroughly glad my machine doesn't come with that ogg file...

projects/norbot_systems_identification/angular_velocity.cc
```diff
@@ -0,0 +1,130 @@
...
+        if (stopwatch.started()) {
+            std::cout << "Ignoring joystick command" << std::endl;
+            return true;
+        } else {
+            return false;
+        }
+    });
+
+    // Toggle testing mode with buttons
+    joystick.addHandler([&](HID::JButton button, bool pressed) {
```
neworderofjamie: Why not just do all the joystick logic by polling? I find the mix of polling calls and callbacks a bit confusing and it splits up the logic between the main loop and the handler (which I'm guessing/hoping is triggered by the ``m_Joystick.update()`` call but might be on a different thread)

robots/car_display.h
```diff
@@ -0,0 +1,200 @@
...
+        SDL_Init(SDL_INIT_VIDEO);
+
+        m_Window = SDL_CreateWindow("Wheeled-robot simulator",
+                                    SDL_WINDOWPOS_UNDEFINED,
+                                    SDL_WINDOWPOS_UNDEFINED,
+                                    WindowWidth,
+                                    WindowHeight,
+                                    0);
+
+        m_Renderer = SDL_CreateRenderer(m_Window, -1, 0);
```
neworderofjamie: I think if you stick that ``SDL_RENDERER_PRESENTVSYNC`` flag here you shouldn't need to sleep manually - it'll keep the whole loop synchronised to 60fps (16.6ms per update)

robots/norbot.h
```diff
@@ -11,20 +11,29 @@
...
+
 public:
     Norbot(const char *path = "/dev/i2c-1", int slaveAddress = 0x29)
       : m_I2C(path, slaveAddress)
-      , m_Left(0.0f)
-      , m_Right(0.0f)
-    {}
+    {
+        // Sometimes Norbots get stuck driving, so let's stop it if we need to
+        stopMoving();
```
neworderofjamie: wise!

robots/tank.h
```diff
@@ -47,13 +64,48 @@ class Tank
                 });
     }
 
+    void controlWithThumbsticks(HID::Joystick &joystick)
```
neworderofjamie: Hah now that's truly a psychotic control scheme

robots/tank_pid.h
```diff
@@ -0,0 +1,176 @@
...
+            } else if (event == Event::Update) {
+                /*
+                 * If m_HeadingOffset is suitably small, we've finished turning
+                 * towards the goal.
+                 */
+                if (abs(m_HeadingOffset) <= m_AngleTolerance) {
+                    // Start driving to the goal in a straight(ish) line
+                    m_StateMachine.transition(TankPIDState::DrivingToGoal);
+                } else if (m_HeadingOffset < 0_deg) {
+                    m_Robot.tank(1.f, -1.f);
```
neworderofjamie: It might be worth slowing this down, especially with small heading offsets to prevent oscillations. Something like ``(min(90, abs(m_HeadingOffset)) / 90) ** 2`` might work nicely or you could just have a threshold, beneath which, it steers slower. That being said, to get passable norbot control/less toppling in the snapshot bot, I gave up on the forementioned approach and just always turned with a really low steering value of ``0.15``.

tools/vicon_arena_constructor/vicon_arena_constructor.cc
```diff
@@ -0,0 +1,219 @@
+
```
neworderofjamie: I kinda feel that maybe this, it's boost dependency and need for the Vicon SDK would maybe be better off in another repo that just uses bob robotics like the snapshot bot or whatever.

vicon/udp.h
```diff
@@ -57,12 +51,7 @@ class ObjectData
         m_FrameNumber = frameNumber;
 
         // Copy vectors into class
-        m_Position[0] = x;
-        m_Position[1] = y;
-        m_Position[2] = z;
-        m_Attitude[0] = yaw;
-        m_Attitude[1] = pitch;
-        m_Attitude[2] = roll;
+        m_Pose = std::move(pose);
```
neworderofjamie: I don't think poses dynamically allocate anything that can be re-purposed so this probably doesn't do anything (the contents of ``pose`` will still have to be copied from the stack/registers to wherever ``m_Pose`` is located)

.gitmodules
```diff
@@ -4,6 +4,9 @@
 [submodule "third_party/imgui/imgui"]
 	path = third_party/imgui/imgui
 	url = ../../ocornut/imgui.git
+[submodule "third_party/vicon_data_stream_sdk"]
+	path = third_party/vicon_data_stream_sdk
+	url = https://github.com/BrainsOnBoard/ViconDataStreamSDK
```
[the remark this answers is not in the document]
alexdewar: The reason I forked it is so as to use the system boost libs (no compiling needed) -- the official Vicon version comes with all the boost libs bundled, which would have been ok except that I couldn't compile using them.

common/pose.h
```diff
@@ -239,3 +287,32 @@ convertUnitArray(const ArrayType &values)
              static_cast<OutputUnit>(values[2]) };
 }
 } // BoBRobotics
+
+template<typename LengthUnit>
+std::ostream &operator<<(std::ostream &os, const BoBRobotics::Vector2<LengthUnit> &position)
```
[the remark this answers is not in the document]
alexdewar: Yes!

common/pose.h
```diff
@@ -40,18 +97,40 @@ class VectorBase
...
     {
         return m_Array;
     }
 
+    template<typename PositionType>
+    LengthUnit distance3D(const PositionType &point) const
+    {
+        const auto derived = static_cast<const Derived *>(this);
+        using namespace units::math;
+        return sqrt(pow<2>(derived->x() - point.x()) +
```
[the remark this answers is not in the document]
alexdewar: Yep! It needs to be a template for the units lib so that the compiler knows the units at compile time.

common/pose.h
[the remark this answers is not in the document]
neworderofjamie: I only found out when I was making a project with multiple object files (vague attempt to save Jetson compilation time) that there's a few places missing inline :disappointed:

examples/robot_positioner/robot_positioner.cc
```diff
@@ -0,0 +1,227 @@
...
+    {
+        // Goal is currently hard-coded
+        std::cout << "Goal: " << Goal << std::endl;
+        m_Positioner.setGoalPose(Goal);
+
+        // Drive robot with joystick
+        m_Tank.addJoystick(m_Joystick);
+
+        // Toggle homing with Y button
+        m_Joystick.addHandler([this](HID::JButton button, bool pressed) {
```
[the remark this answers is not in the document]
alexdewar: The button handling we could do in the loop -- might be a bit cleaner. The axis handling stuff needs to be done this way though, so we can override the motor control if the robot's homing (so we don't e.g. accidentally bump the joystick).

examples/robot_positioner/robot_positioner.cc
```diff
@@ -0,0 +1,227 @@
...
+            if (m_StateMachine.getCurrentState() == Homing) {
+                std::cout << "Ignoring joystick while homing" << std::endl;
+                return true;
+            } else {
+                return false;
+            }
+        });
+
+        // Wait for Vicon system to spot our robot
+        while (m_Vicon.getNumObjects() == 0) {
```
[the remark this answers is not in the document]
alexdewar: Good idea!

robots/tank.h
alexdewar: It's not particularly practical ;-). I ended up sticking it in because it's handy for debugging, because you can control each wheel individually.

robots/tank_pid.h
alexdewar: It does work reasonably well as is, although we could slow it down a bit. The danger is we might get stuck on the wobbly floor if it's too slow... This is just for when the robot's heading is really far out, so it does a big turn to get back facing in roughly the right direction.

neworderofjamie: Hah that delicate balance is pretty representative of my attempts at robotting :smile:

neworderofjamie: Thinking about it, you're probably also operating at a higher framerate which probably helps

robots/tank.h
neworderofjamie: :+1:

examples/robot_positioner/robot_positioner.cc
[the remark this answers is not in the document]
neworderofjamie: You could just call ``m_Tank.drive(m_Joystick);`` in the states where you want manual control

tools/vicon_arena_constructor/vicon_arena_constructor.cc
alexdewar: Yeah I think you're right actually. It is fairly self-contained.

vicon/udp.h
alexdewar: Yeah that's true. We should just do ``m_Pose = pose;``

examples/robot_positioner/robot_positioner.cc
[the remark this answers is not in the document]
alexdewar: Good point.

examples/robot_positioner/robot_positioner.cc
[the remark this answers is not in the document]
alexdewar: Hmm I just tried this and it looks like the API is a bit broken -- sometimes I get an ID of 0, which is good, and sometimes I get 2, which breaks things.

I'll try to fix this in a different PR.

alexdewar:
Hey, it looks like the build's still failing because of a lack of SDL2 on your machine... Would you mind installing it?

examples/robot_positioner/robot_positioner.cc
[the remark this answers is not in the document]
neworderofjamie: Bugger! I feel sure I used it relatively recently...

alexdewar:
Actually my bad -- it was a new warning.

neworderofjamie:
I did also just install SDL :smile:

PR #98, opened 2019-03-23
Simulated ackerman car

tenxlenx:
Created a simulated ackerman-drive car and an example to demonstrate it's workings


alexdewar:
Hey Norbert.

I'm guessing you merged this by accident? I've temporarily reverted it.

I'll just have a look through now.

examples/Simulated_ackerman_car/simulated_ackerman.cc
```diff
@@ -0,0 +1,57 @@
...
+
+int
+bob_main(int, char **)
+{
+    //Robots::SimulatedTank<> robot(1.4_mps, 500_mm); // Tank agent
+    Robots::SimulatedAckermanCar<> car(1.4_mps, 500_mm);
+    Robots::CarDisplay display(10.2_m,160_mm);                     // For displaying the agent
+
+    auto mmps = 1_mps;
+    units::angle::degree_t deg = units::angle::degree_t(20);
```
alexdewar: Because you've included the ``units::literals`` namespace you can write this as ``20_deg``. Just for future reference.

tenxlenx:
hey,

I merged because it was so simple that I assumed it will be fine, I also need it to be able to work with it on another branch.

Let me know though what you think

Norbert

examples/Simulated_ackerman_car/simulated_ackerman.cc
```diff
@@ -0,0 +1,57 @@
...
+    while(display.isOpen()) {
+
+        
+        car.move(mmps, deg);
+        auto key = display.runGUI(car.getPose());
+        
+        
+        if (key.first == SDLK_UP) {
```
alexdewar: I'd do this as a ``switch`` block, because only one of these if-statements are going to be true.

robots/simulatedAckermanCar.h
```diff
@@ -0,0 +1,101 @@
+#pragma once
```
alexdewar: This is a minor nit, but could you rename this to ``simulated_ackerman_car.h``? Just because that's the naming convention we've used everywhere else.

examples/Simulated_ackerman_car/simulated_ackerman.cc
```diff
@@ -0,0 +1,57 @@
+// BoB robotics includes
```
alexdewar: Could you rename this folder to lowercase please?

alexdewar:
It's probably best to wait for review before merging. Not that big a deal (it's not like you're changing existing stuff) but it keeps things tidy.

If you need this code in another branch ASAP you can merge from this branch to the other one directly.

examples/Simulated_ackerman_car/simulated_ackerman.cc
alexdewar: ...``Simulated_ackerman_car``, that is.

examples/Simulated_ackerman_car/simulated_ackerman.cc
neworderofjamie: :+1:  (and that is **with** my hatred of switch statements)

PR #100, opened 2019-03-26
Consolidate makefiles

alexdewar:
Hey,

This is just a simple change adding a generic makefile for building all ``*.cc`` files in the current folder with the relevant options enabled (e.g. ``WITH_EIGEN`` etc.), as with the GeNN makefile. I'm still using ``bob_robotics.mk`` and haven't changed this side of things, so this change shouldn't break anything anywhere.


make_common/default.mk
```diff
@@ -0,0 +1,32 @@
...
+OBJECTS := $(foreach obj,$(basename $(SOURCES)),$(OBJECT_PATH)$(obj))
+DEPS    := $(foreach dep,$(basename $(SOURCES)),$(OBJECT_PATH)$(dep).d)
+
+.PHONY: all clean $(EXTRA_DEPS)
+
+all: $(OBJECTS)
+
+-include $(DEPS)
+
+$(OBJECTS): %: %.cc $(DEPS) $(EXTRA_DEPS)
```
neworderofjamie: What does the double : syntax mean?

alexdewar: It's needed so you can build the list of programs in ``$(OBJECTS)``: you need the weird syntax I guess because they don't have filename extensions.

And as I'm typing this I realise that variable should be called ``$(EXECUTABLES)`` or something instead!

PR #102, opened 2019-04-08
Pure pursuit controller

tenxlenx:
Hi Both,

in this update, I was working on pure pursuit controller. A controller which controls an ackerman driven car to follow a path. For this I have created an example program to demonstrate its workings. 

I had to modify **car_display.h** class:
- clearScreen() function was added to be able to clear the screen before drawing. In the same time, renderClear() was removed from draw(), so it needs to be called manually. getRenderer() was added. These changes allowed to add additional graphics elements to be drawn.
- mmToPixel() and pixelToMM() was moved to public so it can be used 

- added simulatedAckermanCar.h class which allows to simulate an ackerman driven car

let me know what do you think and what needs changing before merging


robots/car_display.h
```diff
@@ -107,6 +107,64 @@ class CarDisplay
         return out;
     }
 
+    std::vector<float> getMouseClickPixelPosition() const
```
neworderofjamie: It'd probably be better to return a ``std::array<float,2>`` here

robots/car_display.h
```diff
@@ -107,6 +107,64 @@ class CarDisplay
...
+    {
+        std::vector<float> out;
+        out.push_back(m_MouseClickPosition[0]);
+        out.push_back(m_MouseClickPosition[1]);
+        return out;
+    }
+
+    //! draws the rectangle at the desired coordinates
+    void drawRectangleAtCoordinates(const SDL_Rect &rectangle)
```
neworderofjamie: Maybe you could abbreviate this to ``drawRectangle``

robots/car_display.h
```diff
@@ -107,6 +107,64 @@ class CarDisplay
...
+
+    //! change from millimeter to pixel
+    void mmToPixel(const millimeter_t x, const millimeter_t y, int &xPixel, int &yPixel) const
+    {
+        xPixel = static_cast<int>(x / m_MMPerPixel) + (WindowWidth / 2);
+        yPixel = -static_cast<int>(y / m_MMPerPixel) + (WindowHeight / 2);
+    }
+
+    //! draw lines between a list of points 
+    void drawLinesBetweenRects(std::vector<SDL_Rect> listRects, SDL_Renderer *renderer) {
```
neworderofjamie: I'm not sure how general the functionality to draw lines between the top-left corner of rectangles is :smile: maybe move this into the example? Also the rects should be passed as a reference i.e. ``const std::vector<SDL_Rect> &listRects`` so they don't get copied every time its called

robots/purePursuitController.h
```diff
@@ -0,0 +1,169 @@
...
+
+    }
+
+    PurePursuitController(millimeter_t lookahead, millimeter_t wheelBaseLength) : 
+                          lookAheadDistance(lookahead), wheelBase(wheelBaseLength) {
+
+    }
+
+    //! set waypoints which forms a path to be followed
+    void setWayPoints(std::vector<Vector2<millimeter_t>> wp) {
```
neworderofjamie: Again, these should be passed as ``const std::vector<Vector2<millimeter_t>> &wp``

robots/purePursuitController.h
```diff
@@ -0,0 +1,169 @@
...
+        std::vector<millimeter_t> lookahead;
+
+        if (wayPoints.size() > 1) {
+            for (unsigned int i = 0; i < wayPoints.size()-1; i++) {
+
+                // path points
+                Vector2<millimeter_t> segmentStart = wayPoints.at(i);
+                Vector2<millimeter_t> segmentEnd = wayPoints.at(i+1);
+
+                std::vector<millimeter_t> p1,p2;
```
neworderofjamie: Are these variable ever used?

robots/purePursuitController.h
```diff
@@ -0,0 +1,169 @@
...
+  
+    //! calculates the turning angle needed to follow the path 
+    degree_t getTurningAngle(millimeter_t x, millimeter_t y, radian_t heading) {
+        std::vector<millimeter_t> lookPoint = getLookAheadPoint(x,y, lookAheadDistance);
+        if (!lookPoint.empty()) return computeTurningAngle(x,y, lookPoint.at(0), lookPoint.at(1), heading);
+        return 0_deg; 
+    }
+
+    //! calculates the look-ahead point the robot follows
+    std::vector<millimeter_t> getLookAheadPoint(millimeter_t x, millimeter_t y, millimeter_t r) {
```
neworderofjamie: It would be nicer if this returned ``Vector2<millimeter_t> ``

robots/purePursuitController.h
```diff
@@ -0,0 +1,169 @@
...
+
+// sign function
+template <typename T> int sgn(T val) {
+    return (T(0) < val) - (val < T(0));
+}
+
+namespace BoBRobotics {
+namespace Robots {
+
+using namespace units::math;
```
neworderofjamie: using namespaces in headers is generally a bad idea as they 'pollute' anything that includes the header. For instance, if I include this and am writing my own code in ``BoBRobotics::Robots``, I wouldn't expect ``sin`` to do something different than normal. In this case, you are only using the literals in ``getTurningAngle`` and the maths functions in ``getLookAheadPoint`` and ``computeTurningAngle`` so you only need to put those ``using`` statements there.

robots/purePursuitController.h
```diff
@@ -0,0 +1,169 @@
...
+
+        if (wayPoints.size() > 1) {
+            for (unsigned int i = 0; i < wayPoints.size()-1; i++) {
+
+                // path points
+                Vector2<millimeter_t> segmentStart = wayPoints.at(i);
+                Vector2<millimeter_t> segmentEnd = wayPoints.at(i+1);
+
+                std::vector<millimeter_t> p1,p2;
+                auto p1x = segmentStart.x() - x;
```
neworderofjamie: @alexdewar - it would be really nice if we had at least the non-ambiguous arithmetic operators i.e. + and - on Vectors

examples/PurePursuitController/purePursuitControllerExample.cc
```diff
@@ -0,0 +1,146 @@
...
+            rekt.w = 5;
+            rekt.h = 5;
+            currentX = xp;
+            currentY = yp;
+            rekt_list.push_back(rekt); // save rectangle
+            
+            // we store the physical unit coordinate in the waypoint list
+            display.pixelToMM(xp, yp,  xMM,  yMM);
+            Vector2<millimeter_t> coords(xMM,yMM);
+            wpCoordinates.push_back(coords);
```
neworderofjamie: You should be able to do these last two lines in one:
``wpCoordinates.emplace_back(xMM,yMM)``

examples/PurePursuitController/purePursuitControllerExample.cc
```diff
@@ -0,0 +1,146 @@
...
+        
+        // set waypoints in controller
+        controller.setWayPoints(wpCoordinates);
+
+        // run a GUI step    
+        auto key = display.runGUI(car.getPose());
+        display.clearScreen();
+
+        // draw all the waypoints on the screen
+        for (unsigned int i =0; i < rekt_list.size(); i++) {
```
neworderofjamie: when you don't actually need the index you can just go:
```c++
for (const auto &r : rekt_list) {
    display.drawRectangleAtCoordinates(r);
}
```

examples/PurePursuitController/Makefile
```diff
@@ -0,0 +1,20 @@
+NO_OPENCV := 1
+WITH_MATPLOTLIBCPP := 1
+WITH_SDL2 := 1
+include ../../make_common/bob_robotics.mk
```
alexdewar: I've simplified the makefile system for BoB robotics now -- hang on and I'll push a simpler version of this to your branch

examples/PurePursuitController/purePursuitControllerExample.cc
```diff
@@ -0,0 +1,146 @@
+/*
```
alexdewar: Would you mind changing the names of this folder and the ``*.h`` and ``*.cc`` files to use snake case please? i.e.
``examples/pure_pursuit_controller/pure_pursuite_controller_example.cc``

It's just that that's what we use elsewhere and it's good to be consistent.

examples/PurePursuitController/purePursuitControllerExample.cc
```diff
@@ -0,0 +1,146 @@
...
+
+    float currentX = 0, currentY = 0;                      // current mouse coordinate click
+
+    // adding the first coordinate
+    //-------------------------------------------------------------------
+    millimeter_t xMM, yMM;
+    display.pixelToMM(currentX, currentY,  xMM,  yMM);
+    std::vector<Vector2<millimeter_t>> wpCoordinates; 
+    Vector2<millimeter_t> firstCoord(xMM, yMM);
+    wpCoordinates.push_back(firstCoord);
```
alexdewar: You could use ``emplace_back`` here:
```
    wpCoordinates.emplace_back(xMM, yMM);
```

examples/PurePursuitController/purePursuitControllerExample.cc
```diff
@@ -0,0 +1,146 @@
...
+    //-------------------------------------------------------------------
+
+    auto mmps = 0_mps;    // speed of robot car
+    degree_t deg = 0_deg; // angle of steering of robot car
+    millimeter_t lookaheadDistance = 1000_mm; // lookahead distance
```
alexdewar: Does ``lookaheadDistance`` change its value? If not, mark it ``constexpr``

examples/PurePursuitController/purePursuitControllerExample.cc
```diff
@@ -0,0 +1,146 @@
...
+
+    auto wheelBase = car.getDistanceBetweenAxis(); // distance between wheel bases
```
alexdewar: ``const``?

examples/PurePursuitController/purePursuitControllerExample.cc
alexdewar: :+1:

examples/PurePursuitController/purePursuitControllerExample.cc
```diff
@@ -0,0 +1,146 @@
...
+            display.pixelToMM(xp, yp,  xMM,  yMM);
+            Vector2<millimeter_t> coords(xMM,yMM);
+            wpCoordinates.push_back(coords);  
+        }
+        
+        // set waypoints in controller
+        controller.setWayPoints(wpCoordinates);
+
+        // run a GUI step    
+        auto key = display.runGUI(car.getPose());
```
alexdewar: ``const``?

examples/PurePursuitController/purePursuitControllerExample.cc
```diff
@@ -0,0 +1,146 @@
...
+            display.drawRectangleAtCoordinates(rkt);
+            rektVec.push_back(rkt);
+            rektVec.push_back(rkt_rob);
+            // draw line between robot and lookahead point
+            display.drawLinesBetweenRects(rektVec, display.getRenderer());
+        }
+
+        // calculate turning angle with controller
+        auto turningAngle = controller.getTurningAngle(robx, roby, heading);
+        degree_t turningAngleDeg = turningAngle;
```
alexdewar: These two lines could be replaced with:
```
const degree_t turningAngleDeg = controller.getTurningAngle(robx, roby, heading);
```

robots/car_display.h
alexdewar: :+1:

robots/car_display.h
```diff
@@ -107,6 +107,64 @@ class CarDisplay
...
+    {
+        SDL_SetRenderDrawColor(m_Renderer, 0, 0, 255, 255);
+        SDL_RenderFillRect(m_Renderer, &rectangle);
+    }
+
+    //! return the current SDL_Renderer pointer if we want to draw additional graphics
+    //! elements to the screen.
+    SDL_Renderer* getRenderer() {
+        return m_Renderer;
+    }
```
alexdewar: If it were me, I'd return a reference rather than a pointer here:
```
SDL_Renderer &getRenderer() {
    return *m_Renderer;
}
```

But it's up to you.

robots/purePursuitController.h
```diff
@@ -0,0 +1,169 @@
...
+
+// Standard C++ includes
+#include <chrono>
+#include <iostream>
+#include <thread>
+
+// sign function
+template <typename T> int sgn(T val) {
+    return (T(0) < val) - (val < T(0));
+}
```
alexdewar: I'd make this function a private ``static constexpr`` function in your class below just so we don't have stray functions floating around (when you can get name collisions)

robots/purePursuitController.h
alexdewar: Agreed. But just to add that I think the ``using namespace units::literals;`` below is fine. String literals are a bit of a special case, I feel.

robots/purePursuitController.h
alexdewar: :+1:

robots/purePursuitController.h
alexdewar: Yeah I agree. Also being able to do e.g. ``myVector * 2`` would also be handy. I'll have a go at this in a separate PR at some point soon.

robots/purePursuitController.h
```diff
@@ -0,0 +1,169 @@
...
+            return lookahead;
+        }
+        return lookahead;
+    }
+
+private:
+
+    millimeter_t lookAheadDistance;                 // the distance to look ahead
+    millimeter_t wheelBase;                         // length between wheel bases
+    std::vector<Vector2<millimeter_t>> wayPoints;   // list of waypoint coordinates
```
alexdewar: I'd rename these members to ``m_LookaheadDistance``, ``m_WheelBase`` and ``m_WayPoints`` to be in keeping with the style we use elsewhere in BoB robotics. It is handy to know if a variable is a member just by looking at the name.

neworderofjamie: :+1:

alexdewar:
Hey @tenxlenx. I've merged in the master branch and simplified your makefile. Hope that's ok.

Also, I noticed you deleted your simulated Ackerman example... would you mind putting that back in? I think it's useful documentation.

examples/PurePursuitController/purePursuitControllerExample.cc
tenxlenx: of course no problem

examples/simulated_ackerman_car/simulated_ackerman_car.cc
```diff
@@ -0,0 +1,63 @@
...
+        display.clearScreen();
+
+        //move car
+        car.move(mmps, deg);
+
+        // run GUI 1 step and get user key command
+        const auto key = display.runGUI(car.getPose());
+        
+        // if key up -> add speed
+        if (key.first == SDLK_UP) {
```
alexdewar: Could you make these ifs into a switch expression instead?

robots/pure_pursuit_controller.h
```diff
@@ -0,0 +1,169 @@
...
+        const auto xlength = sin(bearing) * m_lookAheadDistance;
+
+        // calculating arc to turn
+        const auto kb = (2*xlength*m_wheelBase)/(pow<2>(m_lookAheadDistance));
+        const degree_t turningAngle = atan(kb);
+        return turningAngle;
+    }
+
+    // sign function
+    template <typename T> int sgn(T val) {
```
alexdewar: This should be ``static`` because it doesn't use any of the member vars/methods. Could also be marked ``constexpr``

alexdewar:
@tenxlenx feel free to merge when you're ready.

PR #111, opened 2019-05-11
Vicon: Fix for multiple objects

alexdewar:
This PR fixes the Vicon code for when we have multiple objects. The problem was that we were previously assuming that the order that objects are always in the same order in the packets we're receiving, which isn't always true (e.g. if object number 1 disappears momentarily, then object 2 will come first). So instead I've switched to identifying objects by their names (which we can now store as ``std::string``s).

I also added a ``waitUntilConnected()`` method, mostly to be used internally in the Vicon code (in ``getObjectData()`` etc.), which when first called waits for a very brief time to see if we get a packet, and if we don't, the Vicon system isn't running so it throws an exception. This way we don't need to repeatedly check if we have an object in calling code.


include/common/thread.h
```diff
@@ -5,6 +5,40 @@
 #include <thread>
 
 namespace BoBRobotics {
+//! A simple RAII wrapper for std::thread
```
neworderofjamie: Presumably this requires the thread to implement some sort of signalling to kill it before you join? I kinda feel the RAAIness just makes it easier to create a deadlock on exit

neworderofjamie: Presumably this requires the thread to implement some sort of signalling to kill it before you join? I kinda feel the RAAIness just makes it easier to create a deadlock on exit

include/vicon/udp.h
```diff
@@ -228,28 +242,52 @@ class ObjectReference
...
+        m_ShouldQuit = true;
     }
 
     //----------------------------------------------------------------------------
     // Public API
     //----------------------------------------------------------------------------
     void connect(uint16_t port)
     {
+        // Lock until we have at least one packet from Vicon system
+        m_ConnectionMutex.lock();
```
neworderofjamie: This seems a bit of an abuse of a mutex - I think ``std::condition_variable`` might be better here

include/vicon/udp.h
```diff
@@ -228,28 +242,52 @@ class ObjectReference
...
 {
-public:
-    UDPClient() {}
+private:
+    /*
+     * NB: The first member of this struct *should* be object ID, but the
+     * Vicon system seems to just give a value of zero for every item, which
+     * isn't terribly helpful. So instead we distinguish objects based on their
+     * names.
+     */
```
neworderofjamie: If this is the case - which sucks - it might be better to switch the whole thing to using a ``std::map``/``std::unordered_map`` - my original design only really makes sense if the numeric ids do :disappointed:

include/vicon/udp.h
```diff
@@ -228,28 +242,52 @@ class ObjectReference
...
+        char objectName[24];
+        double position[3], attitude[3];
+    };
+    #pragma pack(pop)
+#else
+    struct __attribute__((__packed__)) RawObjectData {
+        uint8_t unused;
+        uint16_t itemDataSize;
+        char objectName[24];
+        std::array<double, 3> position, attitude;
```
neworderofjamie: I think this would be cleaner if you use an 'old fashioned' array for both - not sure what the bug you're working around here is but I don't think it's worth it :smile:

include/common/thread.h
alexdewar: You could end up with a deadlock, but the problem with C++ threads is if you just let them be destroyed without ``join()``ing or ``deatch()``ing them then the behaviour is undefined. I took this idea from Scott Meyers' book.

include/vicon/udp.h
alexdewar: I don't particularly understand condvars too well... but don't you need a mutex for those anyway?

include/vicon/udp.h
alexdewar: That's probably a good idea

include/vicon/udp.h
alexdewar: That's what I meant to do... the ``#ifdef`` is just because it's a packed struct and you do this differently in gcc and MSVC

include/vicon/udp.h
neworderofjamie: Very good point :smile:

include/common/thread.h
neworderofjamie: I dunno - I think on any platform with an OS (rather than bare metal), the OS's going to kill any lingering threads - in the same way it frees any outstanding heap allocations - bad practise but imo maybe a better default than deadlocking

include/vicon/udp.h
neworderofjamie: Ahh ok - I was wondering if it was some sort of combination of a VC++ bug regarding packing std::array/std::array having some extra debug members on VC++ (which I **think** could be allowed). https://gist.github.com/PhilCK/1534763 could be nicer

alexdewar: Good idea.

examples/vicon/vicon.cc
```diff
@@ -17,16 +18,25 @@ int main(int argc, char **argv)
...
+     * name, otherwise return first discovered object.
+     */
+    ObjectData objectData = [&]() {
+        if (argc > 1) {
+            return vicon.getObjectData(argv[1]);
+        } else {
+            return vicon.getObjectData();
+        }
+    }();
+    const std::string objectName = objectData.getName();
```
neworderofjamie:
```c++
const std::string objectName = (argc > 1) ? vicon.getObjectData(argv[1]) : vicon.getObjectData();
```
might be a little clearer...

include/vicon/udp.h
```diff
@@ -399,13 +385,13 @@ class UDPClient
...
         using namespace units::length;
         using namespace units::angle;
-        m_ObjectData[id].update(frameNumber,
-                                { { millimeter_t(data.position[0]),
-                                    millimeter_t(data.position[1]),
-                                    millimeter_t(data.position[2])},
-                                  { radian_t(data.attitude[2]),
-                                    radian_t(data.attitude[0]),
-                                    radian_t(data.attitude[1]) } });
+        m_ObjectData.at(objectName).update(frameNumber,
```
neworderofjamie: you could save a second lookup by not having ``pos`` be const, setting it to ``m_ObjectData.emplace(...).first`` and calling ``pos->second.update(...)``

include/vicon/udp.h
```diff
@@ -286,61 +315,65 @@ class UDPClient
...
+         waitUntilConnected();
+         std::lock_guard<std::mutex> guard(m_ObjectMutex);
+         return m_ObjectData.begin()->second;
+     }
+
+     auto getObjectReference(Stopwatch::Duration timeoutDuration = 10s) const
+     {
+         waitUntilConnected();
+         std::lock_guard<std::mutex> guard(m_ObjectMutex);
+         return ObjectReference<ObjectDataType>(*this,
```
neworderofjamie: I may be missing something but isn't this too many constructor parameters for an ``ObjectReference``?

include/vicon/udp.h
```diff
@@ -286,61 +315,65 @@ class UDPClient
...
+     }
+
     //! Returns an object whose pose is updated by the Vicon system over time
-    auto getObjectReference(unsigned int id,
-                            Stopwatch::Duration timeoutDuration = Stopwatch::Duration::max())
+    auto getObjectReference(const std::string& name,
+                            Stopwatch::Duration timeoutDuration = 10s) const
     {
-        return ObjectReference<ObjectDataType>(*this, id, timeoutDuration);
+        return ObjectReference<ObjectDataType>(*this,
```
neworderofjamie: Shouldn't this version also include:
```c++
 waitUntilConnected();
std::lock_guard<std::mutex> guard(m_ObjectMutex);
```
?

examples/vicon/vicon.cc
alexdewar: Good point. Not the clearest code...

include/vicon/udp.h
alexdewar: Good idea

include/vicon/udp.h
alexdewar: Oh yeah :-/

include/vicon/udp.h
alexdewar: Not if we get rid of the unnecessary use of ``m_ObjectData`` here.

include/vicon/udp.h
```diff
@@ -40,9 +40,10 @@ class ObjectData
     using millimeter_t = units::length::millimeter_t;
 
 public:
-    ObjectData(const std::string &objectName)
-      : m_Name(objectName)
-    {}
+    ObjectData(const char *objectName)
```
neworderofjamie: As we are using fixed size arrays you can use this slightly aggregious syntax to make that explicit:

```c++
ObjectData(char (&objectName)[24])
{
```

include/vicon/udp.h
```diff
@@ -182,12 +183,13 @@ class ObjectReference
 
 public:
     ObjectReference(UDPClient<ObjectDataType> &client,
-                    const std::string &objectName,
+                    const char *objectName,
```
neworderofjamie: Ditto

include/vicon/udp.h
```diff
@@ -248,7 +283,7 @@ class UDPClient
     BOB_PACKED(struct RawObjectData {
         uint8_t unused;
         uint16_t itemDataSize;
-        char objectName[24];
+        CharArray<24> objectName;
```
neworderofjamie: I still don't see why this is necessary - especially as it leads to the whole COMMA grossness. I don't see any reason not to have this as a ``char[24]`` and the map's key as a ``CharArray<24>``. You've added a ``const char *`` constructor so you won't even need to change anything else.ee

include/vicon/udp.h
```diff
@@ -326,8 +361,14 @@ class UDPClient
     ObjectDataType getObjectData(const std::string &name)
     {
         waitUntilConnected();
+
+        // Convert to fixed-size char array
+        BOB_ASSERT(name.size() < 24);
+        char bytes[24];
+        strcpy(bytes, name.c_str());
```
neworderofjamie: Does the cmake build system define ``_CRT_SECURE_NO_WARNINGS`` on windows? If not, using, ``strcpy`` and its insecure friends results in compile errors.

include/vicon/udp.h
```diff
@@ -427,7 +467,8 @@ class UDPClient
                 // Lock mutex
                 std::lock_guard<std::mutex> guard(m_ObjectMutex);
 
-                const auto objectData = reinterpret_cast<RawObjectData *>(&buffer[5]);
+                auto objectData = reinterpret_cast<RawObjectData *>(&buffer[5]);
+                objectData->objectName[23] = '\0'; // Make sure string is null-terminated
```
neworderofjamie: If the strings the vicon system returns aren't actually properly null-terminated and we always add a null terminator to the end it might actually be better to copy these strings with ``std:::copy_n`` to avoid Windows whining about insecure functions etc

include/vicon/udp.h
alexdewar: Good point.

include/vicon/udp.h
alexdewar: Yep I've defined it in the cmake file

include/vicon/udp.h
alexdewar: Maybe... There is one place where I pass in a ``std::string`` from a call site so this would mean copying it to a ``char[24]`` first -- but I agree this is cleaner

PR #113, opened 2019-05-20
More maps

neworderofjamie:
Totally forgot about this branch but it adds some more helper functions to the map coordinates library (and is particularly desperate for cmaking as it needs a nasty hack atm)


alexdewar:
All looks good! Happy to cmake-ify things when we finish work on that branch.

neworderofjamie:
cool - proper modulification would definitely help with all the NO_HEADER_DEFINITIONS crud and might even mean the rothamsted reconstruction will build on windows..

PR #114, opened 2019-05-20
In-silico fixes

neworderofjamie:
* Include ``m_BeginColumn`` in ``columnToHeading`` calculation
* Use ``columnToHeading`` rather than adhocness in ``InfoMaxRotater::getHeading``

### DIFF
--- a/navigation/infomax.h
+++ b/navigation/infomax.h
@@ -34,7 +34,6 @@ namespace Navigation {
 template<typename FloatType = float>
 class InfoMax : public VisualNavigationBase
 {
-    using radian_t = units::angle::radian_t;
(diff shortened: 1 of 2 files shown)


navigation/infomax.h
```diff
@@ -196,24 +195,32 @@ class InfoMaxRotater : public InfoMax<FloatType>
...
+        const size_t bestIndex = std::distance(m_RotatedDifferences.cbegin(), el);
+
+        // Convert this to an angle
+        radian_t heading = rotater.columnToHeading(bestIndex);
+        while (heading <= -180_deg) {
+            heading += 360_deg;
+        }
+        while (heading > 180_deg) {
+            heading -= 360_deg;
+        }
```
alexdewar: We also have a function to do this in common/circstat.h

PR #115, opened 2019-05-22
Add support for two-wheeled simulated robot in Gazebo

alexdewar:
Hi @shinjanxp,

I've got a few small suggestions about the code and I thought it would probably be easiest to list them here. They're all basically cosmetic and stylistic things.


examples/gazebo_differential_drive_plugin/differential_drive.world
```diff
@@ -0,0 +1,26 @@
...
+    </include>
+    <!-- A ground plane -->
+    <include>
+      <uri>model://ground_plane</uri>
+    </include>
+
+    <!-- A testing model that includes the Velodyne sensor model -->
+    <model name="my_simple_cart">
+      <include>
+        <uri>model://simple_cart</uri>
```
alexdewar: As mentioned in the email, I think this may be the source of the trouble...

examples/gazebo_differential_drive_plugin/differential_drive_plugin.cc
```diff
@@ -0,0 +1,110 @@
...
+    /// \param[in] _model A pointer to the model that this plugin is
+    /// attached to.
+    /// \param[in] _sdf A pointer to the plugin's SDF element.
+    public: virtual void Load(physics::ModelPtr _model, sdf::ElementPtr _sdf)
+    {
+      // Safety check
+      if (_model->GetJointCount() == 0)
+      {
+        std::cerr << "Invalid joint count, DifferentialDrivePlugin not loaded\n";
+        return;
```
alexdewar: I don't know Gazebo that well, so I'm not sure if this is standard practice or not, but I'd suggest in this case that you throw an exception (e.g. ``runtime_error``) rather than just printing a message, as it is essentially a fatal error

examples/gazebo_differential_drive_plugin/differential_drive_plugin.cc
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      // Store the model pointer for convenience.
+      this->model = _model;
```
alexdewar: Note that you don't actually need to put the ``this->`` in C++ cf. some other languages

examples/gazebo_differential_drive_plugin/differential_drive_plugin.cc
```diff
@@ -0,0 +1,110 @@
...
+
+      // Create a topic name
+      std::string topicName = "~/" + this->model->GetName() + "/vel_cmd";
+
+      // Subscribe to the topic, and register a callback
+      this->sub = this->node->Subscribe(topicName, &DifferentialDrivePlugin::OnMsg, this);
+    }
+
+    /// \brief Set the velocity of the Velodyne
+    /// \param[in] _vel New target velocity
```
alexdewar: What are the units of velocity here? Is it m/s or are they angular velocities? Could you add it to the comments?

examples/gazebo_differential_drive_plugin/differential_drive_plugin.cc
```diff
@@ -0,0 +1,110 @@
...
+  {
+    /// \brief Constructor
+    public: DifferentialDrivePlugin() {}
+
+    /// \brief The load function is called by Gazebo when the plugin is
+    /// inserted into simulation
+    /// \param[in] _model A pointer to the model that this plugin is
+    /// attached to.
+    /// \param[in] _sdf A pointer to the plugin's SDF element.
+    public: virtual void Load(physics::ModelPtr _model, sdf::ElementPtr _sdf)
```
alexdewar: In C++ you don't need to mark every function as ``public``, ``private`` etc. (cf. e.g. Java). You usually just declare them all in blocks:
```C++
class MyClass {
public:
    void method1() { ... }
    void method2() { ... }

private:
    std::string m_SomeMember;
    void method3() { ... }
};
```

examples/gazebo_differential_drive_plugin/differential_drive_plugin.cc
```diff
@@ -0,0 +1,110 @@
...
+    /// \brief Handle incoming message
+    /// \param[in] _msg Repurpose a vector3 message. This function will
+    /// only use the x component.
+    private: void OnMsg(ConstVector3dPtr &_msg)
+    {
+      this->SetVelocity(_msg->x(), _msg->y());
+    }
+
+    /// \brief A node used for transport
+    private: transport::NodePtr node;
```
alexdewar: I don't *think* this needs to be a pointer. I think you could just write this as:
```
transport::Node node;
```

And then it'll be created automatically and you won't have to do this above:
```c++
this->node = transport::NodePtr(new transport::Node());
```

examples/gazebo_differential_drive_plugin/differential_drive_plugin.cc
```diff
@@ -0,0 +1,110 @@
...
+
+    /// \brief A subscriber to a named topic.
+    private: transport::SubscriberPtr sub;
+
+    /// \brief Pointer to the model.
+    private: physics::ModelPtr model;
+
+    /// \brief Pointer to the joint.
+    private: physics::JointPtr left_wheel_joint;
+    private: physics::JointPtr right_wheel_joint;
```
alexdewar: We tend to write member names as e.g. ``std::string m_MyMember``, with the ``m_`` in front and the camel case. It's our own weird thing but if you could change it that'd be great :smile:

examples/simulated_tank_gazebo/simulated_tank_gazebo.cc
```diff
@@ -0,0 +1,98 @@
...
+    /************************************Gazebo setup************/
+    // Load gazebo as a client
+    #if GAZEBO_MAJOR_VERSION < 6
+    gazebo::setupClient(0, 0);
+    #else
+    gazebo::client::setup(0, 0);
+    #endif
+    // Create our node for communication
+    gazebo::transport::NodePtr node(new gazebo::transport::Node());
+    node->Init();
```
alexdewar: As above, I think this could be written as: ``gazebo::transport::Node node;``

robots/gazebo_tank.h
```diff
@@ -0,0 +1,113 @@
...
+        updatePose();
+        BOB_ASSERT(left >= -1.f && left <= 1.f);
+        BOB_ASSERT(right >= -1.f && right <= 1.f);
+        m_Left = left * m_MaximumSpeed;
+        m_Right = right * m_MaximumSpeed;
+    }
+
+private:
+    Pose2<LengthUnit, AngleUnit> m_Pose;
+    Stopwatch m_MoveStopwatch;
```
alexdewar: I'd delete unused things like this when you get a chance. Not urgent though

robots/gazebo_tank.h
```diff
@@ -0,0 +1,113 @@
...
+  : public Tank
+{
+    using millimeter_t = units::length::millimeter_t;
+    using meters_per_second_t = units::velocity::meters_per_second_t;
+
+public:
+    GazeboTank(const meters_per_second_t maximumSpeed, const millimeter_t axisLength)
+      : m_MaximumSpeed(maximumSpeed)
+      , m_AxisLength(axisLength)
+    {}
```
alexdewar: I'm guessing that the maximum speed is either set in the Gazebo plugin or there is no maximum speed, so I guess we don't need this here. I'm guessing this is just leftover cruft from the ``SimulatedTank`` class.

examples/gazebo_differential_drive_plugin/differential_drive_plugin.cc
shinjanxp: This was modified from the Gazebo tutorial [Intermediate: Control plugin](http://gazebosim.org/tutorials?tut=guided_i5). I tried to set it to `transport::Node node;`, but the plugin crashed. It seems like the transport node is a shared pointer which gets passed to Gazebo's transport system when we call `node->Subscribe`. This is how it's defined in `gazebo/transport/TransportTypes.hh`:
```cpp
/// \def NodePtr
/// \brief Shared_ptr to Node object
typedef boost::shared_ptr<Node> NodePtr;
```
I think the `DifferentialDrivePlugin` object gets destroyed after `Load`, so when I tried to implement with just `transport::Node node`, the node object also got deleted leading to a crash.

alexdewar: Fair enough.

robots/gazebo_tank.h
shinjanxp: The way it's currently implemented is that the absolute wheel speeds should be sent via gazebo transport messages. So this maximum speed is multiplied with the joystick values to obtain an absolute speed which is finally transmitted to the Gazebo plugin. 
However, the `axisLength` value is not used at all, it was leftover from the SimulatedTank class. I'll remove it in the next commit.

shinjanxp:
All the changes have been incorporated into the gazebo branch.

robots/gazebo_tank.h
```diff
@@ -21,92 +21,39 @@ class GazeboTank
...
 
 private:
-    Pose2<LengthUnit, AngleUnit> m_Pose;
-    Stopwatch m_MoveStopwatch;
+
     const meters_per_second_t m_MaximumSpeed;
-    const millimeter_t m_AxisLength;
     meters_per_second_t m_Right{}, m_Left{};
-    std::pair <meters_per_second_t, meters_per_second_t> wheel_speeds;
+    std::pair <meters_per_second_t, meters_per_second_t> m_WheelSpeeds;
```
alexdewar: As far as I can see, ``m_WheelSpeeds`` just contains the same information as ``m_Right`` and ``m_Left``, right? In which case you'd only need one or the other.

examples/simulated_tank_gazebo/simulated_tank_gazebo.cc
```diff
@@ -53,34 +53,26 @@ bob_main(int, char **)
...
-        gazebo::msgs::Set(&msg, gazebo::math::Vector3((float)wheel_speeds.first, (float)wheel_speeds.second, 0));
+        gazebo::msgs::Set(&msg, gazebo::math::Vector3((float)wheelSpeeds.first, (float)wheelSpeeds.second, 0));
         #else
-        gazebo::msgs::Set(&msg, ignition::math::Vector3d((float)wheel_speeds.first, (float)wheel_speeds.second, 0));
+        gazebo::msgs::Set(&msg, ignition::math::Vector3d((float)wheelSpeeds.first, (float)wheelSpeeds.second, 0));
         #endif
         // Send the message
         pub->Publish(msg);
-        // std::cout<< wheel_speeds.first << "," << wheel_speeds.second << std::endl;        
+        // std::cout<< wheelSpeeds.first << "," << wheelSpeeds.second << std::endl;
```
alexdewar: FYI we are using this logging library internally: https://github.com/SergiusTheBest/plog

It allows you to print things at different log levels, so e.g.:
```c++
LOG_DEBUG << wheelSpeeds.first << "," << wheelSpeeds.second;
```
would only print when your build type is debug.

shinjanxp:
Incorporated the last 2 changes. Also, the gazebo model file had not been committed before because of a global gitignore for .sdf files.

alexdewar:
Hi Shinjan,

Just tried it and it's working great. Good work!

I've got one more teeny suggestion, which is that the robot should probably be stationary to begin with, rather than moving forwards. Other than that I'm happy to merge.

So I guess the next step should probably be to write a basic camera sensor?

shinjanxp:
Set initial velocity to 0. Also, run.sh will now deploy both Gazebo and the simulated_tank_gazebo executables.

PR #118, opened 2019-05-29
SpineML

neworderofjamie:
This PR integrates the new SpineML simulator library (see https://github.com/genn-team/genn/pull/242):
- New Make option to link against the SpineML simulator library
- Example combining various bits of BoB robotics simulation with a very dumb braitenberg vehicle implemented in SpineML
- Added the ability to extract first collision point from the collision detector - this is a bit ugly - any suggestions @alexdewar would be much appreciated


ajc158:
This might take a while - the package dependencies of this are turning out to be problematic on Mac.

neworderofjamie:
Maybe don't worry too much about that for now - @alexdewar is close to completing a new cmake build system in another branch which **should** make it easier to figure out dependencies

alexdewar:
Hopefully CMake should make things a bit easier, though that branch is yet to be tested on a Mac... @ajc158 how are you installing your dependencies? Are you using brew or something, or building from source?

common/collision.h
```diff
@@ -146,6 +149,11 @@ class CollisionDetector {
         // Check for collision
         for (int i = 0; i < m_RobotMap.size().area(); i++) {
             if (m_ObjectsMap.data[i] & m_RobotMap.data[i]) {
+                // Calculate collision point
+                const auto coordPixel = div(i, m_RobotMap.cols);
+                firstCollisionPosition.x() = m_XLower + (coordPixel.rem * m_GridSize);
+                firstCollisionPosition.y() = m_YLower + (coordPixel.quot * m_GridSize);
```
alexdewar: I made a vaguely similar addition in my robot control PR, except I returned which side of the object the robot had collided with. We def could do both, but perhaps I should make the last small changes to that branch and merge it first

examples/spineml_braitenberg/experiment1.xml
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" encoding="UTF-8"?><SpineML xmlns:UL="http://www.shef.ac.uk/SpineMLNetworkLayer" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.shef.ac.uk/SpineMLExperimentLayer" xsi:schemaLocation="http://www.shef.ac.uk/SpineMLNetworkLayer SpineMLNetworkLayer.xsd http://www.shef.ac.uk/SpineMLExperimentLayer SpineMLExperimentLayer.xsd"><Experiment name="robot" description="add a description for the experiment here"><Model network_layer_url="model.xml"/><Simulation duration="1" preferred_simulator="BRAHMS"><EulerIntegration dt="0.1"/></Simulation><ExternalInput target="LeftSensor" port="in" name="sensor left" tcp_port="50091" size="1" command="" host="0.0.0.0" timestep="0"/><ExternalInput target="RightSensor" port="in" name="sensor right" tcp_port="50091" size="1" command="" host="0.0.0.0" timestep="0"/><LogOutput name="steer left" target="SteerLeft" port="out" start_time="0" end_time="1e+8" tcp_port="50091" size="1" command="" host="0.0.0.0" timestep="0"/><LogOutput name="steer right" target="SteerRight" port="out" start_time="0" end_time="1e+8" tcp_port="50091" size="1" command="" host="0.0.0.0" timestep="0"/></Experiment></SpineML>
```
alexdewar: Maybe fix the whitespace in these files with your IDE so they're a bit more readable

examples/spineml_braitenberg/spineml_braitenberg.cc
```diff
@@ -0,0 +1,204 @@
...
+
+private:
+    sf::ConvexShape m_GreenShape, m_RedShape;
+};
+
+
+int bob_main(int, char **)
+{
+    // The x and y dimensions of the robot
+    // **THINK** if would be nice if this was part of a robot class?
```
alexdewar: One could certainly add this to the robot class, but the only thing is that a robot of a given class is not always going to be one size (e.g. if we mod the Lego robot, or with the ``TankNetSink`` class, which can be anything)

examples/spineml_braitenberg/spineml_braitenberg.cc
```diff
@@ -0,0 +1,204 @@
...
+    BOB_ASSERT(sensorLeft != nullptr);
+    BOB_ASSERT(sensorRight != nullptr);
+
+    while(true) {
+        // Move robot
+        collisionDetector.setRobotPose(robot.getPose());
+
+        // Normalise yaw
+        // **YUCK** why is this necessary? I would expect yaw to remain normalised
+        const auto robotYaw = normaliseAngle180(robot.getPose().yaw());
```
alexdewar: I'd expect this too... Maybe change the ``SimulatedTank`` class so it does this internally instead

examples/spineml_braitenberg/spineml_braitenberg.cc
```diff
@@ -0,0 +1,204 @@
...
+
+        // If robot has hit right wall
+        // **TODO** do this more elegantly with err maths
```
alexdewar: We could also check for wall collisions by extending the ``CollisionDetector`` class somehow -- I've been meaning to implement something along those lines because it would be handy

examples/spineml_braitenberg/spineml_braitenberg.cc
```diff
@@ -0,0 +1,204 @@
...
+        // Copy turn signals into sensor input buffers
+        std::fill(sensorLeft->getBufferBegin(), sensorLeft->getBufferEnd(), hitLeft);
+        std::fill(sensorRight->getBufferBegin(), sensorRight->getBufferEnd(), hitRight);
+
+        // Advance SpineML simulation
+        simulator.stepTime();
+
+        // Read steering signals
+        const float left = 10.0f * *steerLeft->getModelPropertyHostStateVarBegin();
+        const float right = 10.0f * *steerRight->getModelPropertyHostStateVarBegin();
```
alexdewar: Why ×10? Maybe a comment is needed

make_common/bob_robotics.mk
```diff
@@ -146,3 +146,11 @@ ifdef WITH_SFML_GRAPHICS
...
         LINK_FLAGS += `pkg-config --libs sfml-graphics`
 endif
+
+ifdef WITH_SPINEML_SIMULATION
+    ifndef GENN_PATH
+        $(error Environment variable GENN_PATH must be defined)
+    endif
+    CXXFLAGS += -I$(GENN_PATH)/include/spineml/simulator
+    LINK_FLAGS += -L$(GENN_PATH)/lib -lspineml_simulator -lspineml_common -ldl
+endif
```
alexdewar: Are we using GeNN v4 or v3 here? If we're using v4 then I guess we don't need ``GENN_PATH`` set, right?

examples/spineml_braitenberg/experiment1.xml
neworderofjamie: Ask @ajc158  about that one - for some reason, while all the other SpineML creator-created xml has line endings - experiments don't

common/collision.h
neworderofjamie: :+1:

examples/spineml_braitenberg/spineml_braitenberg.cc
neworderofjamie: Wouldn't the same argument apply to axle length?

examples/spineml_braitenberg/spineml_braitenberg.cc
neworderofjamie: yeah - that would be a nicer solution - might have a go at that

make_common/bob_robotics.mk
neworderofjamie: GeNN 4 so this isn't otherwise required - I just couldn't think of a better way of locating the GeNN lib directory.

alexdewar:
But if the user has installed GeNN with make install then shouldn't it be somewhere standard? If you just include spineml/simulator/whatever.h then you shouldn't need to put in a full path (I also feel that all the GeNN-related headers should probably live under /usr/include/genn or something so no extra include paths would be needed)

⁣Sent from BlueMail ​

alexdewar:
Heh, good point. We could replace the current ``getRobotWidth()`` function with a ``getRobotSize()`` (it could return e.g. a ``Vector3``)

⁣Sent from BlueMail ​

ajc158:
More that I'm discovering how out of date my macports is - it refuses to
install unless I upgrade and I am worried that might fail and break
everything I have that depends on the ports! Brew refuses to play nice with
macports installed and I'm invested now!

alexdewar:
Ugh, that's annoying. Would you mind trying to build some of the examples on the ``cmake`` branch for me, just to see if it works for you?

In theory, CMake should make cross-platform stuff easier, but in practice it sometimes fails to find various packages you have installed if they're older versions.

make_common/bob_robotics.mk
alexdewar: But if the user has installed GeNN with make install then shouldn't it be somewhere standard? If you just include spineml/simulator/whatever.h then you shouldn't need to put in a full path (I also feel that all the GeNN-related headers should probably live under /usr/include/genn or something so no extra include paths would be needed)

examples/spineml_braitenberg/spineml_braitenberg.cc
alexdewar: Heh, good point. We could replace the current ``getRobotWidth()`` function with a ``getRobotSize()`` (it could return e.g. a ``Vector3``)

make_common/bob_robotics.mk
neworderofjamie: That is optional though - personally I have it in  ~/genn for easy meddling :smile:

alexdewar: Although that does mean that even if you run ``make install`` you should still set ``GENN_PATH``. I feel like that should be optional, but it's no biggie as long as the docs tell you to do that.

neworderofjamie: Through the powers of bash I've come up with a much nicer solution - ``genn-buildmodel.sh`` should always be in your path and should always live in ``$GENN_PATH/bin`` therefore you can easily figure out ``GENN_PATH`` (if you've done ``make install`` it'll be ``/usr/local`` or similar)

alexdewar: Good thinking.

examples/spineml_braitenberg/spineml_braitenberg.cc
neworderofjamie: Fixed this properly now

examples/spineml_braitenberg/spineml_braitenberg.cc
neworderofjamie: Done!

examples/spineml_braitenberg/spineml_braitenberg.cc
neworderofjamie: Shall we leave this for a subsequent PR?

examples/spineml_braitenberg/spineml_braitenberg.cc
neworderofjamie: Shall we leave this for a subsequent PR?

common/collision.h
neworderofjamie: Shall we leave this for a subsequent PR?

examples/spineml_braitenberg/spineml_braitenberg.cc
alexdewar: Yep. Doesn't need doing right now.

examples/spineml_braitenberg/spineml_braitenberg.cc
alexdewar: Yeah, let's also put off this one till later.

neworderofjamie: #120

examples/spineml_braitenberg/spineml_braitenberg.cc
neworderofjamie: #121

neworderofjamie:
@alexdewar  how do we stop this being included in the automated tests? Requiring GeNN for that would add a whole layer of complexity

alexdewar:
Add an empty file called ``.test_skip`` to the directory. I haven't added this feature to the ``cmake`` branch yet, but can do.

PR #119, opened 2019-05-30
Stone CX updates

neworderofjamie:
- Model updated to GeNN 4 - all connectivity setup expressed using GeNN 
- Updated simulator so it works on GPU or CPU 
- Similarly updated Vicon version and full optical flow+IMU version
- Best performing dead reckoning one is however broken with same network compile errors the snapshot bot PR (https://github.com/BrainsOnBoard/snapshot_bot/pull/2) has. @alexdewar - if you could look at this it would be awesome - that way you can stream the visualization from the robot to the laptop


projects/stone_cx/robotVicon.cc
```diff
@@ -43,21 +42,7 @@ int main(int argc, char *argv[])
...
-        preferredAngleTL[i] = preferredAngleTL[8 + i] = (Parameters::pi / 4.0) * (double)i;
-    }
-
-    //---------------------------------------------------------------------------
-    // Build connectivity
-    //---------------------------------------------------------------------------
-    buildConnectivity();
-    
-    initstone_cx();
+    initializeSparse();
```
alexdewar: Alright, this is *really* nit-picky, but we are using British spelling elsewhere so maybe it should be ``initialiseSparse`` for consistency

neworderofjamie: Sadly GeNN has something of a mix but, in GeNN, initialisation occurs in two parts and the first part is triggered with an ``initialize`` function so it would seem like a bad idea to use the British spelling for the 2nd one...

alexdewar: Ahh no worries then. Let's be consistent with GeNN.

neworderofjamie: To be clear, that **is** a call into GeNN-generated code

alexdewar: Ohh I see.

alexdewar:
Are you happy to merge this? Or are you still working on this branch?

PR #122, opened 2019-05-31
Add support for remote controlled car robot

alexdewar:
I've just put in a PR for this so that we remember to review this later.

@tenxlenx


robots/norbot2.h
```diff
@@ -0,0 +1,99 @@
...
+      , m_turningAngle(0.0f)
+    {}
+
+    // destructor		
+    ~Norbot2() 
+    {
+        stopMoving();
+    }
+
+    void move(uint8_t speed, uint8_t turningAngle)
```
neworderofjamie: For consistancy with the other robots, the speed should be a normalised ``float`` i.e. between -1 and 1 and the steering should either be in the same form or as a ``degree_t`` if it actually corresponds to setting an absolute angle

arduino/Norbot_v2.0_arduino/Norbot_v2.0_arduino.ino
```diff
@@ -0,0 +1,120 @@
+
```
neworderofjamie: Can we call this something more consistent with the rest of the naming convention i.e. ``arduino/i2c_slave_code_norbot2/i2c_slave_code_norbot2.ino`` or something

arduino/Norbot_v2.0_arduino/Norbot_v2.0_arduino.ino
```diff
@@ -0,0 +1,120 @@
...
+  pinMode(THROTTLE, INPUT);
+ 
+  esc.attach(THROTTLE_CONTROL);      //attach esc to pin 9
+  steering.attach(STEERING_CONTROL); //attach steering to pin 6
+
+  // if Master sends data -> call receiveEvent()
+  Wire.onReceive(receiveEvent);
+
+  // if Master requests data -> call requestEvent() 
+  Wire.onRequest(requestEvent);
```
neworderofjamie: Do you have to setup an ``onRequest`` if it doesn't do anything?

examples/system_id_experiments/dataCollectorNorbot.cc
```diff
@@ -0,0 +1,81 @@
+#pragma once
```
neworderofjamie: This doesn't really do anything (and seems to be missing a makefile) so is there any point in including it as an example?

robots/norbot2.h
```diff
@@ -0,0 +1,99 @@
...
+    // stops the car 
+    void stopMoving() {
+        move(127, 90);
+    }
+   
+
+    //----------------------------------------------------------------------------
+    // Public API
+    //----------------------------------------------------------------------------
+    template<typename T, size_t N>
```
neworderofjamie: I don't think we really need these public ``read`` and ``write`` methods

arduino/Norbot_v2.0_arduino/Norbot_v2.0_arduino.ino
alexdewar: On a similar note, I feel like Norbot2 isn't the most descriptive name. How about ``RCCarRobot`` or something?

arduino/Norbot_v2.0_arduino/Norbot_v2.0_arduino.ino
```diff
@@ -0,0 +1,120 @@
...
+// called when data is sent through i2c to arduino
+// to update the motor values
+void receiveEvent(int numBytes) {
+  // check incoming bytes from master
+  uint8_t read_array[2];  
+  int i=0;
+  while(Wire.available()) {
+    read_array[i] = Wire.read();
+    i++;
+  }
```
alexdewar: What happens if three bytes are sent? I'd just do:
```c++
  uint8_t read_array[2] = { Wire.read(), Wire.read() };
```

arduino/Norbot_v2.0_arduino/Norbot_v2.0_arduino.ino
```diff
@@ -0,0 +1,120 @@
...
+    return true;
+  } 
+  return false;
+  
+}
+
+
+// called when data is sent through i2c to arduino
+// to update the motor values
+void receiveEvent(int numBytes) {
```
alexdewar: It looks like you're not using ``numBytes`` anywhere.

robots/norbot2.h
alexdewar: :+1: 

I'd also check that the user has passed valid arguments:
```c++
BOB_ASSERT(speed >= -1.f && speed <= 1.f);
BOB_ASSERT(turningAngle >= -45_deg && turningAngle <= 45_deg);
```

robots/norbot2.h
```diff
@@ -0,0 +1,99 @@
...
+    {
+        stopMoving();
+    }
+
+    void move(uint8_t speed, uint8_t turningAngle)
+    {
+      
+        // Cache left and right
+        m_forwardVelocity = speed;      // 0-127 backward, 128-255 forward
+        m_turningAngle = turningAngle;  // 55 left, 90 center, 125 right
```
alexdewar: I'd have the range of ``turningAngle`` be from -45° to 45° or whatever, so that 0° is the centre.

arduino/Norbot_v2.0_arduino/Norbot_v2.0_arduino.ino
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  // If motor input has changed
+  if(read_array[0] != movement[0]) {
+    movement[0] = read_array[0]; // throttle value
+  }
+
+  if(read_array[1] != movement[1]) {
+    movement[1] = read_array[1]; // steering value
+  }
```
alexdewar: You don't really need the if statements here. You could just do:
```c++
movement[0] = read_array[0];
movement[1] = read_array[1];
```

arduino/Norbot_v2.0_arduino/Norbot_v2.0_arduino.ino
```diff
@@ -0,0 +1,120 @@
...
+volatile bool updateMotors; // ture if we got commands from i2c
+int remoteControlSpeed;     // signal from remote control (speed)
+int remoteControlSteering;  // signal from remote control (steering)
+
+int neutralValue = 1500;
+byte STEERING = 3;
+byte THROTTLE = 5;
+
+byte STEERING_CONTROL = 10;
+byte THROTTLE_CONTROL = 11;
```
alexdewar: I'd mark all the constants as ``const``.

alexdewar:
Hey @tenxlenx, thanks for doing this. I still have a few really minor suggestions (sorry!), but it is looking a lot cleaner and it's definitely good to go for anything we want to do with it today. I'm happy to pitch in with tweaking things -- I suspect it will only take a few minutes between us.

I'll stick comments in anyway for future reference.

The other thing is that the CI is failing and generally we don't want to merge things into master in that case because it means the examples won't compile. It turns out it's not working in this case because I didn't have libserialport installed, but after I installed it there still were some compiler warnings which would also make Jenkins complain.

PR #123, opened 2019-05-31
Add support for dGPS

alexdewar:
As with #122, I'm just sticking this in so we remember to do something with it later.

@tenxlenx


common/gps.h
```diff
@@ -0,0 +1,99 @@
...
+
+        arcminute_t latmin;
+        arcminute_t longmin;
+        degree_t lat;
+        degree_t longitude;
+        std::string latDir;
+        std::string longDir;
+
+        // reading and parsing serial data
+        std::string serialString = Serial_reader::readSerialUSB(m_device_path);
```
neworderofjamie: You shouldn't re-open the serial port every time you read a line. Could you try cherry picking 23cf3e0c8c218f18569b0d78b73c6c31395594d8 which intergrates the libserial library? Then all you need to do is open it and set the baud rate in the constructor: 
```c++
Gps::Gps(const std::string &device) : m_SerialPort(device)
{
    m_SerialPort.Open(SerialPort::BAUD_9600);
}
```
and call a single method to get lines:
```c++
const std::string serialString = m_SerialPort.ReadLine(5000) 
```

examples/gps/gps_example.cc
```diff
@@ -0,0 +1,202 @@
+//
```
neworderofjamie: I don't think we need this example in bob_robotics

common/NMEA_parser.h
```diff
@@ -0,0 +1,131 @@
...
+                words.clear();
+            }       
+        }
+        return words;
+    }
+
+    
+    public:
+
+    static bool parseTextGPS(const std::string   &toParse,            // text to parse
```
neworderofjamie: It would be nicer if this let exceptions 'escape' rather than catching them all - that's generally the convention in BoB robotics and, that way, outside code can actually tell WHAT is going on

neworderofjamie: It might be nice if you integrated non-decimal latitude and longitudes into the map_coordinate library we already have (https://github.com/BrainsOnBoard/bob_robotics/blob/master/common/map_coordinate.h). Then you could just return some sort of struct rather than passing all these parameters by reference.

common/NMEA_parser.h
```diff
@@ -0,0 +1,131 @@
...
+                             degree_t            &latitude,           // latitude
+                             arcminute_t         &latitudeMinutes,    // minute part of latitude
+                             std::string         &latDirection,       // latitude direction North or South (N | S)
```
neworderofjamie: these directions should be ``char`` - ``std::string`` is super-wasteful for this

common/gps.h
```diff
@@ -0,0 +1,99 @@
...
+        return didParse;
+
+    }
+
+
+    public:
+
+    Gps(char *device_path) : m_device_path(device_path)  { }
+
+    bool getPosition(degree_t &lat, degree_t &lon, meter_t &altitude) {
```
neworderofjamie: It would be nicer if this returned a ``MapCoordinate::LatLon<MapCoordinate::WGS84>`` so it can be easily integrated with other stuff

alexdewar:
FYI @tenxlenx, the CI check is failing because we're getting compiler warnings (mostly because of unused variables): https://gen-ci.inf.sussex.ac.uk/job/BoB%20robotics/job/bob_robotics/job/gps/18/artifact/msg_INF900801/*view*/

common/NMEA_parser.h
```diff
@@ -0,0 +1,131 @@
+// NMEA message parser
```
alexdewar: Would you mind renaming this file to ``nmea_parser.h``, please, just so we can be consistent with other filenames

common/NMEA_parser.h
```diff
@@ -0,0 +1,131 @@
...
+type 1 or 9 update, null field when DGPS is not used
+14) Differential reference station ID, 0000-1023
+15) Checksum
+*/
+#pragma once
+#include<string>
+#include<vector>
+#include "../third_party/units.h"
+
+class NMEA_parser {
```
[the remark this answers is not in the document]
alexdewar: Also, could you rename it to ``NMEAParser`` please? We're using camel case for class names elsewhere.

common/NMEA_parser.h
```diff
@@ -0,0 +1,131 @@
...
+            string word;
+            pos = 0;   
+            while ((pos = f.find(w_delimiter)) != string::npos) {
+                word = f.substr(0,pos);
+                words.push_back(word);
+                f.erase(0, pos + w_delimiter.length());         
+            }
+
+            // if we find the one, we stop 
+            if (!words.empty() && words.at(0) == NMEA_sentence_id) {
```
alexdewar: Use ``words[0]`` instead. The ``at()`` method performs a bounds check and throws an exception if it's out of range, but we know there will be at least one element because you're checking for it.

common/NMEA_parser.h
alexdewar: :+1:

You could use the existing ``LatLon`` struct in that header. My suggestion would be to make a custom constructor for it that accepts the ``degree_t``, ``arcminute_t`` etc. and does the conversion to a ``degree_t`` for latitude and longitude.

common/NMEA_parser.h
alexdewar: Or even a ``bool``

common/NMEA_parser.h
```diff
@@ -0,0 +1,131 @@
...
+            latitudeMinutes = arcminute_t(std::stod(elements[2].substr(2,8)));     
+            latDirection = elements[3];
+            longitude = degree_t(std::stod(elements[4].substr(0,3)));
+            longitudeMinutes = arcminute_t(std::stod(elements[4].substr(3,9)));
+            longDirection = elements[5];
+            altitude = meter_t(stod(elements[9]));
+
+            // prse GNRMC for velocity
+            vector<string> elementsRMC= parseNMEAstring(toParse, "GNRMC"); // parse string 
+            velocity = knot_t(stod(elementsRMC[7]));
```
alexdewar: Does it really give you velocity in knots?! That seems an odd unit :-/

common/NMEA_parser.h
```diff
@@ -0,0 +1,131 @@
...
+            longDirection = elements[5];
+            altitude = meter_t(stod(elements[9]));
+
+            // prse GNRMC for velocity
+            vector<string> elementsRMC= parseNMEAstring(toParse, "GNRMC"); // parse string 
+            velocity = knot_t(stod(elementsRMC[7])); 
+        } catch(...) {
+            return false;
+        }   
+        return true;
```
alexdewar: I think @neworderofjamie has said something similar elsewhere, but I would just let the exceptions be thrown rather than returning a boolean. It's less code and means you don't have to rely on the user remembering to check the return value.

common/NMEA_parser.h
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    static bool parseTextForMiscData(const   std::string &toParse,    // text to parse     
+                                     int     &gpsQualityIndicator,    // 0 = Invalid, 1 = Valid SPS, 2 = Valid DGPS, 3 = Valid PPS, 4 = rtk gps
```
alexdewar: I feel like this should probably be an ``enum`` or ``enum class``.

common/gps.h
alexdewar: :+1:

common/gps.h
```diff
@@ -0,0 +1,99 @@
...
+        // reading and parsing serial data
+        std::string serialString = Serial_reader::readSerialUSB(m_device_path);
+        bool didParse = NMEA_parser::parseTextGPS(serialString, lat, latmin, latDir, longitude, longmin, longDir, m_altitude, m_velocity);
+
+        m_latitude = lat + latmin;
+        m_longitude = longitude + longmin;
+
+        // West and South has negative angles
+        if (longDir == "W") m_longitude = 0_deg - m_longitude;
+        if (latDir == "S")  m_latitude  = 0_deg - m_latitude;
```
alexdewar: This could just be: ``m_longitude = -m_longitude;`` etc.

common/gps.h
```diff
@@ -0,0 +1,99 @@
...
+    bool getGpsDescription() {
+        std::string serialString =  Serial_reader::readSerialUSB(m_device_path);
+        int quality, numsat;
+        double horizontalDilution;
+        bool didParse = NMEA_parser::parseTextForMiscData(serialString, quality, numsat, horizontalDilution);
+        m_gpsQuality = quality;
+        m_currentNumberOfSats = numsat;
+        m_horizontalDilution = horizontalDilution;
+
+        return didParse;
```
alexdewar: Another reason to use exceptions is so that you don't have to pass all these booleans up your call stack. Using exceptions generally makes life easier.

examples/camera_360/camera_360_test.cc
```diff
@@ -1,44 +1,88 @@
...
+    constexpr float joystickDeadzone = 0.25f;
+    
+    // Create joystick interface
+    BoBRobotics::HID::Joystick joystick(joystickDeadzone);
+    
+    // Create motor interface
+    BoBRobotics::Robots::Norbot robot;
+
+    // output file for gps 
+    std::ofstream file("out.csv");
```
alexdewar: ``std::ifstream`` and ``std::ofstream`` (annoyingly) don't throw exceptions if they fail, so I'd add a:
```c++
BOB_ASSERT(file.good());
```

examples/camera_360/camera_360_test.cc
```diff
@@ -49,17 +93,34 @@ int main()
...
 
-            // Show frame difference
-            cv::imshow("Original", originalImage);
-            cv::imshow("Unwrapped", outputImage);
-
-            if(cv::waitKey(1) == 27) {
+            // save camera
+            char filename[255];
+            sprintf(filename, "file_%u.jpg", frame);
+            cv::imwrite(filename, outputImage);
```
alexdewar: You can also use a ``std::stringstream`` for this:
```c++
std::stringstream ss;
ss < "file_" << frame << ".jpg";
cv::imwrite(ss.str(), outputImage);
```

examples/gps/gps_example.cc
```diff
@@ -0,0 +1,202 @@
...
+    osd = (OsmGpsMapLayer*) g_object_new (OSM_TYPE_GPS_MAP_OSD,
+                        "show-scale",TRUE,
+                        "show-coordinates",TRUE,
+                        "show-crosshair",TRUE,
+                        "show-dpad",TRUE,
+                        "show-zoom",TRUE,
+                        "show-gps-in-dpad",TRUE,
+                        "show-gps-in-zoom",TRUE,
+                        "dpad-radius", 30,
+                        NULL);
```
alexdewar: There is a really good C++ wrapper library for GTK: https://www.gtkmm.org/en/

This might be a better solution if we are planning on using GTK elsewhere.

examples/gps/gps_example.cc
```diff
@@ -0,0 +1,202 @@
...
+                                     50.8636,
+                                     -0.096366,
+                                     70);
+
+    // add updater function
+
+    MapAndGps *mgps = new MapAndGps(map, &gps);
+    g_timeout_add (1000, on_timeout, mgps);
+    gtk_main();
+    delete mgps;
```
alexdewar: You don't need this ``new`` and ``delete`` here.

Just do:
```c++
    MapAndGps mgps{ map, &gps };
    g_timeout_add (1000, on_timeout, &mgps);
    gtk_main();
```

alexdewar: Also in general you want to avoid using ``new`` and ``delete`` altogether in modern C++. Your alternatives are:

1. Allocate on the stack rather than the heap (like here)
2. If you *have* to allocate an array or something on the heap, use a wrapper object like a ``std::vector``
3. If you *need* a pointer to something that has to be allocated on the heap, use a smart pointer like a ``std::unique_ptr`` or something

examples/gps_bot/gps_bot.cc
```diff
@@ -0,0 +1,90 @@
...
+using namespace BoBRobotics;
+using namespace units::angle;
+using namespace units::length;
+
+void gpsThreadFunc(const char *device, std::mutex &mutex, degree_t &latOut, degree_t &lonOut, meter_t &altitudeOut)
+{
+    Gps gps("/dev/ttyACM0");
+
+    std::cout << "GPS created" << std::endl;
+    while(true)
```
alexdewar: Won't this thread run forever? I would use a ``std::atomic_bool`` as a stop flag, then you can set it to false on your main thread, then join this thread while you wait for it to finish.

examples/gps_bot/gps_bot.cc
```diff
@@ -0,0 +1,90 @@
...
+
+    // output file for gps 
+    std::ofstream file("out.csv");
+
+    Timer<> timer("");
+
+    degree_t lat,lon;
+    meter_t altitude;
+    std::mutex gpsMutex;
+    std::thread gpsThread(gpsThreadFunc, "/dev/ttyACM0", std::ref(gpsMutex), std::ref(lat), std::ref(lon), std::ref(altitude));
```
alexdewar: You don't need the ``std::ref``s here; the compiler knows they have to be references based on the function declaration. It won't hurt though.

examples/gps_bot/gps_bot.cc
```diff
@@ -0,0 +1,90 @@
...
+
+    {
+        unsigned int frame = 0;
+        for(frame = 0;; frame++) {
```
alexdewar: Are you using ``frame`` anywhere?

Also you could stick your stop condition here, e.g.:
```c++
while (!joystick.isDown(BoBRobotics::HID::JButton::B)) {
```

examples/gps/gps_example.cc
alexdewar: I personally quite like it. It's quite handy to have some code to check that your GPS unit is working and has found the satellites etc.

It could maybe go in ``tools/`` rather than ``examples/``.

examples/gps_bot/gps_bot.cc
neworderofjamie: Yes...I am to blame for the threading code in this - it was written at Capo Caccia, late at night :smile:

examples/gps_bot/gps_bot.cc
neworderofjamie: Is that true? https://stackoverflow.com/questions/36341724/c-thread-taking-reference-argument-failed-compile and the example code at http://www.cplusplus.com/reference/thread/thread/thread/ suggest that it isn't true...

examples/gps/gps_example.cc
neworderofjamie: If it used OpenCV or ImGUI sure, but my issue is with all the GTK dependencies it adds and the general grossness of GTK code....

examples/gps_bot/gps_bot.cc
alexdewar: Oops, my bad! Ignore me.

examples/gps/gps_example.cc
alexdewar: I think it's ok to have GTK as a dependency -- all Linux users will probably have it already.

I agree that GTK's C API is pretty nasty, so that's why I suggested using gtkmm. That way you get proper encapsulation and don't have to remember to free your ``g_object``s etc. I feel like GTK (or QT or something) is probably easier for making a proper GUI than imgui and I think OpenCV windows aren't very flexible.

alexdewar:
@tenxlenx I'd add ``/gps_example`` to a .gitignore file and delete the binary.

examples/gps/gps_example.cc
tenxlenx: this was just a hacky way to get the example to work to test, but I agree

examples/gps/gps_example.cc
tenxlenx: gtk is used by the mapping widget, otherwise I would never use this crappy library

common/NMEA_parser.h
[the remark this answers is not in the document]
alexdewar: Remember when you put functions straight into a namespace you have to mark them ``inline`` otherwise you get linker errors.

tenxlenx:
Could you guys have another look? I minimised the code a little bit and made the suggested fixes.

alexdewar:
Hey @tenxlenx, thanks for doing this. I still have a few really minor suggestions (sorry!), but it is looking a lot cleaner and it's definitely good to go for anything we want to do with it today. I'm happy to pitch in with tweaking things -- I suspect it will only take a few minutes between us.

I'll stick comments in anyway regardless of whether we get round to doing it today or not.

The other thing is that the CI is failing and generally we don't want to merge things into master in that case because it means the examples won't compile. It turns out it's not working in this case because I didn't have libserialport installed, but after I installed it there still were some compiler warnings which would also make Jenkins complain.

common/gps.h
```diff
@@ -0,0 +1,47 @@
...
+    
+    public:
+
+    Gps(const char *device_path) : m_device_path(device_path) {
+
+        try {
+            sreader.connect(device_path);
+            if (!sreader.isConnected()) throw "device not found";
+            m_device_path = device_path;
+        } catch(const char* error) {
```
alexdewar: I'd throw a ``std::runtime_error`` (``#include <stdexcept>``) rather than a ``const char *``, even though it is technically valid c++

alexdewar: Also I wouldn't actually catch and rethrow the exception here -- just let it pass up the call stack.

Is the issue that if you don't catch it here, you don't get to see what the error is? I know this is an issue on Windows, but not sure about Macs.

common/gps.h
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    GPSData getGPSData() {
+        GPSData data;
+        try {
+            data = NMEAParser::parseNMEA(sreader.readData()); 
+            return data;
+        } catch(const char *s) {
+            std::cout << "[Gps:getGPSPosition]: " << s <<  std::endl;
```
alexdewar: Again, I'd just let it pass up the call stack.

As a side note, for this kind of printing you can use plog:
```c++
LOG_ERROR << s;
```
It prints the function name and line number etc. automatically.

common/nmea_parser.h
```diff
@@ -0,0 +1,213 @@
...
+                throw "[NMEAParser:parseNMEAstring()] : cannot find NMEA id";
+            }
+            auto f = sentences[sentenceNumber];
+            // separating the sentence to words 
+            istringstream splitWord(f);
+            for (string word; getline(splitWord, word, w_delimiter); words.push_back(word));
+
+        } catch(const char *error) {
+            throw error;
+        } catch(...) {}
```
alexdewar: Here you definitely don't need to catch and rethrow, because you're not doing anything after catching anyway.

I also wouldn't do  ``catch(...) {}`` as a rule, because you almost never really want to just ignore all thrown exceptions (in this case, all exceptions that aren't of type ``const char *``)

common/nmea_parser.h
```diff
@@ -0,0 +1,213 @@
...
+#include "map_coordinate.h"
+
+namespace BoBRobotics
+{
+namespace GPS 
+{
+
+enum class GPSQuality
+{
+    INVALID,
```
alexdewar: I'd explicitly number these, i.e.: ``INVALID = 0``. You technically only need to number the first item and the rest will be numbered in order. Then you can do e.g. ``static_cast<GPSQuality>(0)`` and you won't need the ``switch`` statement below

common/nmea_parser.h
```diff
@@ -0,0 +1,213 @@
...
+    static GPSQuality getGpsQuality(const int &qualityId) {     
+        GPSQuality quality;  
+        switch(qualityId) {
+            case 0 : quality = GPSQuality::INVALID; break;
+            case 1 : quality = GPSQuality::GPSFIX; break;
+            case 2 : quality = GPSQuality::DGPSFIX; break;
+            case 3 : quality = GPSQuality::PPSFIX; break;
+            case 4 : quality = GPSQuality::RTK; break;
+            case 5 : quality = GPSQuality::FRTK; break;
+        }
```
alexdewar: I'd drop this function and just cast to a ``GPSQuality`` (see comment above)

common/nmea_parser.h
```diff
@@ -0,0 +1,213 @@
...
+            if (timeString== "") throw "Empty string when reading time";
+            int hour = stoi(timeString.substr(0,2));
+            int min  = stoi(timeString.substr(2,2));
+            int sec  = stoi(timeString.substr(4,2));
+            int msc  = stoi(timeString.substr(7,2));
+            TimeStamp time(hour, min, sec, msc);    
+            return time;
+        } catch(const char *s) {
+            throw s;
+        }  catch(...) { }
```
alexdewar: See comments for other ``catch`` bits

common/nmea_parser.h
```diff
@@ -0,0 +1,213 @@
...
+        meter_t             altitude;         
+        meters_per_second_t velocity;
+        double              horizontalDilution;
+        int                 numberOfSatelites;
+        int                 gpsQualityIndicator;
+        GPSQuality          qualityOfGps;
+        GPSData             data;
+        
+        try {
+            if (toParse == "") throw "Emtpy serial output";
```
alexdewar: *Really* minor thing, but FYI I think ``toParse.empty()`` is supposed to be the way to do this (so that you're not constructing another string object then doing a comparison)

common/nmea_parser.h
```diff
@@ -0,0 +1,213 @@
...
+            numberOfSatelites = stoi(elements[7]); 
+            horizontalDilution = stod(elements[8]);  
+            altitude = meter_t(stod(elements[9]));
+            vector<string> elementsRMC= parseNMEAstring(toParse, "GNRMC"); // parse GNRMC for velocity
+            velocity = units::velocity::knot_t(stod(elementsRMC[7])); // we change unit from knot to meters_per_second
+            qualityOfGps = getGpsQuality(gpsQualityIndicator);
+
+            // adding up the latitude and longitude parts
+            latitude = latitude + latitudeMinutes;
+            longitude = longitude + longitudeMinutes;
```
alexdewar: Or ``+=``

common/serial_reader.h
```diff
@@ -0,0 +1,104 @@
...
+            /* now clean the modem line and activate the settings for the port */
+            tcflush(fileDescriptor, TCIFLUSH);
+            tcsetattr(fileDescriptor,TCSANOW,&settings);
+        }
+        catch(const char *s) {
+            std::cout << "[SerialReader:connect()]: " << s <<  std::endl;
+        }
+    }
+
+    bool isConnected() { return connected; }
```
alexdewar: This method could be ``const``

common/serial_reader.h
```diff
@@ -0,0 +1,104 @@
...
+class SerialReader {
+    private:
+    int fileDescriptor;
+    bool connected = false;
+   
+    struct termios settings;
+
+
+    public:
+    SerialReader() {}
```
alexdewar: You should initialise ``fileDescriptor`` to ``-1``, because otherwise the user could call ``readData()`` before it's set

common/serial_reader.h
```diff
@@ -0,0 +1,104 @@
...
+#include <fcntl.h>
+#include <fstream>
+#include <unistd.h>
+#include <stdio.h>
+#include <sys/ioctl.h>
+#include <chrono>
+#include <thread>
+#include <errno.h>
+
+#include <libserialport.h> // cross platform serial port lib
```
alexdewar: Wait, are you using libserialport in this class anywhere? Or were you saying that it wasn't working for you?

examples/camera_360/camera_360_test.cc
```diff
@@ -1,44 +1,88 @@
 // BoB robotics includes
-#include "common/logging.h"
-#include "common/timer.h"
-#include "imgproc/opencv_unwrap_360.h"
-#include "video/panoramic.h"
+#include "../../common/logging.h"
```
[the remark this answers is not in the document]
alexdewar: @tenxlenx Just copy the folder to a new name, then ``git checkout`` the old example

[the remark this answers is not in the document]
alexdewar: ...or actually, as I guess we have a GPS example I suppose we don't need this one?

examples/gps/gps_example.cc
```diff
@@ -0,0 +1,46 @@
...
+            try {
+                BoBRobotics::GPS::GPSData data = gps.getGPSData();
+                BoBRobotics::MapCoordinate::GPSCoordinate coord = data.coordinate;
+                BoBRobotics::GPS::TimeStamp time = data.time;
+                
+                std::cout << "latitude: " << coord.lat.value()<< " longitude: " <<  coord.lon.value() << "[ "
+                         <<  time.hour.value() << ":" << time.minute.value() << ":" <<  time.second.value() << ":" << time.millisecond.value() << "] " << std::endl;
+                
+                i++;
+            } catch(...) { }
```
alexdewar: I wouldn't ignore the exceptions, otherwise there's not much point in throwing them in the first place :wink:

If they're fatal exceptions, just let them be thrown and kill the program cleanly. I guess there might be some cases where the errors would be non-fatal in which case it would be better to print a warning -- let's disucss

examples/gps/mapviewer.ui
```diff
@@ -0,0 +1,371 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
alexdewar: I guess we don't need this file any more?

examples/gps_bot/gps_bot.cc
```diff
@@ -0,0 +1,90 @@
...
+    Gps gps("/dev/ttyACM0");
+
+    std::cout << "GPS created" << std::endl;
+    while(true)
+    {
+        
+        // get gps data
+        degree_t lat,lon;
+        meter_t altitude;
+        bool didGetGps = gps.getPosition(lat,lon, altitude);
```
alexdewar: Looks like ``didGetGps`` isn't being used

examples/gps_bot/gps_bot.cc
```diff
@@ -0,0 +1,90 @@
...
+    constexpr float joystickDeadzone = 0.25f;
+    
+    // Create joystick interface
+    BoBRobotics::HID::Joystick joystick(joystickDeadzone);
+    
+    // Create motor interface
+    BoBRobotics::Robots::Norbot robot;
+
+    // output file for gps 
+    std::ofstream file("out.csv");
```
alexdewar: As above, I'd do a ``BOB_ASSERT(file.good());`` so you know whether it opened OK or not

common/nmea_parser.h
neworderofjamie: :+1:

common/nmea_parser.h
neworderofjamie: :+1:

examples/camera_360/camera_360_test.cc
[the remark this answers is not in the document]
neworderofjamie: that was more my point...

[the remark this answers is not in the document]
alexdewar: Ah yes. What I meant was that as Norbert's now made a GPS example this one should be reverted to how it was

tenxlenx:
Ah crap it was really late and forgot to delete some files not used. I wasnot really sure about which exceptions to throw so i went with wath seemed logical, but i see what the problem is with it, thanks Alex!

common/nmea_parser.h
```diff
@@ -203,9 +191,13 @@ class NMEAParser {
             data.horizontalDilution = horizontalDilution;
             data.gpsQuality = qualityOfGps;
             data.time = time; // UTC time
-        } catch(const char *error) {
-            throw error;
-        } catch(...) { }
+        } catch(GPSError &e) {
+            throw e;
```
alexdewar: You don't actually need to catch and rethrow this. Just leave it uncaught

common/nmea_parser.h
```diff
@@ -0,0 +1,205 @@
...
+    RTK     = 4,
+    FRTK    = 5
+};
+
+struct TimeStamp
+{
+    units::time::hour_t hour;
+    units::time::minute_t minute;
+    units::time::second_t second;
+    units::time::millisecond_t millisecond;
```
neworderofjamie: Aren't these double-precision times? This doesn't seem the right storage format to me - personally I'd keep them as ``int`` - I don't see how these would be at all useful in contexts we use the units library in...

common/nmea_parser.h
```diff
@@ -0,0 +1,205 @@
...
+        auto f = sentences[sentenceNumber];
+        // separating the sentence to words 
+        istringstream splitWord(f);
+        for (string word; getline(splitWord, word, w_delimiter); words.push_back(word));
+
+        return words;
+    }
+
+    // parsing the string time to numbers
+    static TimeStamp parseStringTime(std::string timeString) {
```
neworderofjamie: This could maybe go into a constructor of ``TimeStamp``

tenxlenx:
sorry I closed the PR, do we need it open?

common/nmea_parser.h
```diff
@@ -0,0 +1,205 @@
...
+        const char delimiter = '$';      // sentences separated by [$]
+        const char w_delimiter = ',';    // elements separated by  [,]
+
+        vector<string> sentences,words;
+        istringstream split(textToParse);
+        // separating the text to sentences
+        
+        if (textToParse.empty()) throw GPSError("Empty string to parse");
+
+        for (string each; getline(split, each, delimiter); sentences.push_back(each));
```
neworderofjamie: Why do you need to read all the sentences when you only use one? Do the outer loop and check if ``each`` contains ``NMEA_sentence_id``

alexdewar:
Hey @neworderofjamie. It looks like the CI is failing because Gazebo can't be found on your machine... Do you have it installed?

common/nmea_parser.h
alexdewar: Good point.

common/nmea_parser.h
alexdewar: :+1:

PR #125, opened 2019-06-02
Add support for two-wheeled Gazebo robot + camera

alexdewar:
Hi Jamie,

I thought it might be an idea to merge @shinjanxp's work on Gazebo integration as we go along rather than doing a big PR at the end. I have already looked through this code and chatted to Shinjan about it. Don't feel you have to go through this in detail but your insights are always welcome :-)


examples/simulated_tank_gazebo/simulated_tank_gazebo.cc
```diff
@@ -0,0 +1,104 @@
...
+#include <cstring>
+// Gazebo includes
+#include <gazebo/gazebo_config.h>
+#include <gazebo/transport/transport.hh>
+#include <gazebo/msgs/msgs.hh>
+#include <gazebo/common/common.hh>
+
+// Gazebo's API has changed between major releases. These changes are
+// accounted for with #if..#endif blocks in this file.
+#if GAZEBO_MAJOR_VERSION < 6
```
neworderofjamie: Can we maybe just not support gazebo < 6?

examples/simulated_tank_gazebo/simulated_tank_gazebo.cc
```diff
@@ -0,0 +1,104 @@
...
+    Robots::GazeboTank<> robot(5_mps); // Tank agent
+    HID::Joystick joystick(0.25f);
+    robot.controlWithThumbsticks(joystick);
+
+    std::cout << "Drive the car using the two thumbsticks: each stick is for one motor" << std::endl;
+
+    do {
+        // Refresh display
+        // car.setPose(robot.getPose());
+        auto wheelSpeeds = robot.getWheelSpeeds();
```
neworderofjamie: I feel this stuff should be wrapped up in the ``GazeboTank`` as it in the camera

examples/simulated_tank_gazebo/simulated_tank_gazebo.cc
alexdewar: Good idea.

examples/simulated_tank_gazebo/simulated_tank_gazebo.cc
alexdewar: Also a good idea!

robots/gazebo_tank.h
```diff
@@ -0,0 +1,53 @@
...
+
+public:
+    GazeboTank(const meters_per_second_t maximumSpeed)
+      : m_MaximumSpeed(maximumSpeed)
+    {}
+
+
+    const auto &getWheelSpeeds(){
+        return m_WheelSpeeds;
+    }
```
alexdewar: This function should be marked ``const``.

video/gazebocamerainput.h
```diff
@@ -0,0 +1,133 @@
...
+    : m_CameraTopic(topic), m_CameraName(cameraName)
+    {
+        // Load gazebo as a client
+        #if GAZEBO_MAJOR_VERSION < 6
+        gazebo::setupClient(0, 0);
+        #else
+        gazebo::client::setup(0, 0);
+        #endif
+        //Setup image subscription node
+        m_ImageNode= gazebo::transport::NodePtr(new gazebo::transport::Node());
```
alexdewar: This member could be initialised with the others as so:
```c++
m_ImageNode(new gazebo::transport::Node)
```

video/gazebocamerainput.h
```diff
@@ -0,0 +1,133 @@
...
+    GazeboCameraInput(gazebo::transport::NodePtr node, const std::string &topic="/gazebo/default/camera/link/camera/image", const std::string &cameraName = GazeboCameraDeviceName)
+      : m_CameraTopic(topic), m_CameraName(cameraName), m_ImageNode(node)
+    {
+        subscribeToGazeboCamera();
+    }
+
+    void subscribeToGazeboCamera(){
+        // Subscribe to the topic, and register a callback
+        m_HaveReceivedFrames = false;
+        m_ImageSub = m_ImageNode->Subscribe(m_CameraTopic, &GazeboCameraInput::OnImageMsg, this);
```
alexdewar: These members could also be initialised with the others. For ``m_HaveReceivedFrames`` you could initialise it C++11 style, by just writing ``std::atomic<bool> m_HaveReceivedFrames{ false };`` where you define it below, which would save on typing.

video/gazebocamerainput.h
```diff
@@ -0,0 +1,133 @@
...
+        std::lock_guard<std::mutex> lck(m_Mtx);
+
+        if(!m_HaveReceivedFrames.load()){
+            // Assuming image size remains constant at runtime, we can set the OutSize only once
+            m_OutSize.width = (int) msg->image().width();
+            m_OutSize.height = (int) msg->image().height();
+            m_HaveReceivedFrames.store(true);
+        }    
+        m_ReceivedImage.create(msg->image().height(), msg->image().width(), CV_8UC3);
+        memcpy(m_ReceivedImage.data, msg->image().data().c_str(), msg->image().data().length());
```
alexdewar: You don't need the separate ``m_OutSize`` member: you can just use ``m_ReceivedImage`` instead.

video/gazebocamerainput.h
```diff
@@ -0,0 +1,133 @@
...
+            return false;
+        std::lock_guard<std::mutex> lck(m_Mtx);
+        outFrame = m_ReceivedImage;
+        // If there's no error, then we have updated frame and so return true
+        return true;
+    }
+
+
+private:
+    const std::string m_CameraTopic;
```
alexdewar: Do you need to store this string once it's used?

video/gazebocamerainput.h
```diff
@@ -0,0 +1,133 @@
...
+    std::string m_CameraName;
```
alexdewar: I guess we don't need this member any more...?

video/gazebocamerainput.h
```diff
@@ -0,0 +1,133 @@
...
+    gazebo::transport::SubscriberPtr m_ImageSub;
+    gazebo::transport::NodePtr m_ImageNode;
+    cv::Size m_OutSize;
+    char *m_Data;
```
alexdewar: And I guess we don't need this one either...

robots/gazebo_tank.h
```diff
@@ -0,0 +1,53 @@
...
+
+// Third-party includes
+#include "../third_party/units.h"
+#include <utility> 
+namespace BoBRobotics {
+namespace Robots {
+using namespace units::literals;
+
+template<typename LengthUnit = units::length::millimeter_t,
+         typename AngleUnit = units::angle::degree_t>
```
alexdewar: We're not using these templated types anywhere, so I'd drop the templating. (Tbh I'm not sure it was even a good idea when I used it for the ``SimulatedTank`` class...)

video/gazebocamerainput.h
shinjanxp: I've shipped this out to common/gazebo_node.h in the latest commit.

video/gazebocamerainput.h
shinjanxp: Ok. Cool

.vscode/c_cpp_properties.json
```diff
@@ -8,10 +8,11 @@
...
                 "${GENN_PATH}/lib/include",
                 "${ARSDK_ROOT}/out/arsdk-native/staging/usr/include",
-                "/usr/include/python3.7m",
-                "/usr/lib/python3.7/site-packages/numpy/core/include",
                 "/usr/include/eigen3",
-                "/usr/include/opencv4"
+                "/usr/include/opencv2",
+                "/usr/include/opencv4",
+                "/usr/include/gazebo-9",
+                "/usr/include"
```
alexdewar: ``/usr/include`` and ``/usr/include/opencv2`` shoudn't be needed here. I think VS code should find them by default anyway.

alexdewar: Also, could you add my Python include back in please :wink:

video/gazebocamerainput.h
shinjanxp: Not really. We can remove the member by passing the string in the subscribeToGazeboCamera() function call

alexdewar: I think that would be cleaner. No need to copy the string if we're only using it once.

.vscode/c_cpp_properties.json
shinjanxp: Fixed in the latest commit

robots/gazebo_tank.h
shinjanxp: Actually, this function is no longer required since I've moved the driving logic into GazeboTank class. This function was earlier being used by simulated_tank_gazebo to set the wheel speed messages. Should we still keep it?

video/gazebocamerainput.h
alexdewar: :+1:

robots/gazebo_tank.h
alexdewar: Nah, let's drop it. We also don't need to store ``m_WheelSpeeds`` any more.

robots/gazebo_tank.h
```diff
@@ -0,0 +1,64 @@
...
+    {
+        // Publish to the  differential_drive_robot topic
+        pub =node->Advertise<gazebo::msgs::Vector3d>("~/differential_drive_robot/vel_cmd");;
+        // Wait for a subscriber to connect to this publisher
+        pub->WaitForConnection();
+    }
+
+    virtual meters_per_second_t getAbsoluteMaximumSpeed() const override
+    {
+        return 0.3* m_MaximumSpeed; //radius of the wheel is 0.3m. v = r × ω
```
alexdewar: I'm getting a compiler error here. The problem is that the units library won't let you convert between different kinds of units here (even though the equation you're using is correct). You have to get a ``double`` out and construct a ``meters_per_second_t`` object, like so:
```c++
return meters_per_second_t{ 0.3 * m_MaximumSpeed.value() };
```

shinjanxp:
I think all the changes have been incorporated. There seems to be some jenkins check which is failing and a conflict with `make_common/bob_robotics.mk`. I don't think I changed this file though.

common/gazebo_node.h
```diff
@@ -0,0 +1,28 @@
...
+#include <gazebo/msgs/msgs.hh>
+#include <gazebo/common/common.hh>
+
+// Gazebo's API has changed between major releases. These changes are
+// accounted for with #if..#endif blocks in this file.
+#if GAZEBO_MAJOR_VERSION < 6
+#include <gazebo/gazebo.hh>
+#else
+#include <gazebo/gazebo_client.hh>
+#endif
```
alexdewar: Let's drop support for Gazebo < v6 here.

alexdewar:
Hi Shinjan,

I've added one minor comment, but otherwise it's looking good!

Currently the CI test is failing because my machine has some local configuration issue... I'll look into this.

common/gazebo_node.h
shinjanxp: Sorry, I missed this. Fixed in commit 6045d5e

PR #126, opened 2019-06-02
Possible CMake fix for Ubuntu 16.04

alexdewar:
Hi @shinjanxp,

Does this fix the CMake issue you were having? Do the examples build now?

### DIFF
--- a/cmake/bob_robotics.cmake
+++ b/cmake/bob_robotics.cmake
@@ -178,9 +178,6 @@ macro(always_included_packages)
     # with the include path and link flags and it seems that this target isn't
     # "passed up" by add_subdirectory(), so we always include these packages on
     # the off-chance we need them.
-    if(NOT TARGET Eigen3::Eigen)
-        find_package(Eigen3 QUIET)
-    endif()
     if(NOT TARGET OpenMP::OpenMP_CXX)
         find_package(OpenMP QUIET)
     endif()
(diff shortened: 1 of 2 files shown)


shinjanxp:
Well, the good news is that it's able to find Eigen
```
-- Checking for module 'eigen3'
--   Found eigen3, version 3.3.7
```
But I'm getting a different error:
```
-- Configuring done
CMake Error at /home/shinjan/Workspaces/bob_robotics/cmake/bob_robotics.cmake:50 (add_executable):
  Target "simulated_tank_pid" links to target "OpenMP::OpenMP_CXX" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
Call Stack (most recent call first):
  simulated_tank_pid/CMakeLists.txt:3 (BoB_project)
```
This error is being thrown for almost all the examples.

alexdewar:
Ok, well at least that's progress :wink: 

I've managed to get it working in an Ubuntu 16.04 docker container, but it would be great if you could also test it. You might want to do ``export NO_I2C=1`` so that the build will skip all the i2c stuff.

neworderofjamie:
btw, I also have a Ubuntu 16 machine at home I can test things on

alexdewar:
Cool, thanks. I think it's mostly working now, except that now one example is failing with linking errors on my machine even though all I've changed is some compiler flags :angry:. Bloody CMake.

shinjanxp:
I think I may have some misconfiguration in my system because it's still failing. I've attached the terminal outputs here:
[build_with_i2c.log](https://github.com/BrainsOnBoard/bob_robotics/files/3250929/build_with_i2c.log)
[build_without_i2c.log](https://github.com/BrainsOnBoard/bob_robotics/files/3250930/build_without_i2c.log)

alexdewar:
Hi Shinjan,

It looks like the build is failing for a few reasons:
1. The I2C-based examples won't compile if ``NO_I2C`` is set
2. The Surveyor robot example is failing for presumably the same unknown reason that it's failing on my machine!
3. It does look like there's some configuration issue to do with OpenCV. It looks like you've got v2 and v3 installed and gcc seems not to like it (I guess there is the potential for linking against the wrong version of the libs, which will break things)

Do you need OpenCV v2 installed for anything? Uninstalling it would probably be the easiest solution, if that's an option.

shinjanxp:
Even after removing OpenCV2, I was getting errors. Apparently, my compiler isn't able to find this function
https://github.com/BrainsOnBoard/bob_robotics/blob/4c058c449dda26052d5cb2a22f8e791e49753350/include/vicon/udp.h#L475
So I changed it to `this->updateObjectData(frameNumber, data);` and now the examples build.

alexdewar:
Oh, that's weird. I've found that some versions of gcc require an extra ``this->`` in some places but not others.

I'll fix that then and merge this branch.

PR #127, opened 2019-06-04
Stone CX + misc updates

alexdewar:
Hey Jamie,

I managed to get the Stone CX model somewhat working on the Lego robot: it at least drives back towards where it came from, though sadly it's not doing the lovely ant-like looping behaviour we see in simulation (although perhaps if we had a big enough room it would eventually double back on itself...).

As it seems like we're not going to be using this demo any time soon, I thought it'd probably be fine to leave it in this mostly working state and put in a PR so we can merge the other bits I've implemented in the process of trying to make this work.

Commit summary:
- CMake-ify your updates to the Stone CX project
- Add ``OPTIONS`` param to BoB robotics modules and projects for setting compile time options (e.g. one can run ``cmake -DUSE_VICON=on ..`` to compile in Vicon stuff)
- Let users choose the type of tank robot they want to use with the ``TANK_TYPE`` CMake option or environment variable, including the option to control the robot over the network
- Add support for the Lego IMU and for streaming its readings over the network


include/os/screen.h
```diff
@@ -9,13 +9,13 @@
...
-// OpenCV
-#include <opencv2/opencv.hpp>
+// Standard C++ includes
+#include <utility>
 
 namespace BoBRobotics {
 namespace OS {
 namespace Screen {
-cv::Size
+std::pair<int, int>
```
neworderofjamie: How come?

include/robots/ev3/mindstorms_imu.h
```diff
@@ -0,0 +1,30 @@
...
+#include <utility>
+
+namespace BoBRobotics {
+class MindstormsIMU {
+    using degree_t = units::angle::degree_t;
+    using degrees_per_second_t = units::angular_velocity::degrees_per_second_t;
+
+public:
+    MindstormsIMU(const ev3dev::address_type &address = ev3dev::INPUT_AUTO);
+    degree_t getYaw();
```
neworderofjamie: Should these be const? I can imagine that could involve making ``m_IMU`` mutable depending on the const-correctness of EV3Dev

include/robots/tank.h
```diff
@@ -70,6 +70,10 @@ class Tank
 
     void stopReadingFromNetwork();
 
+    float getLeft() const;
```
neworderofjamie: Haha I wonder how many branches this change has been made in independently

projects/stone_cx/robotCommon.cc
```diff
@@ -2,22 +2,22 @@
...
+#include <algorithm>
+#include <numeric>
+
+float BoBRobotics::StoneCX::driveMotorFromCPU1(BoBRobotics::Robots::Tank &motor, bool display)
 {
     // Sum left and right motor activity
-    const float leftMotor = std::accumulate(&rCPU1[0], &rCPU1[8], 0.0f);
-    const float rightMotor = std::accumulate(&rCPU1[8], &rCPU1[16], 0.0f);
+    const float leftMotor = std::accumulate(&rCPU1[0], &rCPU1[4], 0.0f);
+    const float rightMotor = std::accumulate(&rCPU1[4], &rCPU1[8], 0.0f);
```
neworderofjamie: I'm **pretty** sure this is wrong

tools/tank_control/install_systemd_service.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
neworderofjamie: Is this for running on EV3Dev or something?

include/os/screen.h
alexdewar: Because we don't use OpenCV elsewhere in the OS module, so this way we can drop the dependency. I did this initially to compile on the EV3 which doesn't have it installed.

include/robots/ev3/mindstorms_imu.h
alexdewar: Good idea

include/os/screen.h
neworderofjamie: :+1:

tools/tank_control/install_systemd_service.sh
alexdewar: Yeah, or the Norbot. It just runs the server on boot and restarts it when it dies, so we don't have to ssh in.

include/robots/tank.h
alexdewar: I'm sure we'll get some lovely merge conflicts when all those branches eventually come back together!

PR #129, opened 2019-06-06
Using LM9DS1 to stabilise unwrapping

neworderofjamie:
I was thinking that if we're going to use the robots on a slope, some sort of stabilisation is going to really help improve the in-silica rotation. This is my, somewhat hacky, attempt at doing it using existing code. However, **ideally** we would do the equivalent of:

1. Call ``unwrapMaps`` once with ``m_CentrePixel`` = (0, 0)
2. Copy ``m_UnwrapMapX`` and ``m_UnwrapMapY`` somewhere.
3. When we change ``m_CentrePixel``, ``cv::add`` the offset to our copy of ``m_UnwrapMapX`` and ``m_UnwrapMapY``

Rather than calling ``unwrapMaps`` every time we update the centre pixel. Do you have any thoughts on how this could be integrated with the OpenCV unwrapper?  I think the ideal API would be:
```c++
void unwrap(const cv::Mat &input, cv::Mat &output, degree_t roll, degree_t pitch)
```

### DIFF
--- a/examples/camera_360_stable/Makefile
+++ b/examples/camera_360_stable/Makefile
@@ -1,0 +1,2 @@
+WITH_I2C := 1
+include ../../make_common/default.mk
--- a/examples/camera_360_stable/camera_360_stable.cc
+++ b/examples/camera_360_stable/camera_360_stable.cc
@@ -1,0 +1,111 @@
+// BoB robotics includes
+#include "common/logging.h"
+#include "common/lm9ds1_imu.h"
+#include "common/timer.h"
+#include "imgproc/opencv_unwrap_360.h"
+#include "video/panoramic.h"
+
+// Standard C++ includes
+#include <iostream>
+
+// Standard C includes
+#include <cmath>
+
+using namespace BoBRobotics;
+using namespace BoBRobotics::ImgProc;
+using namespace BoBRobotics::Video;
+
+int main()
+{
+    const cv::Size unwrapRes(720, 150);
+
+    // Create panoramic camera and suitable unwrapper
+    auto cam = getPanoramicCamera();
+    auto unwrapper = cam->createUnwrapper(unwrapRes);
+    const auto cameraRes = cam->getOutputSize();
+
+    // The cirlce we're unwrapping within represents a 
+    // half sphere, therefore it's radius is the arc length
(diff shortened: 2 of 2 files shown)


alexdewar:
What sort of frame rate do you get for this? I'd imagine that it'd be fairly slow be cool because of the frequent calls to ``updateMaps()``. If it works though I am all in favour of quick hacky solutions to problems 😉

I think probably the most efficient way to do this is to generate a massive lookup table with all of the likely possible combinations of pitch and roll (we should probably cache it on disk so we only need to do it once). Alex Cope did this for his camera code running on a Pi Zero and it worked quite nicely although I seem to remember that the lookup table ended up being at several hundred MB in size. A Jetson could definitely handle this though.

alexdewar:
I think you're right that that's how the API should look btw.

I also think we should maybe change the API for the IMU to use the unit library while we're at it.

neworderofjamie:
15ish doing nothing else on the Jetson so really not too slow. I don't really think a LUT is a sensible solution for this tbh - I would imagine that adding a scalar to an array of a few thousand floats (i.e. small enough to fit in L2 cache) is gonna be faster.

neworderofjamie:
Is my maths right that that is all that's required?

ajc158:
Unfortunately nope - that maths doesn't work! You need to do a whole
rotation matrix on each point then reproject onto the source image - this
is very computationally expensive but I'm working on a faster solution
using quaternions and hacks

neworderofjamie:
Ooh hello! Have you seen our unwrapping code (https://github.com/BrainsOnBoard/bob_robotics/blob/master/imgproc/opencv_unwrap_360.h#L158)?Where I'm talking about adding the position is to the polar coordinate map that go into the ``cv::remap`` function. Isn't that the same?

Multiplying quaternions by vectors **generally** requires more operations that multiplying matrices by vectors btw...

ajc158:
Doing this on my phone so might get it wrong, but the issue would be the
remapping to sphere section, which using sin and cos is most of the
computation and breaks the simple addition

ajc158:
Also, yep - I'm aware of the extra ops if you don't compose a matrix - but
for simplicity sake I'm doing the simple version first as I'm likely memory
bound on the QPUs still!

PR #130, opened 2019-06-07
Gazebo omniwheel robot simulation

alexdewar:
Hi both,

I found a model online of a Gazebo omniwheeled robot: https://github.com/GuiRitter/OpenBase

What do you think about trying to adapt it for the Mecanum robots that we'll be using for the hackathon?

The example code I've written (based on @shinjanxp's good work!) is fairly basic at the moment (e.g. you can't control the rollers). If we decide not to use it I'll probably just keep it as another Gazebo example, once I sort out the current multi-branch git madness we have.


examples/gazebo_omniwheel/omniwheel_plugin.cc
```diff
@@ -0,0 +1,96 @@
...
+#include <gazebo/gazebo.hh>
+#include <gazebo/msgs/msgs.hh>
+#include <gazebo/physics/physics.hh>
+#include <gazebo/transport/transport.hh>
+
+using namespace BoBRobotics;
+using namespace units::velocity;
+
+namespace gazebo {
+/// \brief A plugin to control a simple cart with differential drive.
```
neworderofjamie: This comment might need updating

include/robots/gazebo/omni2d.h
```diff
@@ -0,0 +1,37 @@
...
+// Gazebo
+#include <gazebo/msgs/msgs.hh>
+#include <gazebo/transport/transport.hh>
+
+namespace BoBRobotics {
+namespace Robots {
+namespace Gazebo {
+using namespace units::literals;
+
+class Omni2D
```
neworderofjamie: Might it be nice to add a base class for omniwheels as (in various PRs) we now have 2?

PR #131, opened 2019-06-11
Add support for Gazebo plugins to CMake build system

alexdewar:
This PR is currently just a placeholder so I can stick it on my todo list, as it depends on a few other branches that I should probably merge to ``cmake`` first.


alexdewar:
Hey @neworderofjamie. Welcome back! Hope you had a nice trip.

I think this branch is probably ready to be looked at, if you've got time. Would be handy to get it merged before the Hackathon.

.gitignore
```diff
@@ -26,6 +26,7 @@ bob_robotics.includes
 
 # Ignore object files
 **/*.a
+**/*.so
```
neworderofjamie: For symmetry should we also be ignoring .dll?

examples/gazebo_tank/differential_drive_plugin.cc
```diff
@@ -0,0 +1,110 @@
+#include <gazebo/gazebo.hh>
+#include <gazebo/physics/physics.hh>
+#include <gazebo/transport/transport.hh>
+#include <gazebo/msgs/msgs.hh>
+#include <gazebo/common/common.hh>
+
+namespace gazebo
+{
+  /// \brief A plugin to control a simple cart with differential drive.
```
neworderofjamie: really minor thing - I don't think we generally indent within namespaces (otherwise you end up indented way across the screen before you even start!). Also this looks like it's indented with 2 spaces rather than 4

include/common/gazebo_node.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+// Gazebo includes
+#include <gazebo/transport/transport.hh>
+#include <gazebo/gazebo_client.hh>
+
+using namespace gazebo::transport;
```
neworderofjamie: I don't think this belongs in common and it definitely should be in a namespace - personally I'd say ``BoBRobotics::Gazebo::getNode`` and ``BoBRobotics::Gazebo::shutdown`` would be better names

include/robots/gazebo_tank.h
```diff
@@ -0,0 +1,59 @@
...
+  : public Tank
+{
+    using meters_per_second_t = units::velocity::meters_per_second_t;
+    using radians_per_second_t = units::angular_velocity::radians_per_second_t;
+public:
+    GazeboTank(const radians_per_second_t maximumSpeed, gazebo::transport::NodePtr node)
+      : m_MaximumSpeed(maximumSpeed)
+    {
+        // Publish to the  differential_drive_robot topic
+        pub =node->Advertise<gazebo::msgs::Vector3d>("~/differential_drive_robot/vel_cmd");;
```
neworderofjamie: can't we use a 2d vector here?

.gitignore
alexdewar: I *think* we already are

examples/gazebo_tank/differential_drive_plugin.cc
alexdewar: 👍

include/common/gazebo_node.h
alexdewar: :+1:

include/robots/gazebo_tank.h
alexdewar: Good point.

alexdewar:
Hmm Jenkins seems not be running for some reason... Any thoughts, @neworderofjamie?

PS -- I think at some point you gave me the Jenkins login details but I can't seem to find them in my emalis anywhere, would you mind resending? That way I won't have to pester you so much with Jenkins issues :wink:

neworderofjamie:
Hmm can you even access https://gen-ci.inf.sussex.ac.uk? I'm getting a bad gateway

alexdewar:
Nope :disappointed:

include/robots/gazebo_tank.h
```diff
@@ -44,15 +45,17 @@ class GazeboTank
     {
         BOB_ASSERT(left >= -1.f && left <= 1.f);
         BOB_ASSERT(right >= -1.f && right <= 1.f);
-          // Set the velocity in the x-component
-        gazebo::msgs::Set(&msg, ignition::math::Vector3d(left * m_MaximumSpeed.value(), right * m_MaximumSpeed.value(), 0));
-        // Send the message
+        
+	// Set the velocity in the x-component
+        gazebo::msgs::Set(&msg, ignition::math::Vector3d(left * m_MaximumSpeed.value(), right * m_MaximumSpeed.value()));
```
neworderofjamie: I have no idea how gazebo works but why are we now setting a ``Vector2d`` message as ``Vector3d``?

alexdewar: Ah yes, I missed that one and then didn't build test it :wink:

PR #132, opened 2019-06-11
GPS stuff for Norbert

alexdewar:
Hey Norbert,

I cherry-picked a couple of small commits off another branch which I thought might be useful to you for the GPS stuff.

Commit summary:
- I made ``GPSCoordinates`` an alias for ``LatLon<WGS84>`` because I thought that would be clearer
- I changed the members of ``LatLon`` non-``const`` (e.g. so that you can assign to a ``LatLon``)

It's using your branch as a base, so just click merge if you're happy with it.

### DIFF
--- a/common/map_coordinate.h
+++ b/common/map_coordinate.h
@@ -84,6 +84,9 @@ struct LatLon
     const units::angle::degree_t lon;
 };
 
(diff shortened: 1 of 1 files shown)


tenxlenx:
thanks Alex

PR #139, opened 2019-07-26
Antworld rendering improvements

neworderofjamie:
This P.R. has yet to reach the point the branch name suggested but is a small refactor of some of the ant world rendering code:
1. ``RenderMesh`` previously duplicated a lot of code in ``World::Surface``
2.  Moved ``World::Surface`` and ``World::Texture`` out into top-level classes
3.  Made ``RenderMesh`` a base class using ``World::Surface`` for it's rendering and implemented current render mesh in ``RenderMeshCylinder``
4. Made ``World::Surface`` rather more flexible:
    - To enable indexed rendering (meaning that vertices can be shared between multiple triangles)
    - Used type-traits to implement support for different data formats


neworderofjamie:
Arg, I have no idea what's wrong with ``third_party/ev3dev-lang-cpp`` on my machine - every time I checkout  a branch it complains

include/antworld/surface.h
```diff
@@ -0,0 +1,140 @@
...
+
+        // Calculate number of vertices from positions
+        m_NumVertices = positions.size() / size;
+
+        // Unbind buffer
+        glBindBuffer(GL_ARRAY_BUFFER, 0);
+    }
+
+    template<typename T>
+    void uploadColours(const std::vector<T> &colours, GLint size = 3)
```
alexdewar: If ``size`` can only be e.g. 1 and 3 we could enforce it with a ``BOB_ASSERT``

include/antworld/surface.h
```diff
@@ -0,0 +1,140 @@
...
+
+        // Cache number of indices
+        m_NumIndices = indices.size();
+
+        // **NOTE** GL_ELEMENT_ARRAY_BUFFER works subtly different from GL_ARRAY_BUFFER
+        // as it has no client state/pointer tying it to the VAO. Therefore we need to
+        // leave it bound until after we unbind the VAO - makes sense but ugly
+    }
+    
+    void setTexture(const Texture *texture){ m_Texture = texture; }
```
alexdewar: Could ``texture`` be ``nullptr``? If not, maybe we should pass by reference or check with ``BOB_ASSERT``

src/antworld/render_mesh.cc
```diff
@@ -15,22 +15,30 @@ namespace BoBRobotics
...
+    // Render surface
+    m_Surface.render(GL_QUADS);
+
+    // Unbind surface, leaving texture binding alone
+    m_Surface.unbind(false);
 }
+
+//----------------------------------------------------------------------------
+// BoBRobotics::AntWorld::RenderMeshCylinder
 //----------------------------------------------------------------------------
```
alexdewar: Maybe a one-line comment would be good here just to explain what ``RenderMeshCylinder`` is and how it differs from ``RenderMesh``.

I'm well aware that being a bit of a hypocrite here :wink:

src/antworld/surface.cc
```diff
@@ -0,0 +1,90 @@
...
+{
+    // If surface has a texture, bind it
+    if(unbindTexture && m_Texture != nullptr) {
+        glDisable(GL_TEXTURE_2D);
+        m_Texture->unbind();
+    }
+
+    // Unbind vertex array
+    glBindVertexArray(0);
+}
```
alexdewar: It might be clearer in calling code if these were split into separate functions instead, e.g. ``bind()`` and ``bindWithTexture()``, but up to you.

examples/ant_world/ant_world.cc
```diff
@@ -112,7 +112,7 @@ int main(int argc, char **argv)
                                     4096,
                                     GL_COMPRESSED_RGB);
     } else {
-        renderer.getWorld().load(filesystem::path(argv[0]) / "../../resources/antworld/world5000_gray.bin",
+        renderer.getWorld().load(filesystem::path(argv[0]).parent_path() / "../../resources/antworld/world5000_gray.bin",
```
alexdewar: Oops! I think this was me.

include/antworld/surface.h
neworderofjamie: Yeah, that just signifies no texture

include/antworld/surface.h
neworderofjamie: OpenGL claims that it will produce a ``GL_INVALID_VALUE`` if these are invalid - our code is generally attaching an OpenGL error callback so I think this is probably ok

src/antworld/surface.cc
neworderofjamie: agreed!

src/antworld/render_mesh.cc
neworderofjamie: So, good thing you raised this, turns out it's not a cylinder in any way so renamed the class and added a brief explanation of what it does

src/antworld/surface.cc
neworderofjamie: :+1:

include/antworld/surface.h
alexdewar: Ok cool, as long as there's some error checking going on

include/antworld/surface.h
alexdewar: Yeah ok. A pointer seems right then.

PR #141, opened 2019-08-02
Add drone flight controller in Gazebo

shinjanxp:


examples/gazebo_iris/gazebo_quadcopter_plugin.cc
```diff
@@ -0,0 +1,391 @@
...
+    this->m_LastControllerUpdateTime = curTime;
+}
+/////////////////////////////////////////////////
+void
+GazeboQuadCopterPlugin::ResetPIDs()
+{
+    // Reset velocity PID for rotors
+    for (size_t i = 0; i < this->m_Rotors.size(); ++i) {
+        this->m_Rotors[i].m_Cmd = 0;
+    }
```
alexdewar: Another way to do this would be:
```c++
std::fill(m_Rotors.begin(), m_Rotors.end(), 0);
```

If you haven't come across the STL algorithms before, this is a good talk about them: https://www.youtube.com/watch?time_continue=1&v=2olsGf6JIkU (but don't feel obligated to watch it!)
They have a weird syntax but are really useful once you get past that.

examples/gazebo_iris/gazebo_quadcopter_plugin.cc
```diff
@@ -0,0 +1,391 @@
...
+{
+    MotorMixing(_dt);
+    // update velocity PID for rotors and apply force to joint
+    for (size_t i = 0; i < this->m_Rotors.size(); ++i) {
+        double velTarget = this->m_Rotors[i].m_Multiplier *
+                           this->m_Rotors[i].m_Cmd /
+                           this->m_Rotors[i].m_RotorVelocitySlowdownSim;
+        double vel = this->m_Rotors[i].m_Joint->GetVelocity(0);
+        double error = vel - velTarget;
+        double force = this->m_Rotors[i].m_Pid.Update(error, _dt);
```
alexdewar: These could all be ``const``.

examples/gazebo_iris/gazebo_quadcopter_plugin.hh
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef GAZEBO_PLUGINS_GazeboQuadCopterPlugin_HH_
+#define GAZEBO_PLUGINS_GazeboQuadCopterPlugin_HH_
```
alexdewar: I'd use a ``#pragma once`` instead here.

examples/gazebo_iris/gazebo_quadcopter_plugin.hh
```diff
@@ -0,0 +1,153 @@
...
+namespace gazebo {
+
+class GAZEBO_VISIBLE GazeboQuadCopterPlugin : public ModelPlugin
+{
+public:
+    /// \brief Constructor.
+    GazeboQuadCopterPlugin();
+
+    /// \brief Destructor.
+    ~GazeboQuadCopterPlugin();
```
alexdewar: As I think I said above, you can just omit the ctor and dtor if they don't do anything.

examples/gazebo_iris/bob_iris.cc
```diff
@@ -0,0 +1,69 @@
...
+{
+    /************************************Gazebo setup************/
+
+    // Create our node for publishing joystick values
+    gazebo::transport::NodePtr node = getGazeboNode();
+
+    /************************************Gazebo setup end************/
+    std::unique_ptr<Display> display;
+    std::unique_ptr<GazeboCameraInput> cam;
+    if(argc >= 3) { // Initialize gazebo camera if more than 2 arguements are provided (display switch and camera url)
```
alexdewar: What I'd probably do here is a switch block: if ``argc`` is 1, then there's no camera, if it's 3 then you can parse the args, otherwise you can print usage information

examples/gazebo_iris/iris_with_standoffs_demo/model.config
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0"?>
```
alexdewar: I guess we could drop this folder?

src/robots/gazebo_uav.cc
```diff
@@ -0,0 +1,65 @@
...
+namespace Robots {
+
+void GazeboQuadCopter::takeOff()
+{
+    m_Armed=true;
+    m_Thrust = 1;
+    sendCommand();
+    std::this_thread::sleep_for(500ms);
+    m_Thrust = 0.5;
+    sendCommand();
```
alexdewar: I'm worried there might be some race conditions going on here. If someone presses the button for take-off and then presses some other buttons at the same time then things could be in an inconsistent state. You can sort this with a mutex.

alexdewar:
@neworderofjamie I think the CI checks are failing because of some Gazebo issue on your machine. Did you uninstall it or something?

examples/gazebo_iris/gazebo_quadcopter_plugin.cc
```diff
@@ -0,0 +1,391 @@
...
+            rotor.m_Pid.SetIMin(param);
+
+            getSdfParam<double>(rotorSDF, "vel_cmd_max", param, rotor.m_Pid.GetCmdMax());
+            rotor.m_Pid.SetCmdMax(param);
+
+            getSdfParam<double>(rotorSDF, "vel_cmd_min", param, rotor.m_Pid.GetCmdMin());
+            rotor.m_Pid.SetCmdMin(param);
+
+            // set pid initial command
+            rotor.m_Pid.SetCmd(0);
```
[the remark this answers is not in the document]
shinjanxp: This just sets the initial PID error value to zero.

examples/gazebo_iris/gazebo_quadcopter_plugin.cc
```diff
@@ -0,0 +1,391 @@
...
+
+    if (!this->m_ImuSensor) {
+        gzerr << "imu_sensor [" << imuScopedName
+              << "] not found, abort GazeboQuadCopter plugin.\n"
+              << "\n";
+        return;
+    }
+
+    // Controller time control.
+    this->m_LastControllerUpdateTime = 0;
```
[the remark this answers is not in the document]
shinjanxp: Right. Overlooked `this` mess.

examples/gazebo_iris/gazebo_quadcopter_plugin.cc
```diff
@@ -0,0 +1,391 @@
...
+/////////////////////////////////////////////////
+void GazeboQuadCopterPlugin::OnUpdate()
+{
+  std::lock_guard<std::mutex> lock(this->m_Mutex);
+
+  gazebo::common::Time curTime = this->m_Model->GetWorld()->SimTime();
+
+    this->ApplyMotorForces((curTime -this->m_LastControllerUpdateTime).Double());  
+    this->m_LastControllerUpdateTime = curTime;
+}
```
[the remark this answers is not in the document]
shinjanxp: Thanks for the tip. I was actually a little lazy to do the formatting myself and the default C/C++ formatter didn't do anything.

examples/gazebo_iris/gazebo_quadcopter_plugin.cc
shinjanxp: Ok, but won't this set all the members to 0? I mean we only need to reset the `m_Cmd` member. I don't know how `std::fill` operates on objects.

examples/gazebo_iris/iris_with_standoffs_demo/model.config
shinjanxp: No. Both the folders `iris_with_standoffs` and `iris_with_standoffs_demo` are being used. I could try to consolidate them into one model directory.

examples/gazebo_iris/gazebo_quadcopter_plugin.cc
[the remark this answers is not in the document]
alexdewar: Ohh I see. Maybe you could tweak the comment to say that? The word command seems a bit confusing

examples/gazebo_iris/iris_with_standoffs_demo/model.config
alexdewar: Might make sense, unless they're really for different things. There is also a stray .tar.gz file in there.

shinjanxp: Removed the tar.gz. Will shift them into one folder and commit

examples/gazebo_iris/gazebo_quadcopter_plugin.cc
alexdewar: Oops, I misread it. Yes you're totally right, it wouldn't work. My bad.

examples/gazebo_iris/iris_with_standoffs_demo/model.config
alexdewar: :+1:

alexdewar:
@neworderofjamie ping?

I've also noticed that we now seem to be getting new ``-Wdeprecated-copy`` warnings, but they seem to do with code inside the Eigen headers, so I'm thinking we should just suppress them.

alexdewar:
Hi @shinjanxp. Are you happy with the code now?

alexdewar:
Hmm, it's looking like gcc can't find the same Gazebo header on my machine either... but it worked fine when I logged into my work machine and tried to build it, so it seems like it's only failing when Jenkins tries to build it. Any thoughts, @neworderofjamie ?

@shinjanxp I've just pushed a few commits to try to fix some of the CI warnings.

neworderofjamie:
Apologies, I think I may have muted this thread without reading why I was getting notified :blush: 

Not sure what's wrong with gazebo on my machine, libgazebo9-dev is installed and ``transport.hh`` is sitting in ``/usr/include/gazebo-9/gazebo/transport``

neworderofjamie:
How does cmake find gazebo? If it requires an environment variable, the only reliable solution is to set it via Jenkins (which I can sort out for you)

examples/gazebo_iris/gazebo_quadcopter_plugin.cc
```diff
@@ -0,0 +1,391 @@
...
+    this->m_YawPID.SetCmdMin(param);
+
+    this->m_RollPID.SetCmd(0);
+    this->m_PitchPID.SetCmd(0);
+    this->m_YawPID.SetCmd(0);
+    
+    char* logfile_location=std::getenv("LOG_FILE");
+    if(logfile_location!=NULL){
+        this->m_Logfile.open(logfile_location, std::ios_base::app);
+    }
```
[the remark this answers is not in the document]
shinjanxp: I'm having an issue with the `BOB_ASSERT` statement. I included `common/macros.h` but am getting the following error
```
/usr/bin/ld: BoB/modules/common/./libbob_common.a(macros.cc.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
BoB/modules/common/./libbob_common.a: error adding symbols: Bad value
```
This is a gazebo plugin, so I don't think cmake is being able to include BoB robotics headers.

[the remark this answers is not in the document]
alexdewar: Ah yes. I think there's a ``GZ_ASSERT`` that you could use though?

alexdewar:
@neworderofjamie Gazebo usually tries a bunch of different tricks to find libraries, but you can usually also explicitly set the location with an env var, although unfortunately not in Gazebo's case.

I made another branch to debug this. I've printed out what CMake thinks the necessary include paths and libraries are and it looks totally normal:
```
gazebo include dirs: /usr/include/gazebo-9;/usr/include/sdformat-6.0;/usr/include/ignition/transport4;/usr/include/ignition/msgs1;/usr/include/ignition/fuel_tools1;/usr/include/x86_64-linux-gnu;/usr/lib/x86_64-linux-gnu/libzip/include;/usr/include/ignition/common1;/usr/include/ignition/math4;/usr/include/uuid;/usr/include/glib-2.0;/usr/lib/x86_64-linux-gnu/glib-2.0/include;/usr/include/x86_64-linux-gnu
gazebo library flags: gazebo_transport;gazebo_physics;gazebo_sensors;gazebo_rendering;gazebo_gui;gazebo_client;gazebo_msgs;gazebo_common;gazebo;boost_thread;boost_signals;boost_system;boost_filesystem;boost_program_options;boost_regex;boost_iostreams;boost_date_time;boost_chrono;boost_atomic;pthread;sdformat;tbb;atomic;ignition-transport4;uuid;ignition-msgs1;protobuf;pthread;ignition-math4;ignition-fuel_tools1
```

This is really weird.

src/robots/gazebo_uav.cc
shinjanxp: Put in a mutex on the takeoff and land functions. There's also an atomic_bool variable `is_Armed` that is checked in all the flight control functions

shinjanxp:
I think all the revisions are done from my end.

examples/gazebo_iris/bob_iris.cc
```diff
@@ -0,0 +1,75 @@
...
+            if(strcmp(argv[1], "-p") == 0) {
+                std::cout << "Using panoramic camera.\n";
+                cam = std::make_unique<GazeboCameraInput>(node, argv[2], true);
+                display = std::make_unique<Display>(*cam, cv::Size(640,320)); //unwrap resolution needs to be supplied
+            }
+            // else if(strcmp(argv[1], "-s") == 0){
+            //     std::cout << "Using simple camera.\n";
+            //     cam = std::make_unique<GazeboCameraInput>(node, argv[2], false);
+            //     display = std::make_unique<Display>(*cam); //unwrap resolution needs to be supplied
+            // }
```
alexdewar: I guess we don't need this?

src/robots/gazebo_uav.cc
alexdewar: Looks good to me.

examples/gazebo_iris/bob_iris.cc
shinjanxp: Not really. I kept it in case we need to put a simple camera on the drone. Can be removed if there are no such plans.

alexdewar: Is there already a simple camera on the drone, i.e. one that came with the model? If so, then you could just uncomment this. Otherwise I'd just remove it.

shinjanxp: No there isn't a simple camera model on it. Will remove it then

alexdewar:
@shinjanxp I figured out why the build was breaking. ``gazebo_uav.cc`` was being built along with all the other bits in the ``robots`` module, even for examples which weren't using Gazebo, hence why it couldn't find the headers. So I moved it into the ``robots/gazebo`` module and also put the camera code and ``node.h`` in there while I was at it, so all the Gazebo stuff is in one place.

alexdewar:
I also renamed a couple of the gazebo examples for consistency.

PR #142, opened 2019-08-08
Keyboard joystick

neworderofjamie:
I have been meaning to do this for a long time (and I know it annoys @tenxlenx regularly) but, sitting here without a joystick was the final straw. This is a GLFW keyboard input based implementation of ``HID::JoystickBase``. I added a wrapper function to ant world that I feel might belong somewhere in ``HID`` but any thoughts would be much appreciated.

### DIFF
--- a/examples/ant_world/ant_world.cc
+++ b/examples/ant_world/ant_world.cc
@@ -50,8 +50,9 @@ std::unique_ptr<HID::JoystickBase<HID::JAxis, HID::JButton>> createJoystick(GLFW
     {
         return std::make_unique<HID::Joystick>(0.25f);
     }
-    catch(std::runtime_error &)
+    catch(std::runtime_error &ex)
     {
+        LOGW << "Error opening joystick - \"" << ex.what() << "\" - using keyboard interface";
         return std::make_unique<HID::JoystickGLFWKeyboard>(window);
     }
 }
--- a/examples/ant_world/ant_world.cc
+++ b/examples/ant_world/ant_world.cc
@@ -1,6 +1,7 @@
 // BoB robotics includes
 #include "common/logging.h"
 #include "hid/joystick.h"
+#include "hid/joystick_glfw_keyboard.h"
 #include "antworld/common.h"
 #include "antworld/renderer.h"
 
@@ -16,6 +17,7 @@
 
 // Standard C++ includes
 #include <cstring>
+#include <memory>
 
 using namespace BoBRobotics;
 using namespace units::angle;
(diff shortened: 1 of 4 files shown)


alexdewar:
This is a really good idea. Lots of our code is def too joystick dependent.

I've merged in the ``cmake`` branch to try to fix the failing CI checks.

examples/ant_world/ant_world.cc
```diff
@@ -41,6 +43,18 @@ inline second_t getCurrentTime()
...
+
+std::unique_ptr<HID::JoystickBase<HID::JAxis, HID::JButton>> createJoystick(GLFWwindow *window)
+{
+    try
+    {
+        return std::make_unique<HID::Joystick>(0.25f);
+    }
+    catch(std::runtime_error &)
+    {
+        return std::make_unique<HID::JoystickGLFWKeyboard>(window);
```
alexdewar: Might be worth printing a little message here, in case the user has plugged in a joystick which just isn't working.

src/hid/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.1)
 include(../../cmake/bob_robotics.cmake)
-BoB_module(SOURCES joystick_linux.cc joystick_windows.cc
+BoB_module(SOURCES joystick_linux.cc joystick_windows.cc joystick_glfw_keyboard.cc
            BOB_MODULES common
            PLATFORMS linux windows)
```
alexdewar: We could drop the ``PLATFORMS`` array now, because ``JoystickGLFWKeyboard`` should also work on Macs (I guess?)

src/hid/joystick_glfw_keyboard.cc
```diff
@@ -0,0 +1,60 @@
+#include "hid/joystick_glfw_keyboard.h"
+
+// GLFW
+#include <GLFW/glfw3.h>
+
+namespace BoBRobotics {
+namespace HID {
+
+//! Open default joystick device with (optionally) specified dead zone
```
alexdewar: This comment could be updated

src/hid/joystick_glfw_keyboard.cc
```diff
@@ -0,0 +1,60 @@
...
+
+    // Use page up and down for right trigger
+    setAxisStateKeys(GLFW_KEY_PAGE_UP, GLFW_KEY_PAGE_DOWN, JAxis::RightTrigger);
+
+    setButtonStateKey(GLFW_KEY_1, JButton::A);
+    setButtonStateKey(GLFW_KEY_2, JButton::B);
+    setButtonStateKey(GLFW_KEY_3, JButton::X);
+    setButtonStateKey(GLFW_KEY_4, JButton::Y);
+
+    return true;
```
alexdewar: Maybe in a future PR we could make this configurable, e.g. with a YAML file, because we're using some of the other buttons elsewhere and it might also be able to configure what axis value is used when a keyboard button is pressed

src/hid/CMakeLists.txt
neworderofjamie: :+1:

src/hid/joystick_glfw_keyboard.cc
neworderofjamie: Good idea - opened #143

PR #144, opened 2019-08-10
Fix and tidy up unwrapfile

alexdewar:
Hey,

I tried to use ``unwrapfile`` to unwrap a video recently and had a couple of issues:
* Seemingly my version of OpenCV doesn't like the output format we were using
* It would only extract the frames as JPEGs, because the default value of ``extractFrames`` was set to ``true``

I've fixed up these issues and added a couple of enhancements:
* Use CMake to find the path to ffmpeg instead of just guessing
* Use a third-party library to parse the command-line args more cleanly (https://github.com/brofield/simpleopt)


tools/unwrapfile/unwrapfile.cc
```diff
@@ -181,76 +176,111 @@ void unwrapMP4Frames(const char *filepathRaw, unsigned int frameInterval, const
...
 }   // Anonymous namespace
 
+void usage()
+{
+    std::cout << "usage: unwrapfile [--no-sound] [--extract-frames[=SKIP]] [--camera CAMERA_NAME] [--unwrapped-resolution WIDTHxHEIGHT] [file(s)]" << std::endl;
+}
+
 int main(int argc, char** argv)
 {
     if (argc == 1) {
```
neworderofjamie: For the CLI tools I made for the ALife paper we used a really nice CLI library (https://github.com/dannyneil/spiking_relu_conversion/blob/master/normalize_cnn_data.m#L8) which was much more modern c++ and terser - would you mind horrible switching to that (and adding it to third_party)

tools/unwrapfile/unwrapfile.cc
```diff
@@ -181,76 +176,111 @@ void unwrapMP4Frames(const char *filepathRaw, unsigned int frameInterval, const
...
+        { CAMERA, "--camera", SO_NONE },
+        { UNWRAPPED_RESOLUTION, "--unwrapped-resolution", SO_REQ_SEP },
+        { HELP, "-h", SO_NONE },
+        { HELP, "--help", SO_NONE },
+        SO_END_OF_OPTIONS
+    };
+
     bool copysound = true;
-    bool extractFrames = true;
+    bool extractFrames = false;
```
neworderofjamie: @tenxlenx and I encountered this bug as well - apologies for, presumably, fixing it on my machine and not committing...

tools/unwrapfile/unwrapfile.cc
alexdewar: Done. That is a much nicer library!

The only issue I had is that I couldn't find a way to get the remaining args, which in our case are filenames, so I had to add a ``-i`` option for that instead. Any ideas? There is an ``app.remaining()`` method, which seems to do that, but when I ran the program I was getting an "unrecognised argument" error.

neworderofjamie: I *think* you need to call ``app.allow_extras(true)`` first and *then* ``app.remaining()`` *should* work...

alexdewar: That seemed to do the trick. Thanks!

PR #145, opened 2019-08-13
Add script and C++ code for Giger bee-eye model

alexdewar:
Hey,

As we're using this code in a few places for different demos etc., I thought it would be good to tidy it up and merge it so that we just have one version of it. The other reason is that @shinjanxp was planning on using it for a demo video for his GSoC write-up.

Other small changes:
* Move various scripts into the ``bin/`` folder
* Add functions for getting the path to the BoB robotics repo -- if it's a submodule the ``BOB_ROBOTICS_SUBMODULE_PATH`` macro can be set and then the function will look there instead of using the env var. The current way we're doing it is a bit error prone.


src/imgproc/bee_eye.cc
```diff
@@ -0,0 +1,48 @@
...
+#include "imgproc/bee_eye.h"
+
+// Standard C includes
+#include <cmath>
+
+namespace BoBRobotics {
+namespace ImgProc {
+namespace BeeEye {
+
+Map::Map(const float eyeData[][4], const int eyeDataLength,
```
neworderofjamie: I think this copies the arrays by value which, especially with the eye data is probably not what you want. The syntax here is singularly grim and, I find, very hard to remember but I **think** this should be the correct syntax:
```c++
template <size_t eyeDataLength>
Map::Map(const float (&eyeData)[eyeDataLength][4], const int (&eyeSize)[2], const int (&imageSize)[2], bool useMask)
```
On the bright side, it removes the need for manually handling eyeDataLength but, on the less-bright side it would need to go in the header.

alexdewar:
Ohh crap... I'd forgotten about that.

@ajc158 Is it ok for us to make this public now? If not I can delete the branch.

examples/bee_eye/bee_eye.cc
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    BeeEye::Map eyeMap{ BeeEye::EyeData, BeeEye::EyeDataLength,
+                        BeeEye::EyeSize, BeeEye::ImageSize };
+    cv::Mat outputImage, inputImageRaw, inputImageProcessed;
+    while (true) {
+        try {
+            // Check for new frame
+            if (!video->readFrame(inputImageRaw)) {
+                std::this_thread::sleep_for(2ms);
```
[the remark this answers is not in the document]
alexdewar: Ah yeah, I always forget about vsync. Makes sense.

include/common/bob_robotics_path.h
```diff
@@ -0,0 +1,20 @@
...
+namespace BoBRobotics {
+/**!
+ * \brief Get the path to the BoB robotics repository
+ *
+ * If the current program was built using a git submodule, the path to that is
+ * used, otherwise it falls back on the value of the BOB_ROBOTICS_PATH
+ * environment variable. Throws std::runtime_error if the path cannot be found.
+ */
+filesystem::path
+getBoBRoboticsPath();
```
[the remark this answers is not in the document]
alexdewar: I see what you mean, but I think ``getPath()`` by itself might be a bit ambiguous without the ``BoBRobotics::`` bit in front of it. How about ``getRepoPath()`` or ``getCodePath()`` or something?

src/imgproc/bee_eye.cc
alexdewar: Oh yeah, of course. I forgot that fixed-size arrays get passed by value.

alexdewar: Hmm I tried that and then was getting an error saying I was trying to declare an array of references... So I just changed it to ``float (*eyeData)[4]`` instead and that seemed to work.

neworderofjamie: How very strange - http://cpp.sh/96pyq works fine for me

alexdewar: Ok, so I've got it to work now.

I'm not sure what I was doing wrong. It might have been that there was an earlier error that I missed that was causing this. It's still a bit strange.

neworderofjamie: Awesome! As I said, that is one of my least-favourite bits of C++  syntax, so weird-lookin'

alexdewar:
I think this is ready to merge now -- as long as @ajc158 is ok with this. Alex?

I reorganised the path code a bit: ``getBoBRoboticsPath()`` is now ``BoBRobotics::Path::getRepoPath()``

Issue #143, opened 2019-08-08
Make ``BoBRobotics::HID::JoystickGLFWKeyboard`` more configurable


neworderofjamie:
It would be nice if the mapping of GLFW keys to axis could be configurable - probably via a YAML file.

PR #146, opened 2019-08-13
Replace GLFW with SFML

alexdewar:
Hey,

I know we were talking about this somewhere else, but I can't remember where right now. This PR replaces all the uses of GLFW with their SFML equivalents and drops support for GLFW.

I've replaced the ``JoystickGLFWKeyboard`` class with ``JoystickSFMLKeyboard`` which *should* behave in the same way. I slightly changed the design because SFML handles events differently to OpenGL: You get an ``sf::Event`` struct and then check this to see if the event you're interested in has happened, so it meant that calling different functions for different button presses would be more verbose. So I used a series of ``std::map``s instead (which might also be useful for issue #143) -- it's a bit ugly atm so any suggestions welcome.


alexdewar:
I've left the SDL support for now because @tenxlenx is still using it.

neworderofjamie:
Oh shit! This is a thankless task, mostly involving my code....I was planning to do this honest....

examples/ant_world/ant_world.cc
```diff
@@ -68,41 +52,25 @@ int main(int argc, char **argv)
...
     // Initialize GLEW
     if(glewInit() != GLEW_OK) {
         LOGE << "Failed to initialize GLEW";
         return EXIT_FAILURE;
     }
 
-    // Enable VSync
-    glfwSwapInterval(1);
-
-    glDebugMessageCallback(handleGLError, nullptr);
```
neworderofjamie: Does SFML do this internally and convert GL errors to exceptions? If not, without this second call, you have to use OpenGL in gross "wrapping each function call in a macro and checking it's return" mode

projects/ardin_mb/ardin_mb.cc
```diff
@@ -32,109 +32,27 @@
...
-}   // anonymous namespace
-
-
-
 int main(int argc, char *argv[])
 {
-    // Set GLFW error callback
-    glfwSetErrorCallback(handleGLFWError);
+    sf::ContextSettings settings;
+    settings.antialiasingLevel = 8;
```
neworderofjamie: Is there any particular reason you're doing this?

src/hid/joystick_sfml_keyboard.cc
```diff
@@ -0,0 +1,97 @@
+// BoB robotics includes
+#include "hid/joystick_sfml_keyboard.h"
+
+template<class T, class Func>
+bool runIfFound(const std::map<sf::Keyboard::Key, T> m,
```
neworderofjamie: I think this should be a reference :smile:

src/hid/joystick_sfml_keyboard.cc
```diff
@@ -0,0 +1,97 @@
...
+//------------------------------------------------------------------------
+JoystickSFMLKeyboard::JoystickSFMLKeyboard(sf::Window &window)
+  : JoystickBase(0.0f), m_Window(window)
+{}
+
+//------------------------------------------------------------------------
+bool JoystickSFMLKeyboard::updateState()
+{
+    sf::Event event;
+    if (!m_Window.pollEvent(event)) {
```
neworderofjamie: Hmm, I think this is problematic, it looks like (I haven't actually used SFML before in anger) this function is supposed to be called in your main loop so:

1. Multiple events will need to be processed each update
2. They may be required by some other sub-system

I think what you need is to use ``sf::Keyboard::isKeyPressed``. So many something like:

```c++
for(auto key : ButtonKeys) {
    // Determines whether key is now down (I THINK this must what this API actually returns)
    const bool nowDown = sf::Keyboard::isKeyPressed(key.first);
    
    // If state has changed
    if(previouslyDown != isDown(key.second)) {
        if(previouslyDown) {
           setReleased(key.second, false);
        }
        else {
            setPressed(key.second, false);
        }
    }
}
```
You could also then make it ``ButtonKeys`` a fixed size array ``JButton::Max`` although, this always end  up being ugly with class enums (which is one of the things that really irritates me with C++11)

alexdewar:
Ah that's ok. I started doing this a while ago so it just needed finishing up.

examples/ant_world/ant_world.cc
alexdewar: I assumed it did, but it seems not: https://www.sfml-dev.org/tutorials/2.5/window-opengl.php

I'll reinstate this.

projects/ardin_mb/ardin_mb.cc
alexdewar: I just copied and pasted it from somewhere where I was using anti-aliasing. Would you rather it was turned off?

src/hid/joystick_sfml_keyboard.cc
alexdewar: D'oh! It seems I've forgotten about references entirely recently :worried:

projects/ardin_mb/ardin_mb.cc
neworderofjamie: As these examples are using OpenGL directly rather than SFML, I just suspect it doesn't do anything :)

src/hid/joystick_sfml_keyboard.cc
alexdewar: That is a good point. I did think about this: for example, our ``Viz::SFMLWorld`` class won't work in conjunction with this code because it also calls ``pollEvents()``. Do you think there's a chance it might miss keyboard events this way though?

projects/ardin_mb/ardin_mb.cc
alexdewar: Ah fair enough. I'll drop it then.

src/hid/joystick_sfml_keyboard.cc
neworderofjamie: Yeah, as far as I can tell, ``pollEvents`` should only be called in the outer loop of the application and events dispatched from there - if you've ever had the misfortune of ye olde win 32 API programming it works in a similar way.

Potentially but, assuming we don't lazily do loads of stuff in the main thread (which, admittedly, I often do) it would require users doing some serious sub-ms key-bashing :smile:

alexdewar: Good point. I guess if we're doing too much on the main thread then that's an issue by itself. Could always run ``joystick.runInBackground()`` as a hack in that case.

neworderofjamie:
The only fly in this ointment is #147

alexdewar:
Hmm that is a drag. In the meantime I'll just add CMake support for GLFW back in so we can carry on using imgui.

neworderofjamie:
Can you add deprecation warnings to the cmake magic for SDL and GLFW to encourage @tenxlenx and I to fix our shit up?

alexdewar:
Sure.

@tenxlenx If you're still having issues installing SFML, you could try installing it using vcpkg: https://github.com/Microsoft/vcpkg

alexdewar:
Ok, done. Are you happy to merge?

neworderofjamie:
How about the keyboard stuff?

alexdewar:
Oops forgot about that!

⁣Sent from BlueMail ​

alexdewar:
Ok done. The tests have only been run on my machine though, presumably because yours is asleep?

I dropped the ``std::map``s in the end in favour of arrays of structs.

PR #149, opened 2019-08-14
8-bit vertex colours in LibAntWorld

### LINKED ISSUE #148: 8-bit vertex colours in LibAntWorld
Currently the obj loader loads colours into a 32-bit floating point vertex buffer - this is overkill and will really hit the fillrate on low-end cards. Post #139 this should be trivial to implement

neworderofjamie:
The obj loader loader was uploading colours to the GPU with 32-bit floating point per channel - this is overkill and really hits the fill-rate on low-end cards (results in doubling of frame rate on my not-that low-end card :smile: )
Fixed #148

### DIFF
--- a/src/antworld/world.cc
+++ b/src/antworld/world.cc
@@ -230,6 +230,11 @@ void World::loadObj(const filesystem::path &filename, float scale, int maxTextur
     LOG_DEBUG << "Max texture size: " << maxTextureSize;
 
     // Vector of geometry associated with each named surface (in unindexed triangle format)
+    // Each surface consists of:
+    // 0 - a name
+    // 1 - a vector of floating point vertex positions (x, y, z)
+    // 2 - a vector of 8-bit colours (r, g, b)
+    // 3 - a vector of floating point texture coordinates (U, V)
     std::vector<std::tuple<std::string, std::vector<GLfloat>, std::vector<GLbyte>, std::vector<GLfloat>>> objSurfaces;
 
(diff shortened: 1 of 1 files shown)


src/antworld/world.cc
```diff
@@ -33,11 +33,22 @@ void readVector(std::istringstream &stream, std::vector<GLfloat> &vector, float
...
+    }
+}
+//----------------------------------------------------------------------------
 void readFace(std::istringstream &lineStream,
               const std::vector<GLfloat> &rawPositions,
-              const std::vector<GLfloat> &rawColours,
+              const std::vector<GLbyte> &rawColours,
               const std::vector<GLfloat> &rawTexCoords,
-              std::tuple<std::string, std::vector<GLfloat>, std::vector<GLfloat>, std::vector<GLfloat>> &currentObjSurface)
+              std::tuple<std::string, std::vector<GLfloat>, std::vector<GLbyte>, std::vector<GLfloat>> &currentObjSurface)
```
alexdewar: I'm guessing that the ``std::vector<GLbyte>`` here contains sequences of 3 bytes for each colour channel?

alexdewar: Also, maybe a comment is needed somewhere to describe what goes into these tuples.

neworderofjamie: Added if you wanna merge!

alexdewar: Cool

PR #150, opened 2019-08-15
Use SFML for ImGui

### LINKED ISSUE #147: Write ImGui interface for SFML
SFML isn't currently one of the supported window-creation frameworks for ImGUI - It looks pretty simple and we can probably even push it back upstream like nice open source people

neworderofjamie:
Hopefully this will allow us to move entirely to using SFML. 
1. New example, building IMGUI demo application using BoB robotics build system and SFML for window creation
2. Attempted to directly port GLFW interface for ImGui to SFML - a features aren't supported in SFML 2.4 which is provided by Ubuntu 18 but SFML 2.5 seems far superior - my hope is to push this upstream

@alexdewar I feel it would be better to have the OpenGL stuff as a dependency for the imgui module rather than the example, but I have no idea how to do that :smile: 
@alexdewar - ImGui doesn't build cleanly with your compiler, could you look at what flags are required, this is also beyond my current level of cmake skills
@alexdewar *finally* could you try the example on your machine and check the clipboard and cursor stuff actually works rather than merely compiling

Fixes #147


examples/imgui/imgui.cc
```diff
@@ -20,7 +20,7 @@
 
 int main(int, char**)
 {
-    sf::Window window(sf::VideoMode(1280, 720), "Dear ImGui GLFW+OpenGL2 example");
+    sf::Window window(sf::VideoMode(1280, 720), "Dear ImGui SFML+OpenGL2 example");
```
neworderofjamie: So we are :smile:

alexdewar:
Ok, I think I've fixed everything up now and it seems to be building without warnings on my machine. I've changed ``bob_robotics.cmake`` so that it automatically builds against glew and sfml-graphics.

I've also tested the example and everything seems to be working ok. So I think we're ready to merge!

PR #151, opened 2019-08-19
Add module for interfacing with Python code

alexdewar:
I was talking to @efkag and @FullAPVayne about using Ant World recently and I thought it might be worth adding some basic code to call Python functions from BoB robotics projects. (It might also be handy to have a mechanism to call C++ functions from Python -- I did start this but it was a bit of a faff and I got bored.)

It's based on this little library (from the same guy who wrote the matplotlibcpp code): https://github.com/lava/wrappy

I hacked it a bit to add a few features that we needed (notably python 3 support) and I've added a few functions and a ``NumpyArray`` class for passing numpy arrays to Python code (e.g. if you want to pass a ``cv::Mat`` without copying all the data into a Python array first).

I've added an example using the module with Ant World.

@efkag and @FullAPVayne: Does this look usable for you? Is there anything missing?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "third_party/plog"]
 	path = third_party/plog
 	url = ../../SergiusTheBest/plog.git
+[submodule "third_party/wrappy"]
+	path = third_party/wrappy
+	url = ../../alexdewar/wrappy.git
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -7,6 +7,7 @@
                 "${workspaceFolder}",
                 "${workspaceFolder}/third_party/ev3dev-lang-cpp",
                 "${workspaceFolder}/third_party/plog/include",
+                "${workspaceFolder}/third_party/wrappy/include",
                 "${ARSDK_ROOT}/out/arsdk-native/staging/usr/include",
                 "/usr/include/python3.7m",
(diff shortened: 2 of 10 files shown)


neworderofjamie:
I'll have a look at the actual code tomorrow but, ideologically, I don't think this is a sensible direction for BoB robotics to expand into. Like what's an actual use-case for this feature? You get a frame out of ant world into Python and then what? Unless all of BoB robotics is also wrapped in Python, you can't actually **do** anything so you might as well have just saved the frame to disk and messed with in Python (which I do all the time). IMO BoB robotics is a C++ library for building performant simulations and robots and having it expand outwards from those goals is a bad idea.

alexdewar:
I take your point, but this is mostly just for quick hacky testing rather than anything proper. We definitely don't need e.g. a whole SWIG wrapper for BoB robotics or something.

I think the guys have simulation code written in python, so this would be probably the easiest way to test existing simulation code without having to do a rewrite in C++. If you just want to quickly check whether something works or not you can use this module, but I agree that if it was for something "proper" (e.g. actual experiments) we should probably rewrite whatever algorithms we're using in C++.

neworderofjamie:
I understand that but, my concern is that without SWIG, what does this actually allow you do i.e. what do @efkag and @FullAPVayne actually want to achieve? I don't  see any examples where values are _returned_ from wrappy calls so it's hard to see how you could build something useful with a Python function "in the middle".

PR #152, opened 2019-08-24
Use Rothamsted models for gazebo_uav simulations

shinjanxp:


examples/gazebo_uav/iris_demo_rothamsted.world
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" ?>
```
alexdewar: I'd just call this file ``rothamsted.world``

PR #154, opened 2019-09-09
Restore ANSI colouring of compiler output

neworderofjamie:
Irritatingly, neither GCC nor Clang produce nice ANSI-coloured output if they detect that output "isn't a terminal - which seems to include whatever pipe-magick cmake includes.

Curtesy of https://medium.com/@alasher/colored-c-compiler-output-with-ninja-clang-gcc-10bfe7f2b949

### DIFF
--- a/cmake/bob_robotics.cmake
+++ b/cmake/bob_robotics.cmake
@@ -348,11 +348,11 @@ macro(BoB_build)
     # that output "isn't a terminal" - which seems to include whatever pipe-magick cmake includes.
     # https://medium.com/@alasher/colored-c-compiler-output-with-ninja-clang-gcc-10bfe7f2b949
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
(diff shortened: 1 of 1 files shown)


neworderofjamie:
Further usage suggests that the ANSIness doesn't seem to get applied when building the source code of the executable (rather than it's dependent modules) - not sure why as ``BoB_build`` is called from ``BoB_project``

alexdewar:
Not sure *exactly* why this is, but I swapped the ``add_compile_options()`` function for my ``add_compile_flags()`` function and this seems to fix things, at least for me.

neworderofjamie:
Ahhh ok - based on https://stackoverflow.com/questions/39501481/difference-between-add-compile-options-and-setcmake-cxx-flags - that sorta makes sense

neworderofjamie:
There are some exciting bleeding-edge arch things happening on your machine that are making the tests fail btw :smile:

alexdewar:
Ugh. I have no idea what those warnings are or what they mean :confused:

PR #158, opened 2019-09-20
CMake I2C fix

neworderofjamie:
This test was implemented backwards in the cmake version (see https://github.com/BrainsOnBoard/bob_robotics/blob/norbot_navigation/common/flags.mk#L94)

### DIFF
--- a/include/robots/norbot.h
+++ b/include/robots/norbot.h
@@ -8,6 +8,14 @@
 // Third-party includes
 #include "third_party/units.h"
 
+#if TEGRA_CHIP_ID == 33
+    #define USER_I2C_DEVICE "/dev/i2c-1"
+#elif TEGRA_CHIP_ID == 24
+    #define USER_I2C_DEVICE "/dev/i2c-0"
+#else
+    #error Unsupported tegra chip id
+#endif
+
 namespace BoBRobotics {
 namespace Robots {
 //----------------------------------------------------------------------------
@@ -20,7 +28,7 @@ class Norbot : public Tank
     using millimeter_t = units::length::millimeter_t;
 
 public:
-    Norbot(const char *path = "/dev/i2c-1", int slaveAddress = 0x29);
+    Norbot(const char *path = USER_I2C_DEVICE, int slaveAddress = 0x29);
 
     //----------------------------------------------------------------------------
     // Tank virtuals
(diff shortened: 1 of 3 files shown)


neworderofjamie:
I also subsequently found another issue where the user-facing I2C bus numbers are different between Jetson TX1 and TX2. As far as I can tell, there's no compile-time way of adding one so I attempted to do something in cmake - see what you think!

include/robots/norbot.h
```diff
@@ -8,6 +8,14 @@
 // Third-party includes
 #include "third_party/units.h"
 
+#if TEGRA_CHIP_ID == 33
+    #define USER_I2C_DEVICE "/dev/i2c-1"
+#elif TEGRA_CHIP_ID == 24
+    #define USER_I2C_DEVICE "/dev/i2c-0"
+#else
+    #error Unsupported tegra chip id
+#endif
```
alexdewar: I think this is a really nice solution. My only reservation is that we now can't test build it on a non-Tegra platform. How about wrapping this like so:
```c++
#ifdef TEGRA_CHIP_ID
... // your code
#else
    #define USER_I2C_DEVICE "/dev/null"
#endif
```

alexdewar:
I've just pushed a change because it wasn't building on my machine now. It turns out there were two bugs: the I2C test was inverted and the path for ``is_i2c_tools_new.sh`` was wrong. So my two fuck-ups happened to cancel each other out :wink:

include/robots/norbot.h
alexdewar: ...on second thoughts, you could do this without making it Tegra-specific at all. You could just make the default ctor try to open files from ``/dev/i2c-0`` to ``/dev/i2c-$some_number`` and stop when it finds the one that works.

I think @ajc158 might be using the I2C code on a Raspberry Pi and this change would break that. Have I got that right, Alex?

neworderofjamie:
Haha, I did wonder how it worked as actually finding the script (I just put it down to cmake magick). Hmmm, I think that's a little risky, as I understand it, I2C is used for fairly low level managent business on desktop machines (hence the smbus) - maybe defaulting to /dev/i2c-1 in the absence of tegra horror might make sense (and not change current behaviour)

alexdewar:
That's a good point. Or we could still default to ``/dev/null`` and then if you want to run it on e.g. a Raspberry Pi you'd just have to specify the path explicitly.

neworderofjamie:
Yeah, I think that makes sense. We can only default it on tegra - on other platforms you need to set the device yourself

neworderofjamie:
I have implemented your idea - what do you think? It doesn't appear to break anything when I try and build on a non-Jetson

alexdewar:
Looks good to me!

PR #159, opened 2019-09-24
Windows fixes

neworderofjamie:
A few things I needed to fix in order to get these modules building on Visual Studio 2019

### DIFF
--- a/src/video/rpi_cam.cc
+++ b/src/video/rpi_cam.cc
@@ -40,7 +40,7 @@ RPiCamera::getOutputSize() const
 bool
 RPiCamera::readFrame(cv::Mat &outFrame)
 {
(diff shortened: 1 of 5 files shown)


src/video/rpi_cam.cc
```diff
@@ -40,7 +40,7 @@ RPiCamera::getOutputSize() const
 bool
 RPiCamera::readFrame(cv::Mat &outFrame)
 {
-    uint8_t buffer[72 * 6 * 3];
+    char buffer[72 * 6 * 3];
```
alexdewar: Is there no ``uint8_t`` in Windows? I thought it lived in ``cstdlib`` -- not that it matters here.

PR #163, opened 2019-10-21
Mecanum

neworderofjamie:
This fixes a number of inconsistancies with the new mecanum robot (and unrelated things I discovered enroute)
1. Disables arduino code for turning off motors when there's no input as it doesn't play well with our code e.g. our joystick interface
2. Moved python stuff in ``navigation`` module out into ``navigation_plotting`` so navigation doesn't require python - which isn't necessarily something you want on a robot
3. Some bits of mecanum code were using x, y, rot and some forward, sideways and turn - in places they were swapped and it all went wrong - I think I have sorted this out!
4. The mecanum robot I have has its motors wired differently to the ones in sheffield :man_facepalming: so I've added a flag to switch between the two
5. Moved more functionality down to ``Robot`` base class so it's useful
6. Clearly I installed the OpenCV dependencies _too_ well and I was getting weird issues  with the code using gstreamer instead of V4L so I made it explicit for the panoramic camera - not sure how good an idea this is


arduino/i2c_slave_code_mecanum/i2c_slave_code_mecanum.ino
```diff
@@ -155,7 +155,7 @@ void loop() {
   } else {
     cnt = 0;
   }
-  has_serial = 0;
+  has_serial = 0;*/
```
alexdewar: Maybe bung in a comment saying why this is commented out

include/navigation_plotting/plot.h
```diff
@@ -4,11 +4,11 @@
...
 namespace BoBRobotics {
-namespace Navigation {
+namespace NavigationPlotting {
 
 //! Plot a line graph of a single RIDF
 void
 plotRIDF(const std::vector<float> &differences);
 
-} // Navigation
+} // NavigationPlotting
```
alexdewar: It might make more sense just to stick this in the ``Viz`` module rather than making a new one

projects/snapshot_bot/config.h
```diff
@@ -39,7 +39,9 @@ class Config
     bool shouldUseInfoMax() const{ return m_UseInfoMax; }
     bool shouldSaveTestingDiagnostic() const{ return m_SaveTestingDiagnostic; }
     bool shouldStreamOutput() const{ return m_StreamOutput; }
-
+    bool shouldUseNorbot() const{ return m_UseNorbot; }
+    bool shouldUseMecanum() const{ return m_UseMecanum; }
```
alexdewar: You can already set the type of tank robot you're using at compile time by either setting the ``TANK_TYPE`` env var or adding e.g. ``-DTANK_TYPE=EV3`` to the CMake args. I've been fiddling with that a bit on another branch, so I could add a similar mechanism to include Mecanum robots if you like?

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -311,98 +326,100 @@ class RobotFSM : FSM<State>::StateHandler
...
+                        // Calculate difference image
+                        cv::absdiff(matchedSnapshot, m_Unwrapped, m_DifferenceImage);
+
+                        char status[255];
+                        sprintf(status, "Angle:%f deg, Min difference:%f", degree_t(perfectMemory->getBestHeading()).value(), perfectMemory->getLowestDifference());
+                        cv::putText(m_DifferenceImage, status, cv::Point(0, m_Config.getUnwrapRes().height -20),
+                                    cv::FONT_HERSHEY_COMPLEX_SMALL, 1.0, 0xFF);
+
+                        // Send annotated difference image
+                        //m_NetSink.sendFrame(m_DifferenceImage);
```
alexdewar: Does the networking code work uncommented? I seem to remember that the only thing broken was that there was a change to the API for ``Net::Server`` so now you need to call ``waitForConnection()`` and use the ``Net::Connection`` object returned to send messages.

Alternatively you could delete this stuff if you don't want it.

src/video/panoramic.cc
```diff
@@ -80,7 +80,7 @@ std::unique_ptr<Input> getPanoramicCamera()
     }
     // PixPro
     else if(prefCamNum == 1) {
-        return std::make_unique<OpenCVInput>(deviceNum, cv::Size(1440, 1440), "pixpro_usb");
+        return std::make_unique<OpenCVInput>(deviceNum + cv::CAP_V4L, cv::Size(1440, 1440), "pixpro_usb");
```
alexdewar: What happens if you don't use ``cv::CAP_V4L``?

src/robots/mecanum.cc
```diff
@@ -13,29 +13,26 @@
 
 namespace BoBRobotics {
 namespace Robots {
-Mecanum::Mecanum(const char *path)
-  : m_Serial(path)
-  , m_Forward(0.0f)
-  , m_Sideways(0.0f)
-  , m_Turn(0.0f)
+Mecanum::Mecanum(const char *path, bool alternativeWiring)
+  : m_Serial(path), m_Forward(0.0f), m_Sideways(0.0f), m_Turn(0.0f), m_AlternativeWiring(alternativeWiring)
```
alexdewar: I'd probably make it so that ``alternativeWiring`` is by default set depending on the value of an env var (either at compile time or runtime), just so we don't end up with Sussex/Sheffield specific code.

projects/snapshot_bot/config.h
neworderofjamie: Sure! It doesn't exactly make a lot of sense to have this configured at runtime anyway...

src/video/panoramic.cc
neworderofjamie: it uses gstreamer which doesn't SEEM to support setting resolution - this isn't a nice solution though, I might add the a defaulted backend parameter to this function

include/navigation_plotting/plot.h
neworderofjamie: :+1:

projects/snapshot_bot/snapshot_bot.cc
neworderofjamie: I will have a go at re-enabling it

src/video/panoramic.cc
alexdewar: Ah I see. I think this is fine, but it maybe needs a comment in case this workaround doesn't work on someone else's machine.

projects/snapshot_bot/snapshot_bot.cc
neworderofjamie: Oh yeah right - I remember what the problem was now. Whether streaming is enabled or not is controllable at runtime so, as a ``Connection`` is required to make a ``NetSink``, I'd need to switch to using a ``std::unique_ptr<NetSink>`` which is fine *but*  ``Server::waitForConnection`` returns an un-copyable ``Connection`` which I can't create a unique pointer from...what are your thoughts?

src/video/panoramic.cc
neworderofjamie: See what you think of my improved solution (could also be useful for convincing @tenxlenx 's mac to work with a USB camera)

projects/snapshot_bot/snapshot_bot.cc
alexdewar: I wrote a move constructor for ``Net::Connection``s (though in retrospect I should probably just have returned a ``std::unique_ptr`` from ``waitForConnection()`` instead), so you should be able to do this:
```c++
auto p = std::make_unique<Net::Connection>(server.waitForConnection());
```

neworderofjamie: Woah! I had no idea you could move construct unique pointers from stack-allocated memory :astonished:

neworderofjamie:
Is this ok to merge? I've now handed the mecanum robot over to stathis (after a classic last minute fail of plugging the battery in backwards and blowing up the power supply) so it would be good if the code was on master

alexdewar:
Sure!

PR #164, opened 2019-10-29
Add support for selecting Mecanum robots at compile time

alexdewar:
Hey,

I've added support to the compile-time ``TANK_TYPE`` API for Mecanum robots (renamed to ``ROBOT_TYPE``) and tweaked ``snapshot_bot`` to use it. You choose to build support for Mecanum robots with ``snapshot_bot`` (or any of the other programs using ``ROBOT_TYPE``) like so:
```sh
cmake -DROBOT_TYPE=Mecanum ..
```
... or set the ``ROBOT_TYPE`` env var.

I made a couple of other fixlets on the way:
* Missing ``#include`` in ``robots.h``
* Stop ``Omni2D`` robots moving in the destructor, like we do for other robot classes


alexdewar:
No worries

PR #167, opened 2019-11-11
UTM branch

FullAPVayne:
0 modified map_coordinates.h by:
-- adding UTM coordinate support
-- adding operator definitions (+ and +=) for Cartesian
-- adding a function that transforms from LatLon to cartesian, adds a vector to it and return LatLon
-- adding a function that projects from UTM to cartesian,  , adds a vector to it and return LatLon

0 added experiments/shiftCoordTest/:
-- folder containing simple c++ file test the above describes utilities


include/common/map_coordinate.h
```diff
@@ -190,8 +267,8 @@ inline OSCoordinate latLonToOS(const LatLon<OSGB36> &latLon)
 
     const meter_t a = 6377563.396_m, b = 6356256.909_m;                 // Airy 1830 major & minor semi-axes
     constexpr double f0 = 0.9996012717;                                 // NatGrid scale factor on central meridian;
-    const radian_t phi0 = 49.0_deg, lambda0 = -2.0_deg;                 // NatGrid true origin is 49°N,2°W
-    const meter_t n0 = -100000.0_m, e0 = 400000.0_m;                    // northing & easting of true origin, metres
+    const radian_t phi0 = 49.0_deg, lambda0 = -2.0_deg;                 // NatGrid true utm is 49°N,2°W
+    const meter_t n0 = -100000.0_m, e0 = 400000.0_m;                    // northing & easting of true utm, metres
```
neworderofjamie: was this a bad search and replace?

include/common/map_coordinate.h
```diff
@@ -98,11 +97,89 @@ struct Cartesian
...
+    LatLon<Datum> G;
+    G.lat = degree_t(lat);
+    G.lon = degree_t(lon);
+    G.height = utm.height;
+
+    // Calculate new position in terms of lat long
+    G = shiftLatLon(G,v);
+    
+    // Convert back to UTM
+    using UTM::LLtoUTM;
```
neworderofjamie: If you only use it once, I would say there's no point in seperately bringing it into your local namespace, just call it with ``UTM::LLtoUTM``

include/common/map_coordinate.h
```diff
@@ -98,11 +97,89 @@ struct Cartesian
...
+        Cartesian<D> out;
+        out.x = x + coords.x;
+        out.y = y + coords.y;
+        out.z = z + coords.z;
+        return out;
+    };
+
+    void operator+=(const Cartesian<D> &coords)
+    {
+        x = x + coords.x;
```
neworderofjamie: you can just do ``x += coords.x`` here

include/common/map_coordinate.h
```diff
@@ -98,11 +97,89 @@ struct Cartesian
...
+
 };
 
+//! A standard GPS coordinate
+using GPSCoordinate = LatLon<WGS84>;
+
+//! Adds shift to a LatLon<Datum> by converting to Cartesian and adding
+//! a Cartesian v to it. 
+template<typename Datum>
+inline LatLon<Datum> shiftLatLon (LatLon<Datum> &latLon, Cartesian<Datum> &v)
```
neworderofjamie: When you're passing parameters by reference to save copying them, rather than because you want to modify them in the function, you should mark them as ``const`` i.e. ``const LatLon<Datum> &latLon``

examples/grid_experiment/grid.cc
```diff
@@ -0,0 +1,73 @@
...
+using GPSCoordinate = LatLon<WGS84>;
+
+int main()
+{
+    // test for GPS Coordinate by moving it 10 meters in x direction
+    double SHIFT = 30;
+
+    GPSCoordinate home;
+    home.lat = degree_t(50.87002350275288);
+    home.lon = degree_t(0.0018240860639551215);
```
neworderofjamie: If you put ``using namespace units::literals;`` at the top of this function you can then specify literals in degrees like ``50.87002350275288_deg``

examples/grid_experiment/grid.cc
```diff
@@ -0,0 +1,73 @@
...
+    home.height = meter_t(0);
+    
+    std::cout.precision(17);
+    std::cout << "Original Lat: " << std::setw(10)<< home.lat << "\n";
+    std::cout << "Original Long: " << std::setw(10) << home.lon << "\n";
+
+    CARCoordiante moveVector;
+    moveVector.x = meter_t(SHIFT);
```
neworderofjamie: Likewise, these can be specified as ``0_m``

examples/coordShiftTest/xRun.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+cd ./build 
+make -j 4
+cd ..
+./grid
+
```
neworderofjamie: I think this rather duplicates the functionality of the script in ``bin/bob_make`` so probably shouldn't be commited

examples/grid_experiment/grid.cc
```diff
@@ -0,0 +1,73 @@
+#include "common/map_coordinate.h"
```
neworderofjamie: This example doesn't seem to have any corresponding CMakeLists to build it

examples/coordShiftTest/shift.cc
```diff
@@ -0,0 +1,111 @@
...
+using BoBRobotics::MapCoordinate::cartesianToLatLon;
+using BoBRobotics::MapCoordinate::shiftLatLon;
+using BoBRobotics::MapCoordinate::shiftUTM;
+
+using UTM::LLtoUTM;
+using UTM::UTMtoLL;
+using CARCoordiante = Cartesian<WGS84>;
+using GPSCoordinate = LatLon<WGS84>;
+
+// template <typename T>
```
neworderofjamie: Maybe remove this

examples/coordShiftTest/shift.cc
```diff
@@ -0,0 +1,111 @@
+#include "common/map_coordinate.h"
```
neworderofjamie: Really small point - I think our directory naming convention is lower_snake_case so this should be ``coord_shift_test``

examples/coordShiftTest/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+cmake_minimum_required(VERSION 3.1)
+include(../../cmake/bob_robotics.cmake)
+BoB_project(SOURCES shift.cc
+            BOB_MODULES common
+	    THIRD_PARTY third_party_modules)
```
alexdewar: You don't need the ``THIRD_PARTY`` array here. That's just for third-party modules which are git submodules or require extra dependencies etc.

alexdewar: Also you don't want to commit your executables. I'd add them to a ``.gitignore`` file.

examples/coordShiftTest/shift.cc
alexdewar: :+1:

examples/coordShiftTest/shift.cc
```diff
@@ -0,0 +1,111 @@
...
+//     LLtoUTM(G.lat,G.lon,target.northing,target.easting,origin.zone);
+
+//     return target;
+
+// }
+
+int main()
+{
+    // test for GPS Coordinate by moving it 10 meters in x direction
+    double SHIFT = 30;
```
alexdewar: I'd use a ``meter_t`` type here

include/common/map_coordinate.h
```diff
@@ -98,11 +97,89 @@ struct Cartesian
...
+{
+    LatLon<Datum> target = cartesianToLatLon(latLonToCartesian(latLon) + v);
+    return target;
+}
+
+//! Adds shift to a UTM Coordinate by converting to LatLon<Datum> and then
+//! to Cartesian. Than add a Cartesian v to it and transfer back.
+//! a Cartesian v to it. 
+template<typename Datum>
```
alexdewar: I'm not sure if doxygen will like this format. For longer comments like this I'd split it up like so:
```c++
/**!
 * \brief Short comment
 *
 * longer description here
 */
```

include/common/map_coordinate.h
```diff
@@ -113,7 +190,7 @@ inline Cartesian<Datum> latLonToCartesian(const LatLon<Datum> &latLon)
     using namespace units::length;
     using namespace units::literals;
 
-    const meter_t h = 0.0_m; // height above ellipsoid - not currently used
+    const meter_t h = latLon.height; // height above ellipsoid - not currently used
```
alexdewar: I'd change the comment because now we are using it :-)

alexdewar:
Also if I were you I'd add your uni email to Github so it knows who your commits are by.

alexdewar:
Another thing: It looks like the test build is failing (though Jenkins isn't reporting this), because of something not related to your code. It's been fixed in ``master`` though, so I'd merge it in just so we can make sure that your branch is passing tests properly.

FullAPVayne:
Hey guys,
I have implemented all required fixes.
Thank you for the feedback,

All the best,
Stefan

include/common/map_coordinate.h
alexdewar: You actually need to put ``\brief`` explicitly here.

third_party/UTM.h
```diff
@@ -0,0 +1,256 @@
...
+     *
+     * East Longitudes are positive, West longitudes are negative.
+     * North latitudes are positive, South latitudes are negative
+     * Lat and Long are in fractional degrees.
+     *
+     * Written by Chuck Gantz- chuck.gantz@globalstar.com
+     */
+    template<typename T>
+    static inline void UTMtoLL(const T UTMNorthing, const T UTMEasting,
+                               const char* UTMZone, double& Lat,  double& Long )
```
alexdewar: I've only just noticed this... This function doesn't actually need to be templated. It looks like it'll only work with ``double``s anyhow. You could just make ``UTMNorthing`` and ``UTMEasting`` ``double``s or ``meter_t``s instead. Up to you which you go for.

alexdewar: Something else I forgot to mention before... I'd also add member functions for doing the conversions to ``LatLon`` and ``UTMCoordinate`` just to make it a cleaner interface. Then you could just write e.g. ``auto utm = myGPSCoords.toUTM();``.

alexdewar:
Cool, looks good. I've added a couple of minor comments. There are some things I forgot to mention before.

alexdewar:
Also as I said I'd merge in ``master`` branch to get some fixes for the build errors (see Jenkins output here: https://gen-ci.inf.sussex.ac.uk/job/BoB%20robotics/job/bob_robotics/job/UTM/14/).

One of the warnings actually is related to the UTM code. My gcc is complaining that where you do ``sprintf`` to get the UTM zone the result might be longer than the 4 bytes you've allotted for it (if the number you pass is huge). We know that the output won't be longer than that, so you could either use ``snprintf`` and specify that the maximum output is 4 bytes, or you could sack off using strings altogether and represent the UTM zone as e.g. a ``std::pair<int, char>``.

third_party/UTM.h
FullAPVayne: I am afraid I don't understand this change.

the functions 
latLonToCartesian(...) and
cartesianToLatLon(...) 

are elements of the MapCoordinate namespace and not members of any class, right?
Wouldn't it be more intuitive to have the conversion functions cartesianToUTM(...) etc. as elements of the namespace as well, instead of member functions of coordinate types?

neworderofjamie: Yeah, kinda agree with @FullAPVayne here. For better or worse, my original design was to have free funtions for doing the conversion and light structs to hold the coordinates so I see no particular reason not to keep doing that...

neworderofjamie: But also, this function is in a 3rd party library and, while I agree it's not an ideal signature, we probably shouldn't mess with it! My general assumption would be that things in third part remain unmodified unless we actually intend to upstream our modifications and this guy is probably not going to accept a PR adding a units requirement to his library!

FullAPVayne: Ok, I'll change back the UTM stuff and adapt the calls in map_coordinate.h
Guess I'll be done in a bit, I'll just push again and write another comment.

FullAPVayne:
> Also as I said I'd merge in `master` branch to get some fixes for the build errors (see Jenkins output here: https://gen-ci.inf.sussex.ac.uk/job/BoB%20robotics/job/bob_robotics/job/UTM/14/).
> 
> One of the warnings actually is related to the UTM code. My gcc is complaining that where you do `sprintf` to get the UTM zone the result might be longer than the 4 bytes you've allotted for it (if the number you pass is huge). We know that the output won't be longer than that, so you could either use `snprintf` and specify that the maximum output is 4 bytes, or you could sack off using strings altogether and represent the UTM zone as e.g. a `std::pair<int, char>`.

If I change this, I have to change it in UTM.h, which would collide again with @neworderofjamie concept of not changing third party code. So what shall I do?

FullAPVayne:
Hey guys,
So I did the following:
- reset UTM.h to original version
-- one small change: I omitted the ofMathConstants.h he includes and used two constants as #define instead
- adapted the map_coordinate.h to interface with UTM.h

Note: 
While both shiftUTM and shiftLatLon follow the same signature structure, lower level function calls within them are different.

The functions in shiftLatLon follow the classic out = function (in) format

```c++
template<typename Datum>
inline LatLon<Datum> shiftLatLon (const LatLon<Datum> &latLon, const Cartesian<Datum> &v)
{
    // Setup return value
    LatLon<Datum> target;

    // Convert to Cartesian
    Cartesian<Datum> C = latLonToCartesian(latLon);
    
    // Calculate new position
    C += v;

    // Convert back to LatLon
    target = cartesianToLatLon(C);

    return target;
}
```

The functions in shiftUTM follow function(in,out) format.

```c++
template<typename Datum>
inline UTMCoordinate shiftUTM (UTMCoordinate &utm, Cartesian<Datum> &v)
{
    // Setup return value
    UTMCoordinate target;
    target.height = utm.height;

    // Convert LL members to unit library classes
    Cartesian<Datum> C;
    utmToCartesian(utm,C);
    
    // Calculate new position
    C += v;
    
    // Convert Cartesian back to UTM
    cartesianToUTM(C, target);

    return target;
}
```
Reason being that going from UTM to Cartesian needs to be templated because of the datum, and I didnt know how to pass a type as a parameter so I can create the Cartesian<type> instance inside. So instead the user provides a Cartesian<Type> container which omits this problem.

third_party/UTM.h
alexdewar: Apologies for the late reply. This all sounds sensible to me.

alexdewar:
> Also as I said I'd merge in `master` branch to get some fixes for the build errors (see Jenkins output here: https://gen-ci.inf.sussex.ac.uk/job/BoB%20robotics/job/bob_robotics/job/UTM/14/).
> 
> One of the warnings actually is related to the UTM code. My gcc is complaining that where you do `sprintf` to get the UTM zone the result might be longer than the 4 bytes you've allotted for it (if the number you pass is huge). We know that the output won't be longer than that, so you could either use `snprintf` and specify that the maximum output is 4 bytes, or you could sack off using strings altogether and represent the UTM zone as e.g. a `std::pair<int, char>`.

Hey @FullAPVayne. I merged in ``master`` and it seems we're still getting this new warning. Other than that I'm happy to merge, but I don't want to break Jenkins :-)

PR #172, opened 2019-12-16
Gazebo updates

alexdewar:
This is just some tidying up and minor fixes for the Gazebo code, so we at least have some working examples now.

@tenxlenx if you wanted to do some BoB robotics stuff using Gazebo then you'll probably need these fixes.


alexdewar:
@neworderofjamie Any chance you could have a peek at this at some point? It's not terribly urgent -- just some tidying and tweaking of the Gazebo code with a view to making it more modular and integrated with the ``ROBOT_TYPE`` stuff etc.

cmake/bob_robotics.cmake
```diff
@@ -138,12 +116,15 @@ macro(BoB_project)
         endif()
     endif()
     string(TOUPPER ${ROBOT_TYPE} ROBOT_TYPE_UPPER)
+    string(REGEX REPLACE :: _ ROBOT_TYPE_UPPER ${ROBOT_TYPE_UPPER})
```
neworderofjamie: I think I get the logic here but some comments would be nice

cmake/bob_robotics.cmake
```diff
@@ -207,6 +188,31 @@ macro(BoB_module_custom)
     # Check we're on a supported platform
     check_platform(${PARSED_ARGS_PLATFORMS})
 
+    if(PARSED_ARGS_GAZEBO_PLUGINS)
+        # I'm sometimes getting linker errors when ld is linking against the
+        # static libs for BoB modules (because Gazebo plugins, as shared libs,
+        # are PIC, but the static libs seem not to be). So let's just compile
+        # everything as PIC.
+        if(GNU_TYPE_COMPILER)
+            add_definitions(-fPIC)
```
neworderofjamie: Having hit thrilling issues of this sort with GeNN, I **think** you can use ``-fPIE`` here which may perform a tiny bit better

cmake/bob_robotics.cmake
```diff
@@ -256,12 +262,15 @@ macro(always_included_packages)
     if(NOT TARGET GLEW::GLEW)
         find_package(GLEW QUIET)
     endif()
+    if(NOT TARGET FreeImage::FreeImage)
```
neworderofjamie: does gazebo include freeimage somehow?

cmake/bob_robotics.cmake
```diff
@@ -207,6 +188,31 @@ macro(BoB_module_custom)
...
+        endif()
+
+        # We need to link against Gazebo libs
+        BoB_external_libraries(gazebo)
+
+        # Dump plugins into source dir
+        set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR})
+
+        foreach(plugin IN LISTS PARSED_ARGS_GAZEBO_PLUGINS)
+            get_filename_component(shortname ${plugin} NAME)
```
neworderofjamie: what does this do?

cmake/external_libs/gazebo.cmake
```diff
@@ -1,10 +1,7 @@
 # If Gazebo is added as a dependency multiple times (e.g. from
```
neworderofjamie: I guess this comment is a little out of date?

cmake/bob_robotics.cmake
alexdewar: Good point. This is a bit opaque...

cmake/bob_robotics.cmake
alexdewar: It's good to know about this option but it seems not to work here :-(. SO says that it may not play nice with the ``-static`` option, which could be something to do with it: https://stackoverflow.com/questions/2463150/what-is-the-fpie-option-for-position-independent-executables-in-gcc-and-ld

neworderofjamie: ahh my bad - that was probably the solution to another random problem :smile:

cmake/bob_robotics.cmake
alexdewar: Yeah. Unfortunately it tries to create the ``FreeImage::FreeImage`` target unconditionally, so if you do ``find_package(gazebo)`` twice it fails because it already exists and errors out. Ugh.

cmake/bob_robotics.cmake
alexdewar: Gets the filename from the path. Then we trim off the extension to get the target name.

I've stuck in a comment explaining this

cmake/bob_robotics.cmake
neworderofjamie: god I hate build systems

cmake/external_libs/gazebo.cmake
alexdewar: True!

PR #174, opened 2020-01-17
Update Stone CX

neworderofjamie:
Mostly due to my past self, the Stone CX robot example was a bit of a shit-show of macros. I've replaced most of them with a little class hierachy to abstract away the various bits of hardware and, on the mecanum robot with the BN055, it works pretty nice. Is there something that makes the EV3 a special case (i.e. doesn't ``ROBOT_TYPE`` cover it) or was the ``USE_EV3`` business just defensive programming against past-me? If the latter I was thinking I'd add a better CLI so you can chose between the smorgasbord of options now available :)

Also, again probably due to past me, the getters for forward, sideways and turn were in ``Robots::Mecanum`` rather than ``Robots::Omni2D`` breaking any chance of stuff working with future omniwheelers (and quite possibly breaking ``Omni2D::onJoystickEvent`` which used them)


alexdewar:
Sorry about the somewhat long delay in looking at this ;-). I seem to remember that the ``USE_EV3`` macro predates the whole ``ROBOT_TYPE`` business so it's probably unnecessary now. I'll take a look at things now

projects/stone_cx/robot.cc
```diff
@@ -1,170 +1,413 @@
...
         }
 
         // Read magneto
         float magnetoData[3];
-        imu.readMagneto(magnetoData);
+        m_IMU.readMagneto(magnetoData);
 
         // Calculate heading angle from magneto data and set atomic value
-        heading = atan2(magnetoData[0], magnetoData[2]);
+        return radian_t(std::atan2(magnetoData[0], magnetoData[2]));
```
alexdewar: Or you could use ``units::math::atan2`` here. Either way's fine though

projects/stone_cx/robot.cc
```diff
@@ -1,170 +1,413 @@
...
+    {
+    }
+    
+    virtual void drive(float left, float right, bool log) override
+    {
+        // Calculate differential steering signal
+        const float steering = left - right;
+        if(log) {
+            LOGI << "Steer:" << steering;
+        }
```
alexdewar: It might be easier just to use ``LOGD`` and then if you do want to see the output you can just run ``LOG_LEVEL=DEBUG ./my_program``

projects/stone_cx/robot.cc
```diff
@@ -1,170 +1,413 @@
...
+
+        // Create server and sink for sending activity image over network
+        server = std::make_unique<Net::Server>();
+        connection = std::make_unique<Net::Connection>(server->waitForConnection());
+        netSink = std::make_unique<Video::NetSink>(*connection,
+                                                   cv::Size(activityImageWidth, activityImageHeight),
+                                                   "activity");
+        connection->runInBackground();
+    }
+#endif
```
alexdewar: There should be a macro ``ROBOT_TYPE_EV3`` defined when using the EV3 robot, so we shouldn't need ``USE_EV3`` anywhere now.

alexdewar: ...and while we're on the subject of ``ROBOT_TYPE``s, FYI there is a ``BundledTankNetSink`` class which sets up the ``Server`` and listens etc. without any extra logic needed. It's a bit of a hack, but it does make it easy to switch between remote and non-remote robots. (Not saying you should necessarily use it here though)

alexdewar:
> All looks good!

...although maybe check the macro stuff I mentioned.

projects/stone_cx/robot.cc
neworderofjamie: ahh - is **that** how you set the bob robotics log-level :smile:

projects/stone_cx/robot.cc
neworderofjamie: I get a ``error: static assertion failed: The quantity y/x must yield a dimensionless ratio.`` if I switch this to ``units::math::atan2`` - any thoughts?

neworderofjamie:
What do you think @alexdewar

alexdewar:
Looks good to me

PR #175, opened 2020-01-31
Turned on LFS and added Seville environment

neworderofjamie:
I'm bored of having to keep scping it between my machines and send google drive links to everyone so I think it makes sense to add the (slightly-postprocessed) version of this environment to bob robotics. It's already online at https://insectvision.dlr.de/3d-reconstruction-tools/habitat3d so I can't see why it would be an issue


neworderofjamie:
@FullAPVayne - if you update to master this means you won't need my google drived seville model :smile:

PR #176, opened 2020-02-05
AAAS demo

neworderofjamie:
This work is primarily for the AAAS demo but we can definitely use it to improve the snapshot bot demo we do at other outreach and it might be useful for use while doing actual experiments too. Basic additions are:
- Added support for cropping image to snapshot bot - theory is it'll be useful for chopping out looming people in demo environment
- More elaborate streaming 'protocol' between snapshot bot and new visualizer
- Added 'pause' button on A so testing can be paused e.g. to point out if the robot is doing something excitingly dense
- New visualizer tool (out of date video [here](https://drive.google.com/open?id=1kC7rF9MgrmEpbiGSiv2VVa9jpf2kPUv1)) 
  - Shows 'live' view of what robot sees, 
  - When training, shows snapshots getting stored in perfect memory
  - When testing, shows best matching and current snapshot, RIDF and difference image
  - Includes some passable B.O.B. branding but that can be easily customized by changing image + yaml


tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
```diff
@@ -0,0 +1,436 @@
...
+                    const int numRectangles = m_Config.getTrainingSnapshotRects().size();
+
+                    const int numSnapshotsToDisplay = std::min(numTrainingSnapshots, numRectangles);
+                    const int startDisplaySnapshot = std::max(0, numTrainingSnapshots - numRectangles);
+
+                    // Loop through available rectangles
+                    for(int i = 0; i < numSnapshotsToDisplay; i++) {
+                        resizeImageIntoOutputROI(m_Config.getTrainingSnapshotRects()[i],
+                                                 m_TrainingSnapshots[startDisplaySnapshot + i]);
+                    }
```
alexdewar: What are the rectangles for? Are you cropping training images to different sizes or something?

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
```diff
@@ -0,0 +1,436 @@
...
+                }
+
+                // Show output image
+                cv::imshow("Output", m_OutputImage);
+
+                // Read input and update
+                const int key = cv::waitKey(20);
+                if(key == 27) {
+                    return false;
+                }
```
alexdewar: To make it cross-platform, you can do this:
```c++
#include "os/keycodes.h"
...
                // Read input and update
                const int key = cv::waitKey(20);
                if((key & OS::KeyMask) == OS::KeyCodes::Escape) {
                    return false;
                }
```

alexdewar: I'm not sure why OpenCV doesn't make ``waitKey()`` more cross-platform to begin with :confused:

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
```diff
@@ -0,0 +1,436 @@
...
+                            (const cv::Mat &fr, const cv::Mat &, size_t i)
+                            {
+                                // Calculate image difference
+                                auto diffIter = differencer(fr, m_BestSnapshot, m_SnapshotDifferenceScratch);
+
+                                float sumDifference = std::accumulate(diffIter, diffIter + imageSize, 0.0f);
+
+                                // Store mean difference in RIDF
+                                m_RIDF[numScanColumns - 1 - i] = Navigation::AbsDiff::mean(sumDifference, imageSize);
+                            });
```
alexdewar: Why are you filling ``m_RIDF`` from the end here?

alexdewar: Ahh I see. nvm.

An alternative to having ``rotatorLeft`` and ``rotatorRight`` would be to just have a single ``rotator`` then doing ``std::rotate`` on it. But this is fine too

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
```diff
@@ -0,0 +1,436 @@
...
+
+                                float sumDifference = std::accumulate(diffIter, diffIter + imageSize, 0.0f);
+
+                                // Store mean difference in RIDF
+                                m_RIDF[(2 * numScanColumns) - 1 - i] = Navigation::AbsDiff::mean(sumDifference, imageSize);
+                             });
+
+                    // Get range of
+                    const float minRIDF = *std::min_element(m_RIDF.cbegin(), m_RIDF.cend());
+                    const float maxRIDF = *std::max_element(m_RIDF.cbegin(), m_RIDF.cend());
```
alexdewar: Could use ``std::minmax`` instead here

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
```diff
@@ -0,0 +1,436 @@
...
+    {
+        if(command[1] == "TRAINING") {
+            m_SnapshotBotState = State::Training;
+        }
+        else if(command[1] == "TESTING") {
+            m_SnapshotBotState = State::Testing;
+        }
+        else {
+            LOGE << "Unknown snapshot bot state '" << command[1] << "'";
+        }
```
alexdewar: Maybe you want to make this error fatal?

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
alexdewar: ...although, on third thoughts, I think I might not understand this after all :wink:. Is it just that you want the RIDF centred on 0°? If so, you still wouldn't want ``m_RIDF`` reverse-filled, just offset... right?

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
```diff
@@ -0,0 +1,436 @@
...
+                        cv::line(roiRIDF, lastPoint, point, m_Config.getTestingRIDFLineColour(), m_Config.getTestingRIDFLineThickness(), cv::LINE_AA);
+                        lastPoint = point;
+                    }
+
+                    // Get index of min RIDF element and hence calculate pixels to roll snapshot
+                    // **YUCK** this information was clearly present deep within Navigation::InSilicoRotater
+                    const int minRIDFIdx = (int)std::distance(m_RIDF.cbegin(), std::min_element(m_RIDF.cbegin(), m_RIDF.cend()));
+                    const int rollPixels = (minRIDFIdx < numScanColumns) ? (numScanColumns - 1 - minRIDFIdx) : (imageWidth + numScanColumns - 1 - minRIDFIdx);
+                    BOB_ASSERT(rollPixels >= 0);
+                    BOB_ASSERT(rollPixels < imageWidth);
```
alexdewar: Out of interest, is there a reason that you didn't use ``PerfectMemoryRotater`` for this? (Other than the sprawling confusing interface :wink:).

``PerfectMemory::getHeading()`` returns a tuple with the best-matching snapshot as the second element (if you're using ``BestMatchingSnapshot``).

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
neworderofjamie: yeah, they seemed the most intuitive way of expressing scaling of snapshots into areas of the screen

alexdewar: Ah I see

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
neworderofjamie: So we're only rotating +-90 degrees and that gives you two small RIDFs, one of which is reversed. This code combines them together into an RIDF centred at 0 degrees as you say.

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
neworderofjamie: ooh I aways forget about that beast

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
neworderofjamie: Hmm, good point. Using a perfect memory, adding the single image and then clearing it would probably be nicer than this nightmare

tools/snapshot_bot_remote_display/snapshot_bot_remote_display.cc
neworderofjamie: oh yey! I'd forgotton about that horror - thanks

neworderofjamie:
I realised that, amongst the many loose ends I've left in BoB robotics, this is one. Have implemented the smaller things (as it works, I'm reticent to do any of the more complex ones in the absence of a robot!). Is this now ok to merge?

alexdewar:
This looks good and I'd be happy to merge, although the CI is currently failing on my machine because of the new Gazebo/C++17 problem.

I've merged in ``master`` and pushed, so hopefully that'll fix things.

alexdewar:
Success!

PR #178, opened 2020-02-11
Ardin MB fixes

neworderofjamie:
As you witnessed, the version of this project in master wasn't in a state where it could easily be used for anything (apologies @FullAPVayne)! This P.R. fixes the following:
- Adds a CMake option ``NO_GENN`` which doesn't build the GeNN mushroom body and defaults to using the mushroom body
- Fixed outdated SFML key enum
- Fixed incorrect SFML event processing (you couldn't close the window)
- Added a proper CLI using CLI11 to enable you to load different world models and tweak the bits required to make it work with them e.g. loads the new seville ant world model in such a way that it works
  ``./ardin_mb --world ../../resources/antworld/seville_vegetation_downsampled.obj --height 1.6 --min-bound 0 0 0 --max-bound 10 10 2.8 --clear-colour 0.8 0.8 0.8 1.0``
- Tweaked the way the top-down view gets rendered - there was a weird swapping of x and y in the previous rendering code which definitely shouldn't have been there and I am 90% the answer was that it was rendering the world upside down
- Tweaked the clipping when rendering the world top-down so big environments don't get clipped

### DIFF
--- a/projects/ardin_mb/ardin_mb.cc
+++ b/projects/ardin_mb/ardin_mb.cc
@@ -89,6 +89,15 @@ int main(int argc, char *argv[])
     // Loop until window should close
     sf::Event event;
     for (unsigned int frame = 0; window.isOpen(); frame++) {
+        // Process events
+        sf::Event event;
+        while (window.pollEvent(event)) {
+            // Close window: exit
+            if (event.type == sf::Event::Closed) {
+                window.close();
+            }
+        }
+
         // Clear colour and depth buffer
         glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 
--- a/projects/ardin_mb/ardin_mb.cc
+++ b/projects/ardin_mb/ardin_mb.cc
@@ -55,6 +55,7 @@ int main(int argc, char *argv[])
     glLineWidth(4.0);
     glPointSize(4.0);
 
+    Logging::Logger::getInstance();
     // Create memory
 #ifdef NO_GENN
     Navigation::PerfectMemory<> memory(cv::Size(MBParams::inputWidth, MBParams::inputHeight));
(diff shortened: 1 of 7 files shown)


projects/ardin_mb/ardin_mb.cc
```diff
@@ -50,25 +69,87 @@ int main(int argc, char *argv[])
...
     glPointSize(4.0);
 
     // Create memory
-    //Navigation::PerfectMemory<> memory(cv::Size(MBParams::inputWidth, MBParams::inputHeight));
+#ifdef NO_GENN
+    Navigation::PerfectMemory<> memory(cv::Size(MBParams::inputWidth, MBParams::inputHeight));
     //Navigation::InfoMax<float> memory(cv::Size(MBParams::inputWidth, MBParams::inputHeight), 0.01f);
+#else
     MBMemory memory;
+#endif
```
alexdewar: This is a really minor nit, but I feel like ``NO_GENN`` probably isn't the best name for this option. It's not obvious that ``NO_GENN`` means perfect memory instead of MB model.

But it's not super important. I would probably stick some ``message()``s in the ``CMakeLists.txt`` file though saying which algo it's building for just to make it obvious.

projects/ardin_mb/state_handler.cc
```diff
@@ -55,7 +73,7 @@ bool StateHandler::handleEvent(State state, Event event)
     if(event == Event::Update) {
 
         // Render top down and ants eye view
-        m_Renderer.renderPanoramicView(m_Pose.y(), m_Pose.x(), 0.01_m,
+        m_Renderer.renderPanoramicView(m_Pose.x(), m_Pose.y(), m_PathHeight,
```
alexdewar: Out of interest, why were the x and y inverted before and why not now?

projects/ardin_mb/ardin_mb.cc
```diff
@@ -35,6 +34,26 @@ using namespace BoBRobotics;
...
 int main(int argc, char *argv[])
 {
+    // Default parameters"
+    std::string worldFilename = "";
+    std::string routeFilename = "";
+    std::string logFilename = "";
+    float heightMetres = 0.01f;
+    std::vector<float> minBound;
+    std::vector<float> maxBound;
+    std::vector<float> clearColour{0.0f, 1.0f, 1.0f, 1.0f};
```
alexdewar: I'm guessing ``std::array<...>`` wouldn't work instead here?

projects/ardin_mb/state_handler.cc
alexdewar: Nvm -- I just noticed that you said why in the PR (which I probably should have read before the code review rather than after :confused: )

projects/ardin_mb/ardin_mb.cc
neworderofjamie: Sadly not - I would agree that that would be the correct container for this use-case but [only std::vector is supported](https://github.com/BrainsOnBoard/bob_robotics/blob/master/third_party/CLI11.hpp#L2646-L2688)

projects/ardin_mb/ardin_mb.cc
neworderofjamie: ``MEMORY_TYPE`` in the style of robot type would perhaps be the best solution

projects/ardin_mb/state_handler.cc
neworderofjamie: I had a quick look through the history and exactly what happened here isn't clear but still...

PR #179, opened 2020-02-17
Automatically delete old builds on Jenkins

neworderofjamie:
I noticed that the Jenkins server is basically out of disk space. I did some manual deleting of old build but this **should** set some automatic build discarding rules based on https://stackoverflow.com/questions/47778019/how-to-use-options-in-a-jenkins-scripted-pipeline - if you think archiving the 100 latest builds isn't ideal, feel free to change

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,12 @@
 #!groovy​
 
+// only keep 100 builds to prevent disk usage from growing out of control
+properties([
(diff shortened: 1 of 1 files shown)


alexdewar:
No, this seems all good. I was waiting for the test build to finish, then I sort of forgot about this PR...

alexdewar:
Do we still want to do this, @neworderofjamie ?

neworderofjamie:
yeah - this remains an issue!

PR #182, opened 2020-04-16
Add support for C++17 and Gazebo v11+

### LINKED ISSUE #103: Upgrade to C++17
C++17 adds a number of useful features, among them inline variables, which would mean we could get rid of the ``NO_HEADER_DEFINITIONS`` hack we're using all over the place.

The latest version of MSVC has good support for C++17 as do gcc and clang. The version of gcc in Ubuntu 16.04 (our current de facto lowest common denominator) is 5.3, which has rather poor C++17 support (see https://gcc.gnu.org/projects/cxx-status.html). So it would seem sensible to bump the minimum required version of gcc to v7+ at the same time (v7 is included in Ubuntu 18.04, the latest LTS release).

alexdewar:
Closes #103.

The newest version of Gazebo requires C++17 and seeing as we were thinking of upgrading to C++17 anyway I thought now might be the time to do it. g++ v7 seems to have good support (https://gcc.gnu.org/projects/cxx-status.html) and that's the default in Ubuntu 18.04 so I think this should be relatively safe to do. Plus it would be handy to have inline variables in a few places instead of nasty macro hacks.

### DIFF
--- a/cmake/bob_robotics.cmake
+++ b/cmake/bob_robotics.cmake
@@ -16,7 +16,7 @@ macro(BoB_project)
     include(CMakeParseArguments)
     cmake_parse_arguments(PARSED_ARGS
                           "GENN_CPU_ONLY"
-                          "EXECUTABLE;GENN_MODEL"
+                          "EXECUTABLE;GENN_MODEL;CXX_STANDARD"
                           "SOURCES;BOB_MODULES;EXTERNAL_LIBS;THIRD_PARTY;PLATFORMS;OPTIONS"
                           "${ARGV}")
     BoB_set_options()
@@ -35,6 +35,11 @@ macro(BoB_project)
         get_filename_component(NAME "${CMAKE_CURRENT_SOURCE_DIR}" NAME)
     endif()
     project(${NAME})
+
+    # Allow for setting the C++ standard on a per-project basis.
+    if(PARSED_ARGS_CXX_STANDARD AND NOT DEFINED CMAKE_CXX_STANDARD)
+        set(CMAKE_CXX_STANDARD ${PARSED_ARGS_CXX_STANDARD})
+    endif()
 
     # Include local *.h files in project. We don't strictly need to do this, but
     # if we don't then they won't be included in generated Visual Studio
--- a/cmake/bob_robotics.cmake
+++ b/cmake/bob_robotics.cmake
@@ -392,7 +392,7 @@ macro(BoB_build)
     # Conversely, only setting the compiler flag means that the surveyor example
     # mysteriously gets linker errors on Ubuntu 18.04 and my Arch Linux machine.
     #       - AD
-    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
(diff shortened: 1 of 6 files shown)


neworderofjamie:
My only concern here is the Jetsons. As far as I can recall, they all run Ubuntu 16.04 which comes with GCC 5.4 which has only got very partial C++17 support. Can we now built gazebo with whatever C++ standard it needs and leave the rest alone?

alexdewar:
Arg, that's a shame. Yeah, we could have it as an extra build option.
Alternatively, we could just make sure we don't use any C++17-isms for
code that'll run on Jetsons. What do you think?

cmake/bob_robotics.cmake
```diff
@@ -365,17 +365,24 @@ macro(BoB_build)
...
+    endif()
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+    # On Ubuntu 16.04, seemingly setting CMAKE_CXX_STANDARD by itself doesn't
     # work, so add the compiler flag manually.
     #
     # Conversely, only setting the compiler flag means that the surveyor example
     # mysteriously gets linker errors on Ubuntu 18.04 and my Arch Linux machine.
     #       - AD
     if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
neworderofjamie, line 382: does this get set for clang as well? I know in some contexts e.g. #defining __GNUC__ clang pretends to be GCC but not sure about this one

cmake/bob_robotics.cmake
```diff
@@ -365,17 +365,24 @@ macro(BoB_build)
         set(CMAKE_EXE_LINKER_FLAGS "-Wl,--allow-multiple-definition")
     endif()
 
-    # Use C++14. On Ubuntu 16.04, seemingly setting CMAKE_CXX_STANDARD doesn't
+    # If C++ standard has not been specified explicitly either in a
+    # CMakeLists.txt file or on the command line, set the standard to C++14, the
+    # minimum supported by BoB robotics.
```
neworderofjamie, line 370: Unless we actually plan on updating BoB robotics to use a bunch of C++17 features I think it might be better to default to C++14 (as the comment says 😄) and set the standard to C++17 for gazebo projects

alexdewar:
Yeah I think you're right that it's probably a bit overkill to enable it anywhere...

So I've come up with a new approach: allow for setting the C++ standard on a per-project basis, so we can enable it for the Gazebo examples (so that it'll compile on my desktop machine for the CI). Then users can choose whether to use it or not, if they happen to fancy using some C++17 features.

You can also now set it globally with the ``BOB_ROBOTICS_CXX_STANDARD`` environment variable.

Does this sound sensible?

cmake/bob_robotics.cmake
alexdewar: No, this'll only work for gcc. It *might* not be an issue depending on whether or not clang also needs this hack in order to set the c++ standard correctly, but I've now also enabled it for clang for consistency.

alexdewar:
Cool. Thanks for reviewing :-)

PR #184, opened 2020-05-06
Fix logging initialisation

alexdewar:
For some reason, plog isn't printing its output for me any more and I suspect it's because it isn't being properly initialised. As we're currently relying on static initialisation to start plog, it seemed like a good idea to do this in a less hacky way anyway.

So I've now moved the plog initialisation into the ``main()`` function in ``main.cc`` and replaced all uses of ``main()`` elsewhere with ``bobMain()``. This is a bit cleaner -- it lets us drop ``common/main.h`` and ``common/logging.h`` -- but it does force users to define a ``bobMain()`` rather than a ``main()`` function. We should probably stick a big warning about this in the documentation somewhere.

I based this branch off #183 so we don't get a bunch of merge errors.


projects/snapshot_bot/offline_train.cc
```diff
@@ -3,7 +3,7 @@
 #include "memory.h"
 
 // BoB robotics includes
-#include "common/logging.h"
+#include "plog/Log.h"
 
 int main(int argc, char *argv[])
```
neworderofjamie, line 8: one main missing!

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -6,7 +6,7 @@
 
 // BoB robotics includes
 #include "common/fsm.h"
-#include "common/logging.h"
+#include "plog/Log.h"
```
neworderofjamie, line 9: there's another main deep down here! I wonder if your regex didn't get ones with arguments

alexdewar: Hmm it looks like I missed loads, partly because I forgot to commit even the ones I did find :blush:

PR #185, opened 2020-05-12
New looser GeNN integration and updated Ardin Webb 

neworderofjamie:
Rather than tightly link the build systems, GeNN can also operate in a mode where you dynamically load the model dll and call functions from it (using dl on *nix). This is much simpler to integrate, works painlessly on Windows (or at least past-me already suffered the pain) and lets you write code that loads different GeNN models (which wouldn't otherwise work as, because the generated code is essentially C, they otherwise clash). 

Using this and imgui this PR also provides a nice GUI for visualizing what's going on in the Ardin model:

a_windows_miracle

this also uses AntWorld render targets to write rendered output to texture which, at least on my machines, is much faster than the approach used in the older examples.

There is a weird issue with the top-down view of the old antworld being upside down which I still need to figure out but a review of the rest would be awesome (this branch has been malingering around for way too long having its bones gradually cherry-picked away)


projects/ardin_mb/state_handler.cc
```diff
@@ -367,14 +486,111 @@ bool StateHandler::checkAntPosition()
...
+    // If loading route is successful
+    m_Route.load(filename);
+
+    // Get bounds of route
+    const auto &routeMin = m_Route.getMinBound();
+    const auto &routeMax = m_Route.getMaxBound();
+
+    // Create vector field geometry to cover route bounds
+    m_VectorField.createVertices(routeMin[0] - 20_cm, routeMax[0] + 20_cm, 20_cm,
+                                 routeMin[1] - 20_cm, routeMax[1] + 20_cm, 20_cm);
```
alexdewar, line 500: I'm guessing the +-20cm is just so that there's a bit of a margin in the display? Maybe a comment is needed

projects/ardin_mb/state_handler.cc
```diff
@@ -367,14 +486,111 @@ bool StateHandler::checkAntPosition()
...
+    if(ImGui::BeginMainMenuBar()) {
+        if(ImGui::BeginMenu("File")) {
+            if(ImGui::BeginMenu("Open Route")) {
+                char routeFilename[32];
+                for(unsigned int r = 1; r <= 14; r++) {
+                    sprintf(routeFilename, "ant1_route%u.bin", r);
+                    if(ImGui::MenuItem(routeFilename)) {
+                        const char *bobRoboticsPath = std::getenv("BOB_ROBOTICS_PATH");
+                        assert(bobRoboticsPath != nullptr);
+                        std::string routePath = std::string(bobRoboticsPath) + "/resources/antworld/";
```
alexdewar, line 560: I'd use the helper function in ``path.h`` for this, partly so that you can get the correct path even when ``BOB_ROBOTICS_PATH`` isn't set.

projects/ardin_mb/state_handler.h
```diff
@@ -118,21 +131,23 @@ class StateHandler : public BoBRobotics::FSM<State>::StateHandler
...
-    BoBRobotics::AntWorld::SnapshotProcessorArdin m_SnapshotProcessor;
+    //! OpenCV texture wrapper used to render final snapshot
+    OpenCVTexture m_FinalSnapshotTexture;
 
     //! Class for handling rendering of vector field
     VectorField m_VectorField;
 
     //! Ant's current pose
-    BoBRobotics::Pose2<meter_t, degree_t> m_Pose;
+    BoBRobotics::Pose3<meter_t, degree_t> m_Pose;
```
alexdewar, line 150: How come the change to 3d, btw?

projects/stone_cx/simulator.cc
```diff
@@ -13,7 +13,9 @@
 // Common includes
 #include "plog/Log.h"
 #include "common/von_mises_distribution.h"
-#include "genn_utils/analogue_csv_recorder.h"
+
+// GeNN userproject includes
+#include "analogueRecorder.h"
```
alexdewar, line 18: This is a bit off-topic, but maybe it would make more sense to have these headers in ``genn/utils`` or something instead of with the userprojects.

projects/ardin_mb/mb_memory.h
```diff
@@ -29,13 +55,162 @@ class MBMemory : public BoBRobotics::Navigation::VisualNavigationBase
...
+    template<typename R>
+    void calcImageDifferences(R &rotater) const
+    {
+        // Ensure there's enough space in rotated differe
+        m_RotatedDifferences.reserve(rotater.numRotations());
+        m_RotatedDifferences.clear();
+
+        // Populate rotated differences with results
+        rotater.rotate([this] (const cv::Mat &image, auto, auto) {
+            m_RotatedDifferences.push_back(this->test(image));
```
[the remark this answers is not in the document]
neworderofjamie: I've never encountered that one - I thought explicit ``this`` was **normally** only required when templated functions are somehow involved...

projects/stone_cx/simulator.cc
neworderofjamie: that would probably be nicer but probably not enough to break backward compatibility

projects/ardin_mb/state_handler.h
neworderofjamie: FlyingAnts4

projects/ardin_mb/plot.py
```diff
@@ -61,11 +61,11 @@ def get_column_safe(data, column, dtype):
...
 
-        pn_spike_mask = (pn_spike_times >= start_time) & (pn_spike_times < end_time)
-        kc_spike_mask = (kc_spike_times >= start_time) & (kc_spike_times < end_time)
+        ##pn_spike_mask = (pn_spike_times >= start_time) & (pn_spike_times < end_time)
+        #kc_spike_mask = (kc_spike_times >= start_time) & (kc_spike_times < end_time)
 
-        print "\tPN:%u" % len(np.unique(pn_spike_neuron_id[pn_spike_mask]))
-        print "\tKC:%u" % len(np.unique(kc_spike_neuron_id[kc_spike_mask]))
+        print "\tPN:%u (%u)" % (len(np.unique(pn_spike_neuron_id)), len(pn_spike_neuron_id))
+        print "\tKC:%u (%u)" % (len(np.unique(kc_spike_neuron_id)), len(kc_spike_neuron_id))
```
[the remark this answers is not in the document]
neworderofjamie: Python 2.7 clearly seemed a much more viable choice in 2018. Bigger problem was that the code that produced the data for most of these scripts was no longer present. I've deleted them but added buttons for saving spike data to the GUI as that seems useful.

projects/ardin_mb/mb_memory.cc
```diff
@@ -75,149 +115,158 @@ void MBMemory::train(const cv::Mat &image)
...
+    BOB_ASSERT(image.cols == m_InputWidth);
+    BOB_ASSERT(image.rows == m_InputHeight);
+    BOB_ASSERT(image.type() == CV_8UC1);
 
-    // Convert simulation regime parameters to timesteps
-    const unsigned long long rewardTimestep = convertMsToTimesteps(MBParams::rewardTimeMs);
-    const unsigned int endPresentTimestep = convertMsToTimesteps(MBParams::presentDurationMs);
-    const unsigned int postStimuliDuration = convertMsToTimesteps(MBParams::postStimuliDurationMs);
+     // Convert simulation regime parameters to timesteps
+    const unsigned long long rewardTimestep = convertMsToTimesteps(m_RewardTimeMs);
```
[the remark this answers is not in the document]
alexdewar: Haha, we're probably safe then ;)

projects/ardin_mb/mb_memory.h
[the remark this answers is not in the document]
alexdewar: Well as long as it's passing the CI checks it probably doesn't matter anyway.

projects/ardin_mb/plot.py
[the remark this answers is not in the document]
alexdewar: :+1:

projects/ardin_mb/state_handler.h
alexdewar: :laughing:

neworderofjamie:
Are you ok for me to merge?

alexdewar:
Sure

PR #187, opened 2020-05-13
Provide git hash(es) of project to C++

alexdewar:
A few times now I've generated image databases with bugs in them and it's often difficult to remember which version of the code you used to generate your data files, so I thought it probably makes sense to write the git commit into the database metadata file. (It probably also makes sense to do this elsewhere for "proper" datasets when we want to be able to repeat the processing that was done reliably.)

I've added separate macros for ``BOB_ROBOTICS_GIT_COMMIT`` and ``BOB_PROJECT_GIT_COMMIT``. The latter is for when you've got a project outside of the BoB robotics tree and want to be able to get at that too.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -26,7 +26,9 @@
             "defines": [
                 "ROBOT_TYPE_TANK",
                 "ROBOT_TYPE=Tank",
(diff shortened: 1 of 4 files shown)


alexdewar:
I suppose an alternative would be to put both hashes into one version string... Maybe that would be a bit more compact.

PR #189, opened 2020-05-18
AntWorld updates

alexdewar:
This PR introduces a couple of AntWorld things I've been working on:
* Using Mike's new antworld for the databases
* A python wrapper for a subset of the BoB robotics ant world functionality (requested by Andy)

Currently actually building the python code into a module is a bit of a mess on Windows (I think I'll just bung everything in a zipfile rather than trying to make a proper package), but it should work!

Any feedback welcome :-)


python/antworld/pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "wheel", "scikit-build", "cmake", "cv2", "numpy"]
```
neworderofjamie, line 2: why not just use setuptools? it's a bit shit but almost universal - I've never encountered scikit-build before...

include/antworld/surface.h
```diff
@@ -56,7 +56,7 @@ class Surface
         glEnableClientState(GL_VERTEX_ARRAY);
 
         // Calculate number of vertices from positions
-        m_NumVertices = positions.size() / size;
+        m_NumVertices = static_cast<unsigned int>(positions.size() / size);
```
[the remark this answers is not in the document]
alexdewar: Good point.

python/antworld/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+include(../../cmake/bob_robotics.cmake)
+BoB_project(SOURCES antworld.cc
+            PYTHON_MODULE antworld
+            BOB_MODULES antworld
+            EXTERNAL_LIBS numpy)
+
+# On gcc I'm getting this warning for static definitions of structs, but it
+# doesn't matter if the remaining fields are filled with zeros, so just suppress
+# it. -- AD
+if(GNU_TYPE_COMPILER)
```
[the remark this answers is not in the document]
alexdewar: A few of the static ones give this warning. The problem is that I want the structs to be const, but I also don't want to have to explicitly set all fields (because they're not all present in all versions of python).

I'm guessing with them being static the members will all just be zeroed anyway, plus examples I found online all seem to do this.

python/antworld/pyproject.toml
alexdewar: It's just like setuptools (I think) except it builds your CMake project for you beforehand, which is why I chose it.

alexdewar:
I only discovered pybind after I'd already written it :-(, I figured my choices were either Boost::Python or using Python's C API.

If we add anything to it we could change over to pybind, but I figure it's probably ok as is for now, just a tad verbose.

python/antworld/pyproject.toml
neworderofjamie: Ahh cool - that makes sense - setup tools does demand that it's in charge of building extensions

neworderofjamie:
Maybe make an issue for it? I want to switch GeNN from SWIG to pybind one day and this would be a good practice

alexdewar:
Sure! Pybind does look rather nice.

Other than that, are you happy to merge?

PR #190, opened 2020-05-20
Smarter copying of DLL files

neworderofjamie:
So there are two issues with the current strategy of building executables into the bin directory:
1. It's inconsistant with Linux and Mac
2. It breaks GeNN which deposits its dlls into the executable directory to work around the same problem

I've come up with a slightly hacky batch-file based solution which recursively calls [dumpbin](https://docs.microsoft.com/en-us/cpp/build/reference/dumpbin-reference) on the executable to find what dlls it needs and, if they're in VCPKG, copies them to the executable directory if they're newer/not present.

### DIFF
--- a/bin/copy_dependencies_vcpkg.bat
+++ b/bin/copy_dependencies_vcpkg.bat
@@ -1,0 +1,17 @@
+ECHO OFF
+SETLOCAL ENABLEDELAYEDEXPANSION
+
+REM Loop through dependencies of dll/executable, filtering out lines which contain dll names
+FOR /F %%d IN ('dumpbin /dependents %1 ^| FINDSTR /I /R "[a-zA-Z_-]*\.dll"') DO (
+    REM If dll is supplied by VCPKG (others are likely to be in system32)
+    SET "VCPKG_FILE=%2\bin\%%d"
+    IF EXIST !VCPKG_FILE! (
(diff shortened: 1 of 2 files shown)


alexdewar:
This looks really useful! I think it'll also help with packaging the python antworld module on Windows, which I was rather struggling with.

bin/copy_dependencies_vcpkg.bat
```diff
@@ -0,0 +1,15 @@
+ECHO OFF
```
alexdewar, line 1: My batch file skills are pretty rusty, but I have a feeling that this should maybe be ``@ECHO OFF`` so that the command itself isn't written to the terminal. Is that right...?

neworderofjamie:
Yeah, copying all of the results into the python module should help there too.

bin/copy_dependencies_vcpkg.bat
neworderofjamie: In this context (for whatever reason) that doesn't happen but I think you're correct. Batch files are the worst - especially all the hacks they've added over time (if you ever want to be horrified try and understand what ``SETLOCAL ENABLEDELAYEDEXPANSION`` does)

neworderofjamie:
Are you happy with this @alexdewar ?

alexdewar:
Yep, looks good.

PR #192, opened 2020-06-02
Fixed upside-down rendering of top-down view

neworderofjamie:
I meant to fix this in #185 but totally forgot to. The basic problem was an OpenGL one but there were additional issues with correctly rendering routes and vector fields **above** the ground. This also fixes issues with the ant world map creator and takes advantage of ``Vector2<>`` and ``Pose2<>`` for various bits of route API

### DIFF
--- a/src/antworld/renderer.cc
+++ b/src/antworld/renderer.cc
@@ -244,7 +244,7 @@ void Renderer::renderTopDownView(GLint viewportX, GLint viewportY, GLsizei viewp
     // Build modelview matrix so that the 'top' is up against the near clip plane
     glMatrixMode(GL_MODELVIEW);
     glLoadIdentity();
-    glTranslatef(0.0f, 0.0f, -maxBound[2].value());
+    glTranslatef(0.0f, 0.0f, -static_cast<GLfloat>(maxBound[2].value()));
 
     // Render geometry
     renderTopDownGeometry();
--- a/src/antworld/renderer.cc
+++ b/src/antworld/renderer.cc
@@ -232,18 +232,19 @@ void Renderer::renderTopDownView(GLint viewportX, GLint viewportY, GLsizei viewp
     const auto &minBound = getWorld().getMinBound();
     const auto &maxBound = getWorld().getMaxBound();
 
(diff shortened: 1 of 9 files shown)


src/antworld/renderer.cc
```diff
@@ -232,18 +232,19 @@ void Renderer::renderTopDownView(GLint viewportX, GLint viewportY, GLsizei viewp
...
             minBound[1].value(), maxBound[1].value(),
-            -1.0, maxBound[2].value() - minBound[2].value());
+            0.0, maxBound[2].value() - minBound[2].value());
 
-    // Build modelview matrix to centre world
+    // Build modelview matrix so that the 'top' is up against the near clip plane
     glMatrixMode(GL_MODELVIEW);
     glLoadIdentity();
-    glScalef(1.0f, 1.0f, -1.0f);
+    glTranslatef(0.0f, 0.0f, -maxBound[2].value());
```
alexdewar, line 247: This could use a ``static_cast<GLfloat>(...)`` for extra fussy compilers...

neworderofjamie: oh yes - missed one

src/antworld/route_continuous.cc
```diff
@@ -246,35 +246,35 @@ void RouteContinuous::render(meter_t antX, meter_t antY, degree_t antHeading) co
...
+bool RouteContinuous::atDestination(const Vector2<meter_t> &position, meter_t threshold) const
 {
     // If route's empty, there is no destination so return false
     if(m_Waypoints.empty()) {
         return false;
     }
     // Otherwise return true if
     else {
-        return distance(m_Waypoints.back(), x, y) < threshold;
+        return distance(m_Waypoints.back(), position.x(), position.y()) < threshold;
```
alexdewar, line 266: You could also do ``m_Waypoints.back().distance2D(position)`` instead.

neworderofjamie: Sadly that doesn't work as ``m_Waypoints`` are passed directly to OpenGL so are unitless ``GLfloat``s

PR #193, opened 2020-06-05
Mostly routes

neworderofjamie:
- Most important thing is a small fix to ``RouteContinuous`` so you can make repeated calls to ``load``. 
- Added a method to ``RouteContinuous`` to generate Levy walks (based on the wikipedia entry so, if you have any deeper knowledge, it'd be appreciated)
- Fixed a minor windows issue where the wrong library directory was used when making debug builds

Also, in #183, you introduced a fix for the duplicate path-less libraries being added to the Visual Studio project. Adding the library directory fixes this for release but these still need manually removing in debug as, whatever is adding them, doesn't take into account the debugness and adds the versions with the wrong names (which don't exist in the debug library directory). Not sure if that's a helpful clue as to the underlying issue or not...

### DIFF
--- a/src/antworld/route_ardin.cc
+++ b/src/antworld/route_ardin.cc
@@ -275,7 +275,7 @@ void RouteArdin::render(const Pose2<meter_t, degree_t> &pose, meter_t height) co
 
     glBindVertexArray(m_OverlayVAO);
 
-    glTranslatef(static_cast<GLfloat>(pose.x().value()), static_cast<GLfloat>(pose.y().value()), 0.0f);
+    glTranslatef(static_cast<GLfloat>(pose.x().value()), static_cast<GLfloat>(pose.y().value()), 0.01f);
     glRotatef(static_cast<GLfloat>(-pose.yaw().value()), 0.0f, 0.0f, 1.0f);
     glDrawArrays(GL_LINES, 0, 2);
     glPopMatrix();
--- a/src/antworld/route_continuous.cc
+++ b/src/antworld/route_continuous.cc
@@ -124,6 +124,7 @@ void RouteContinuous::load(const std::string &filename)
     // Clear existing waypoints and headings
     m_Waypoints.clear();
     m_Headings.clear();
+    m_CumulativeDistance.clear();
 
     // Open file for binary IO
     std::ifstream input(filename, std::ios::binary);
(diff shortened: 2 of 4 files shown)


src/antworld/route_ardin.cc
```diff
@@ -275,7 +275,7 @@ void RouteArdin::render(const Pose2<meter_t, degree_t> &pose, meter_t height) co
 
     glBindVertexArray(m_OverlayVAO);
 
-    glTranslatef(static_cast<GLfloat>(pose.x().value()), static_cast<GLfloat>(pose.y().value()), 0.0f);
+    glTranslatef(static_cast<GLfloat>(pose.x().value()), static_cast<GLfloat>(pose.y().value()), 0.01f);
```
alexdewar, line 278: I'm guessing you are hardcoding the height to 10cm here, right? Shouldn't z be set to ``height`` instead?

src/antworld/route_continuous.cc
```diff
@@ -235,18 +161,17 @@ void RouteContinuous::render(const Pose2<meter_t, degree_t> &pose, meter_t heigh
...
     if(m_RouteNumPoints > 0) {
         glBindVertexArray(m_RouteVAO);
-
         glDrawArrays(GL_LINE_STRIP, 0, m_RouteNumPoints);
     }
 
     glBindVertexArray(m_OverlayVAO);
 
-    glTranslatef(static_cast<GLfloat>(pose.x().value()), static_cast<GLfloat>(pose.y().value()), 0.0f);
+    glTranslatef(static_cast<GLfloat>(pose.x().value()), static_cast<GLfloat>(pose.y().value()), 0.01f);
```
alexdewar, line 174: See above

src/antworld/route_ardin.cc
neworderofjamie: So, (deprecated but convenient) OpenGL provides you with matrix stack onto which you can pop and push transforms. At the top of these functions we push the height z-transform before doing any (essentially 2D) drawing. This additional translate is just to make sure the arrow renders above the route

alexdewar: Ahhh I see. Makes sense. There's so much I don't know about OpenGL :laughing:

PR #195, opened 2020-09-15
Hash navigation code for Norbert

alexdewar:
Hey,

I'm just going to ring you about this, but I thought I'd post the code here so you can see what I'm talking about and in case you want to download my changes directly.


projects/PM_hash_navigation/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+cmake_minimum_required(VERSION 3.1)
+include(../../cmake/bob_robotics.cmake)
+BoB_project(SOURCES main.cc
+            EXTERNAL_LIBS opencv
+            BOB_MODULES common imgproc robots video)
```
alexdewar, line 5: I've ported the code to the new version of BoB robotics. *If* you can make this work for you then that'd probably be best because we've fixed a lot of bugs since the version you're using...

projects/PM_hash_navigation/main.cc
```diff
@@ -1,27 +1,29 @@
 //main
+#include "pm_control.h"
+#include "pm_hasher.h"
+#include "snapshot_db.h"
 
 // BoB robotics includes
-#include "common/logging.h"
+// #include "common/logging.h" // not needed in new BoB robotics
+#include "common/main.h"
```
alexdewar, line 8: These changes, like a few of the other ones, might not work with old BoB robotics

projects/PM_hash_navigation/main.cc
```diff
@@ -48,7 +50,7 @@ angleNormalise(int angle)
 }
 
 int
-main(int argc, char **argv)
+bobMain(int argc, char **argv)
```
alexdewar, line 53: We've started using ``bobMain`` functions instead of ``main`` ones, because that way you can make sure the robot stops moving if you get an error.

projects/PM_hash_navigation/main.cc
```diff
@@ -95,14 +97,14 @@ main(int argc, char **argv)
...
     unsigned int frame = 0;
 
+    cv::Mat rotatedView, resizedRotatedView, image, resized;
+    cv::Mat outputImage(unwrapRes, CV_8UC3);
     for (frame = 0;; frame++) {
         // Read from camera
-        if (!cam->readFrame(originalImage)) {
-            return EXIT_FAILURE;
-        }
+        cam->readFrameSync(originalImage);
```
alexdewar, line 104: When ``readFrame`` returns ``false`` that's because there's no frame available. It doesn't mean that there was an error (you get an exception in this case)

projects/PM_hash_navigation/main.cc
```diff
@@ -95,14 +97,14 @@ main(int argc, char **argv)
     // ----- camera loop ---------------
     unsigned int frame = 0;
 
+    cv::Mat rotatedView, resizedRotatedView, image, resized;
+    cv::Mat outputImage(unwrapRes, CV_8UC3);
```
alexdewar, line 101: Declare these outside the body of the loop. If you do it inside the loop then for every iteration your program will have to walk the heap to find space to allocate memory, allocate it and then walk it again to free the memory :wink:, so this could be part of the reason things are running slowly for you!

alexdewar (commented):
Have a look through the comments and see what you think.

projects/PM_hash_navigation/snapshot_db.h
```diff
@@ -146,13 +148,11 @@ class Snapshot_DB
...
-            cv::Mat out = currentView.clone();
-            cv::Mat currCopy = currentView.clone();
-            rollImage(currCopy, out, i);
+            rotatedViews.emplace_back(currentView.size(), currentView.type());
+            rollImage(currentView, rotatedViews.back(), i);
             unsigned long long int rotHash;
-            m_hasher.computeDCT_Hash(out, rotHash);
+            m_hasher.computeDCT_Hash(rotatedViews.back(), rotHash);
             rotationHashes.push_back(rotHash);
-            rotatedViews.push_back(out);
```
alexdewar, line 155: I've removed some more in-loop allocations here. You don't want to repeatedly make copies of ``currentView`` if you can possibly avoid it.

alexdewar: In an ideal world you don't really want to regenerate these ``std::vector``s every time the function is called either. ``rotatedViews`` could be populated once in the constructor and its contents could be reused and the other ``std::vector``s could also be members and then ``clear()``ed every time the function is called.

projects/PM_hash_navigation/pm_hasher.h
```diff
@@ -22,7 +25,7 @@ class PM_Hasher
         return count;
     }
 
-    void computeDCT_Hash(cv::Mat &in, unsigned long long &hash)
+    void computeDCT_Hash(const cv::Mat &in, unsigned long long &hash)
```
neworderofjamie, line 28: ``unsigned long long computeDCT_Hash(const cv::Mat &in)`` would be a more intuitive signature

projects/PM_hash_navigation/snapshot_db.h
neworderofjamie: I have clearly suceeded in my quest to make @alexdewar deeply afraid of heap allocations 🥳

alexdewar: :laughing: And not just that! I now have nightmares about const correctness :fearful:

neworderofjamie: and thus the C++ cargo cult continues :smile:

neworderofjamie:
how come this didn't get merged?

alexdewar:
I just put this up because it was an easier way of commenting on @tenxlenx 's code than over email. We'll resubmit it once he's using the latest version of BoB robotics and it's all finished.

PR #196, opened 2020-09-17
Image database code updates

alexdewar:
Hey!

I've made a couple of small changes to the image database code:
1.  Allow for loading a simple folder of images with ``ImageDatabase``. It just makes it easier to run simulations over image sets without any metadata (e.g. a single route recorded without positional data).
2. Let the user skip to a specific frame with ``image_database_browser`` by typing in a number and pressing g. Handy for when you're analysing image data and what to know what the world looked like at e.g. frame 100 or whatever.

I also tweaked the ``BOB_ASSERT`` macro so it's more like a proper function.

Let me know what you think :smiley:

### DIFF
--- a/src/navigation/image_database.cc
+++ b/src/navigation/image_database.cc
@@ -360,7 +360,7 @@ ImageDatabase::readDirectoryEntries()
         for (char &c : ext) {
             c = ::tolower(c);
         }
-        if (ext == "jpg" || ext == "png") {
+        if (ext == "jpg" || ext == "jpeg" || ext == "png") {
             // Save details to vector
             Entry entry{
                 { millimeter_t(NAN),
--- a/src/navigation/image_database.cc
+++ b/src/navigation/image_database.cc
@@ -1,7 +1,11 @@
 // BoB robotics includes
-#include "plog/Log.h"
-#include "imgproc/opencv_unwrap_360.h"
+#include "common/macros.h"
 #include "navigation/image_database.h"
+#include "imgproc/opencv_unwrap_360.h"
+
+// Third-party includes
+#include "plog/Log.h"
+#include "third_party/tinydir.h"
 
 // Standard C includes
 #include <ctime>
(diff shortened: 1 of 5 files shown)


tools/image_database_browser/image_database_browser.cc
```diff
@@ -26,35 +28,56 @@ int bobMain(int argc, char **argv)
...
         cv::imshow(argv[0], images[i]);
-        cv::setWindowTitle(argv[0], ss.str());
+        cv::setWindowTitle(argv[0], ssTitle.str());
+
+        size_t newi = i;
         do {
-            switch (cv::waitKeyEx(50) & OS::KeyMask) {
+            const auto key = cv::waitKeyEx(50) & OS::KeyMask;
+            if (key >= '0' && key <= '9') {
+                ssNumber << static_cast<char>(key);
```
neworderofjamie, line 50: What's the ``ssNumber`` logic doing? it's hard to tell

src/navigation/image_database.cc
```diff
@@ -284,8 +300,8 @@ ImageDatabase::ImageDatabase(const filesystem::path &databasePath, bool overwrit
         while (std::getline(lineStream, field, ',')) {
             // Trim whitespace from left (not C++ at its prettiest!)
             field.erase(field.begin(), std::find_if(field.begin(), field.end(), [](int ch) {
-                return !std::isspace(ch);
-            }));
+                            return !std::isspace(ch);
+                        }));
```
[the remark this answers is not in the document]
alexdewar: Hmm, yeah. It seems like ``clang-format`` is about as good at figuring out the right indentation for weird C++ constructions as the rest of us!

I just moved the lambda out of line here to make it more explicit.

src/navigation/image_database.cc
```diff
@@ -309,14 +325,59 @@ ImageDatabase::ImageDatabase(const filesystem::path &databasePath, bool overwrit
...
+        if (file.is_dir) {
+            continue;
+        }
+
+        const filesystem::path fileName = file.path;
+        auto ext = fileName.extension();
+        for (char &c : ext) {
+            c = ::tolower(c);
+        }
+        if (ext == "jpg" || ext == "png") {
```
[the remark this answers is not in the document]
alexdewar: :+1:

tools/image_database_browser/image_database_browser.cc
alexdewar: Hmm I didn't really comment this terribly well because I expected that I would probably be the only person using it :wink:, but I agree that the logic's a bit opaque here. The idea is that if you want to jump to e.g. frame 100, you can do that by entering: 1, 0, 0, g. ``ssNumber`` just keeps a record of the last number keys that were pressed.

I've put in a comment to make this more obvious.

neworderofjamie: That is definitely smarter than I assumed it was :smile:

PR #198, opened 2020-10-04
Parallelise perfect memory

alexdewar:
I've finally got round to parallelising the perfect memory code (partly because I got a shiny new Ryzen processor and wanted to try it out :happy:). It ended up just involving adding two OpenMP directives and then tinkering with CMake to get OpenMP to actually work (apparently it wasn't actually working with Eigen before).

It looks like Eigen is still running single threaded, even though OpenMP is enabled properly now, so I guess the Infomax code isn't run in parallel anyway. Anyhoo, that's a job for another day.

Single threaded results:
```
17:50:05.369 [bobMain@30] Testing with best-matching snapshot method...
17:50:05.369 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:50:06.330 [trainRoute@20] Loaded 812 snapshots
17:50:07.265 [bobMain@44] Heading: 0 rad
17:50:07.265 [bobMain@45] Best-matching snapshot: #10
17:50:07.265 [bobMain@46] Difference score: 0
Time taken for testing: 935.128
17:50:07.265 [bobMain@50] Testing with best-matching snapshot method with partial rotation...
17:50:07.265 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:50:08.223 [trainRoute@20] Loaded 812 snapshots
17:50:08.679 [bobMain@67] Heading: -0.0349066 rad
17:50:08.679 [bobMain@68] Best-matching snapshot: #11
17:50:08.679 [bobMain@69] Difference score: 0.0221338
Time taken for testing: 455.141
17:50:08.679 [bobMain@73] Testing with RMS image difference...
17:50:08.679 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:50:09.631 [trainRoute@20] Loaded 812 snapshots
17:50:10.654 [bobMain@85] Heading: 0 rad
17:50:10.654 [bobMain@86] Best-matching snapshot: #10
17:50:10.654 [bobMain@87] Difference score: 0
Time taken for testing: 1023.24
17:50:10.654 [bobMain@92] Testing with 3 weighted snapshots...
17:50:10.654 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:50:11.607 [trainRoute@20] Loaded 812 snapshots
17:50:12.543 [bobMain@103] Heading: -9.9519e-06 rad
17:50:12.543 [bobMain@105] Snapshot 1: #10 (0)
17:50:12.543 [bobMain@105] Snapshot 2: #11 (0.0221338)
17:50:12.543 [bobMain@105] Snapshot 3: #9 (0.022976)
Time taken for testing: 935.673
17:50:12.543 [bobMain@111] Testing with HOG...
17:50:12.543 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:50:12.543 [BoBRobotics::Navigation::PerfectMemoryStore::HOG<Differencer>::HOG@36] Creating perfect memory for 720 entry HOG features
17:50:13.671 [trainRoute@20] Loaded 812 snapshots
17:50:43.818 [bobMain@125] Heading: 0 rad
17:50:43.818 [bobMain@126] Best-matching snapshot: #10
17:50:43.818 [bobMain@127] Difference score: 3.41447e-13
Time taken for testing: 30147
```

Multi-threaded results:
```
17:52:25.782 [bobMain@30] Testing with best-matching snapshot method...
17:52:25.782 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:52:26.758 [trainRoute@20] Loaded 812 snapshots
17:52:27.052 [bobMain@44] Heading: 0.0349066 rad
17:52:27.052 [bobMain@45] Best-matching snapshot: #9
17:52:27.052 [bobMain@46] Difference score: 0.022976
Time taken for testing: 293.967
17:52:27.053 [bobMain@50] Testing with best-matching snapshot method with partial rotation...
17:52:27.053 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:52:28.011 [trainRoute@20] Loaded 812 snapshots
17:52:28.177 [bobMain@67] Heading: -0.10472 rad
17:52:28.177 [bobMain@68] Best-matching snapshot: #756
17:52:28.177 [bobMain@69] Difference score: 0.0343338
Time taken for testing: 166.37
17:52:28.178 [bobMain@73] Testing with RMS image difference...
17:52:28.178 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:52:29.136 [trainRoute@20] Loaded 812 snapshots
17:52:29.713 [bobMain@85] Heading: 0.139626 rad
17:52:29.713 [bobMain@86] Best-matching snapshot: #72
17:52:29.713 [bobMain@87] Difference score: 0.0975347
Time taken for testing: 576.563
17:52:29.713 [bobMain@92] Testing with 3 weighted snapshots...
17:52:29.713 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:52:30.676 [trainRoute@20] Loaded 812 snapshots
17:52:30.931 [bobMain@103] Heading: 0.10472 rad
17:52:30.931 [bobMain@105] Snapshot 1: #483 (0.0342453)
17:52:30.931 [bobMain@105] Snapshot 2: #615 (0.0342889)
17:52:30.931 [bobMain@105] Snapshot 3: #208 (0.0343198)
Time taken for testing: 254.292
17:52:30.931 [bobMain@111] Testing with HOG...
17:52:30.931 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

17:52:30.931 [BoBRobotics::Navigation::PerfectMemoryStore::HOG<Differencer>::HOG@36] Creating perfect memory for 720 entry HOG features
17:52:32.042 [trainRoute@20] Loaded 812 snapshots
17:52:36.927 [bobMain@125] Heading: 0 rad
17:52:36.927 [bobMain@126] Best-matching snapshot: #10
17:52:36.927 [bobMain@127] Difference score: 3.41447e-13
Time taken for testing: 4885.23
```


alexdewar:
And I've just realised that this totally won't work :rofl:. We've got some intermediate variables that'll be clobbered by the different threads.

I'll hack this a bit more tomorrow.

neworderofjamie:
No worries - let me know when you want a review! I can imagine there being issues with the various 'scratch' variables in the perfect memory...Making them all thread local storage would be the solution in C++11 but I'm not sure how well/if that plays with OpenMP.

alexdewar:
Right, so I think I'm finally happy with it :stuck_out_tongue:. I *think* I've now got all the right ``thread_local``s in the right places and I've added a unit test to check this (if I cherry-pick the commit onto the master branch it also passes).

By doing the parallelisation over the loop doing the rotation, it's considerably faster than my previous broken attempt and we also get a speed-up for the infomax code (and any future algorithms we implement...). I'm now a getting >6x speed-up, which makes sense with my 6-core machine. (I think it'll be slightly faster still with repeated calls to ``getHeading()`` as there won't be so many allocations for thread-local variables needed.)

I also got slightly carried away with the parallelisation and parallelised image loading in the ``ImageDatabase`` code while I was at it.

New benchmarks:
```
15:42:38.536 [bobMain@22] Loaded 812 snapshots
15:42:38.536 [bobMain@25] Testing with best-matching snapshot method...
15:42:38.536 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

15:42:38.690 [bobMain@39] Heading: 0 rad
15:42:38.690 [bobMain@40] Best-matching snapshot: #10
15:42:38.690 [bobMain@41] Difference score: 0
Time taken for testing: 151.544
15:42:38.690 [bobMain@45] Testing with best-matching snapshot method with partial rotation...
15:42:38.690 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

15:42:38.758 [bobMain@62] Heading: -0.0349066 rad
15:42:38.758 [bobMain@63] Best-matching snapshot: #11
15:42:38.758 [bobMain@64] Difference score: 0.0221338
Time taken for testing: 65.8388
15:42:38.758 [bobMain@68] Testing with RMS image difference...
15:42:38.758 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

15:42:38.930 [bobMain@80] Heading: 0 rad
15:42:38.930 [bobMain@81] Best-matching snapshot: #10
15:42:38.930 [bobMain@82] Difference score: 0
Time taken for testing: 169.689
15:42:38.930 [bobMain@87] Testing with 3 weighted snapshots...
15:42:38.930 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

15:42:39.089 [bobMain@98] Heading: -9.9519e-06 rad
15:42:39.089 [bobMain@100] Snapshot 1: #10 (0)
15:42:39.089 [bobMain@100] Snapshot 2: #11 (0.0221338)
15:42:39.089 [bobMain@100] Snapshot 3: #9 (0.022976)
Time taken for testing: 157.52
15:42:39.089 [bobMain@106] Testing with HOG...
15:42:39.089 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@26] Running on 12 threads

15:42:39.089 [BoBRobotics::Navigation::PerfectMemoryStore::HOG<Differencer>::HOG@36] Creating perfect memory for 720 entry HOG features
15:42:43.667 [bobMain@120] Heading: 0 rad
15:42:43.667 [bobMain@121] Best-matching snapshot: #10
15:42:43.667 [bobMain@122] Difference score: 0
Time taken for testing: 4394.73
```

include/common/omp.h
```diff
@@ -0,0 +1,50 @@
...
+//! Returns current thread ID or 0 if OpenMP is not in use
+size_t currentThread();
+
+//------------------------------------------------------------------------
+// BoBRobotics::OMP::Array
+//------------------------------------------------------------------------
+/*!
+ * \brief A wrapper providing thread-local storage for e.g. scratch data
+ */
+template<class T>
```
neworderofjamie, line 21: I don't quite see why this is necessary as you're using it on static members which can be marked thread local with 
```c++
#pragma omp threadprivate
```
(according to https://www.openmp.org/spec-html/5.0/openmpsu104.html)

include/navigation/insilico_rotater.h
```diff
@@ -51,36 +50,36 @@ struct InSilicoRotater
...
-            auto i = m_BeginRoll;
-            while (true) {
-                func(m_Image, m_MaskImage, distance(m_BeginRoll, i));
-
-                i += m_ScanStep;
-                if (i >= m_EndRoll) {
-                    break;
-                }
-
+            #pragma omp parallel for
```
neworderofjamie, line 66: I don't see how this loop can be parallelised - you're rolling the same image pixel by pixel so each iteration depends on the previous one.

include/navigation/perfect_memory.h
```diff
@@ -115,7 +103,29 @@ class PerfectMemory : public VisualNavigationBase
...
+        const size_t numSnapshots = getNumSnapshots();
+        BOB_ASSERT(numSnapshots > 0);
+
+        // Clear differences
+        thread_local std::vector<float> diffs;
+        diffs.reserve(numSnapshots);
+        diffs.clear();
+
+        // Loop through snapshots and caculate differences
+        for (size_t s = 0; s < numSnapshots; s++) {
```
neworderofjamie, line 123: shouldn't *this* be the loop you parallelise as the iterations are entirely independent

include/navigation/perfect_memory.h
```diff
@@ -169,46 +179,42 @@ class PerfectMemoryRotater : public PerfectMemory<Store>
         std::vector<float> minDifferences;
         minDifferences.reserve(numSnapshots);
         for (size_t i = 0; i < numSnapshots; i++) {
-            const auto elem = std::min_element(std::cbegin(m_RotatedDifferences[i]), std::cend(m_RotatedDifferences[i]));
-            bestColumns.push_back(std::distance(std::cbegin(m_RotatedDifferences[i]), elem));
-            minDifferences.push_back(*elem);
+            size_t minIndex;
```
neworderofjamie, line 182: likewise, this loop seems a better candidate for parallelism

include/navigation/perfect_memory_store_raw.h
```diff
@@ -71,7 +69,8 @@ class RawImage
     {
         // Calculate difference between image and stored image
         const int imSize = image.rows * image.cols;
-        auto diffIter = m_Differencer(image, m_Snapshots[snapshot], m_DiffScratchImage);
+        thread_local cv::Mat diffScratchImage(image.size(), image.type());
```
neworderofjamie, line 72: sadly, it seems (https://stackoverflow.com/a/21376829/1476754) that mixing C++11 ``thread_local`` and OpenMP isn't generally a good idea. Maybe you can use the ``threadprivate`` pragma instead?

neworderofjamie (commented):
My knowledge of OpenMP is very minimal (basically that it exists and uses pragmas) but there's a couple of things that don't _look_ right that might be worth checking - 6x speedup on a 6 core machine is very impressive nonetheless - I'd expect these algorithms to be super-memory-bound

neworderofjamie:
One other thing is that it might be good if the tests **didn't** use all your CPU cores - I remember that they had a habit of freezing my desktop already...

include/navigation/insilico_rotater.h
alexdewar: We might have done that at some point in the distant past, but we were using a scratch variable to store the rotated image here even before I started mangling things :smile: .

I actually did what you suggested initially (i.e. parallelising the inner loop which iterates over the different snapshots), but it doesn't perform as well, I guess a) because the rotations are still being done on a single thread and b) because whatever OpenMP synchronisation magic happens when the loop finishes would have to happen ``numRotations`` times. The other advantage to parallelising the outer loop is that any other algorithms using ``InSilicoRotater`` also get a speedup (Infomax now also runs much faster, but I didn't print the figures).

include/navigation/perfect_memory_store_raw.h
alexdewar: The reason I did this initially was because I didn't know that you could use the ``#pragma omp`` to do this in random places in your code away from the loop... but I just tried it and it seems to work! You're right that this seems a more sensible approach.

alexdewar:
> One other thing is that it might be good if the tests **didn't** use all your CPU cores - I remember that they had a habit of freezing my desktop already...

We're not actually running the tests on Jenkins (although we probably should be); we just test-build the examples folder. We do use all threads for this though, which is probably why your PC is freezing. We could either just use one make thread or we could do ``nice -n19 ...`` to lower the priority of the build process.

If you set the ``OMP_NUM_THREADS`` environment variable, you can reduce the number of threads being used.

include/navigation/insilico_rotater.h
neworderofjamie: Oh yeah, you're totally right - seems I've forgotten how this code works. Each iteration of this loop i.e. thread makes a rotated version of the image and tests it against all the snapshots. This is entirely where it should be parallelised.

neworderofjamie:
Hmm, that is strange - I had always assumed this was infomax churning away. GeNN also builds using all physical cores and doesn't seem to cause the same issue so very strange.

neworderofjamie:
Also, would you like any Windows-fixing assistance?

include/common/omp.h
alexdewar: Good point. I've dropped this code now that I better understand how to use ``threadprivate` :stuck_out_tongue:

alexdewar:
> Hmm, that is strange - I had always assumed this was infomax churning away. GeNN also builds using all physical cores and doesn't seem to cause the same issue so very strange.

Weird. Because GeNN takes a while to build, right? I'm not sure what the deal is tbh. I don't think building the BoB robotics examples should be much more CPU-demanding than GeNN. I'd be interested to know if putting a ``nice -n19 ...`` in front of our build command fixes it -- even if it's hogging the CPU it still shouldn't freeze up your desktop if it's running at a low priority.

Re Windows: It seems like MSVC only lets you parallelise for loops where your iterator variable is an ``int`` :worried: . I'm not sure if the MSVC peeps are just adhering to the standards more strictly or if it's a limitation (clang and gcc seem happy with it), but either way it would be annoying to fix up the ``rotate`` function because it uses iterators (and worse than that, the iterator type is templated!)... On second thoughts, maybe we should take a more granular approach and just wrap the ``#pragma``s for that function with a couple of ``#ifndef _MSC_VER``s. As it is I'm probably hobbling other bits of code by disabling OpenMP everywhere.

alexdewar:
> Re Windows: It seems like MSVC only lets you parallelise for loops where your iterator variable is an `int`  . I'm not sure if the MSVC peeps are just adhering to the standards more strictly or if it's a limitation (clang and gcc seem happy with it), but either way it would be annoying to fix up the `rotate` function because it uses iterators (and worse than that, the iterator type is templated!)... On second thoughts, maybe we should take a more granular approach and just wrap the `#pragma`s for that function with a couple of `#ifndef _MSC_VER`s. As it is I'm probably hobbling other bits of code by disabling OpenMP everywhere.

After messing around in Visual Studio a bit I think it's probably easiest if we just leave it disabled. Aside from having to make all the iterators ``int``s, MSVC also doesn't support dynamic initialisation of ``threadprivate`` variables, so I had to change e.g. ``static std::vector<float>`` to ``static std::unique_ptr<std::vector<float>>`` and then initialise as needed. And after all that it didn't work anyway (I think MSVC is doing something weird with my thread-local variables).

neworderofjamie:
Yep - that sounds like getting it to work on Windows is more effort than it's worth *and* would make the code significantly more gross in the process

alexdewar:
Is this one good to merge @neworderofjamie? I think I've sorted all the suggestions now.

include/navigation/insilico_rotater.h
```diff
@@ -49,31 +48,31 @@ struct InSilicoRotater
...
-                const auto index = toIndex(i);
-                rollImage(m_Image, scratchImage, index);
-                if (!scratchMask.empty()) {
-                    rollImage(m_MaskImage, scratchMask, index);
+            #pragma omp parallel
+            {
+                static cv::Mat scratchImage, scratchMask;
+                #pragma omp threadprivate(scratchImage, scratchMask)
+
+                scratchImage.create(m_Image.size(), CV_8UC1);
```
neworderofjamie, line 61: Isn't the whole purpose of the scratch images that they **don't** have to get re-allocated every time? This should be doable with function-scope statics but you'd have to replace the create calls with:
```
static cv::Mat scratchImage(m_Image.size(), CV_8UC1);
static cv::Mat scratchMask(m_MaskImage.size(), CV_8UC1);
```

neworderofjamie: Actually, as ``create`` is lazy, I think this is fine but it's probably worth putting a comment to that effect.

include/navigation/insilico_rotater.h
```diff
@@ -91,7 +90,7 @@ struct InSilicoRotater
         const size_t m_ScanStep;
         const IterType m_BeginRoll, m_EndRoll;
         const cv::Mat &m_Image, &m_MaskImage;
-        static OMP::Array<cv::Mat> m_ScratchImages, m_ScratchMaskImages;
+        static cv::Mat m_ScratchImages, m_ScratchMaskImages;
```
neworderofjamie, line 93: how come there's function scope and class scope statics here? As far as I can tell these aren't actually being initialised anywhere...

include/navigation/insilico_rotater.h
alexdewar: Yep. Plus this way if the size of ``m_Image`` changes between calls then ``scratchImage`` will still be the right size.

alexdewar: I've stuck a comment in now

include/navigation/insilico_rotater.h
alexdewar: Oops, forgot to take these out

PR #199, opened 2020-10-10
Fix build errors and warnings

alexdewar:
Some of the tools and projects weren't building any more. While I was at it, I also ran clazy over the code base to find a few other minor bugs.

### DIFF
--- a/examples/ant_world/ant_world.cc
+++ b/examples/ant_world/ant_world.cc
@@ -36,13 +36,13 @@ void handleGLError(GLenum, GLenum, GLuint, GLenum severity, GLsizei, const GLcha
 {
     if (severity == GL_DEBUG_SEVERITY_HIGH) {
         LOGE << message;
-    } 
+    }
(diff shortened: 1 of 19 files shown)


neworderofjamie (approved):
All changes look sensible to me - where did the mysterious ``window`` parameter come from originally? Was it required for whatever we replaced with SFML?

alexdewar:
> All changes look sensible to me - where did the mysterious `window` parameter come from originally? Was it required for whatever we replaced with SFML?

I've no idea! Maybe I just assumed we'd need a ref to the ``sf::Window`` at some point, then didn't notice we weren't using it :laughing: 

Cheers for looking at this! I just pushed a commit to this branch to also remove the ``AntWorldRotater``... not sure how you feel about that? I don't think we really need it and it was causing build failures on @tenxlenx 's machine so I thought it'd be easiest just to remove it.

alexdewar:
Good good!

PR #200, opened 2020-10-13
OpenMP+macOS fixes

alexdewar:
It turns out that even though the code was supposed to build when OpenMP wasn't found, that didn't work in practice so we were getting build errors on macOS. And even when libomp *was* installed with homebrew (it's needed for OpenMP support in clang) it wouldn't build anyway because CMake couldn't find it.

I've fixed these issues as well as a couple of other minor things. @tenxlenx can you please confirm that this fixes things for you? (Just check out this branch and try building e.g. the ``examples/perfect_memory`` project.)


src/navigation/read_objects.cc
```diff
@@ -23,9 +23,9 @@ readObjects(const filesystem::path &objectFilePath)
 
     cv::FileStorage fs(objectFilePath.str(), cv::FileStorage::READ);
     std::vector<double> vertex(2);
-    for (const auto &objectNode : fs["objects"]) {
+    for (const auto objectNode : fs["objects"]) {
```
neworderofjamie, line 26: is this a bad merge or did it upset clang somehow?

cmake/bob_robotics.cmake
```diff
@@ -302,11 +302,23 @@ macro(always_included_packages)
...
+            set(ENABLE_OPENMP TRUE)
+        endif()
+    endif()
+
     # Annoyingly, these packages export a target rather than simply variables
     # with the include path and link flags and it seems that this target isn't
     # "passed up" by add_subdirectory(), so we always include these packages on
     # the off-chance we need them.
-    if(NOT WIN32 AND NOT TARGET OpenMP::OpenMP_CXX)
+    if(ENABLE_OPENMP)
```
neworderofjamie, line 318: Does the above comment belong below this? I don't think the annoyance referred to is the same one you're fixing here...

src/navigation/read_objects.cc
alexdewar: Clang helpfully pointed out that ``fs["objects"]`` doesn't actually return a reference so it was actually correct before I changed it in the last PR :stuck_out_tongue:

cmake/bob_robotics.cmake
alexdewar: Good point. I've fixed this now.

alexdewar:
@tenxlenx Would you mind checking that this branch fixes the OpenMP issues you were having on your Macbook when you get a mo? Just check out the branch, build ``examples/perfect_memory`` and it should tell you that it's using multiple threads when you run it. (Maybe something to do while your Jetson is busy flashing.... :wink:)

alexdewar:
I'll just merge this for now as it's a pretty harmless changeset (and I'd like to have the option to turn OpenMP on/off elsewhere). Plus I think @tenxlenx is on Linux again for the next while...

PR #201, opened 2020-10-14
dct hash with an example, not sure if imgproc is the best place for the header though

tenxlenx:
Hi,

Just wanted to get some review for the dct hash code to merge.


include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,82 @@
...
+    
+    
+    //! computes the DCT hash 
+    static unsigned long long computeHash(const cv::Mat &in) {
+        
+		unsigned long long hash;
+        cv::Mat dct_mat;
+        cv::dct(in, dct_mat);
+        cv::Rect roi( 0, 0, 8, 8); 
+        cv::Mat low_dct( dct_mat, roi );
```
alexdewar, line 37: How big will ``dct_mat`` be? Why are we only using part of the image?

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,82 @@
...
+
+		std::vector<float> vec;
+		for (int row = 0; row < 9; row++) {
+			for (int col = 0; col < 9; col++) {
+			vec.push_back(low_dct.at<float>(row,col));
+			}
+		}
+
+		// calculate median value 
+		std::sort(std::begin(vec), std::end(vec));
```
alexdewar, line 47: Glad to see you went with the STL algo in the end :wink:. As I mentioned, you could also probably use ``std::partial_sort`` for a slight speedup, but that's not terribly important...

You could also drop ``vec`` and just operate on ``low_dct`` directly:
```c++
std::sort(low_dct.begin<float>(), low_dct.end<float>());
```
etc. I know ``vec`` is tiny but we want to avoid unnecessary heap allocations!

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,82 @@
...
+
+		std::vector<int> hashvector;
+		row.row(0).copyTo(hashvector);
+		std::stringstream ss;
+		std::string binstring;
+		for (int v : hashvector) { ss << v; }
+		ss >> binstring;
+		unsigned long long int binary = std::stoull(binstring, 0, 2);
+		hash = binary;
+		return hash;
```
alexdewar, line 68: This is a fairly roundabout way of achieving this...

How about:
```c++
unsigned long long hash = 0;
for (size_t i = 0; i < 64; i++) {
    if(reinterpret_cast<int *>(mask.data)[i]) {
        hash |= 1 << i;
    }
}
```
(I'm assuming that ``mask`` is an array of ``int``s.)

alexdewar (commented):
I've made some suggestions but on the whole it looks really good :smile:. Once you've tweaked things a bit more we should be able to drop most of the heap allocations in ``computeHash()`` which should speed things up a lot.

And I think ``ImgProc`` is probably the right place for this too.

examples/dct_hash/dct_hash.cc
```diff
@@ -0,0 +1,29 @@
...
+	cv::imshow("Display window1", img1);
+	cv::imshow("Display window2", img2);
+
+	auto hash1 = DCTHash::computeHash(img1);
+	auto hash2 = DCTHash::computeHash(img2);
+
+	int distance = DCTHash::distance(hash1,hash2);
+	std::cout << "the distance between the 2 images is " << distance << std::endl;
+   
+	int k = cv::waitKey(0); // Wait for a keystroke in the window
```
alexdewar, line 26: You don't need variable ``k`` here. It's causing a warning which I think is why the CI is currently failing.

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,82 @@
...
+    //! computes the DCT hash 
+    static unsigned long long computeHash(const cv::Mat &in) {
+        
+		unsigned long long hash;
+        cv::Mat dct_mat;
+        cv::dct(in, dct_mat);
+        cv::Rect roi( 0, 0, 8, 8); 
+        cv::Mat low_dct( dct_mat, roi ); 
+
+		std::vector<float> vec;
```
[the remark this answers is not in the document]
neworderofjamie: Also, ``low_dct`` is only 8x8 so you're going out of bounds in your loop

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,82 @@
...
+		std::vector<int> hashvector;
+		row.row(0).copyTo(hashvector);
+		std::stringstream ss;
+		std::string binstring;
+		for (int v : hashvector) { ss << v; }
+		ss >> binstring;
+		unsigned long long int binary = std::stoull(binstring, 0, 2);
+		hash = binary;
+		return hash;
+
```
neworderofjamie, line 69: This seems overly convoluted - you have a 8x8 mask image with 0 or 255 in each 8-bit pixel and you want to turn that into a 64-bit string you can just do something like:
```c++
const cv::Mat row = mask.reshape(1,1)
uint64_t binary = 0;
for(int i = 0; i < 64; i++) {
    if(row.at<uint8_t>(i) > 0) {
        binary |= (1 << i);
    }
}
return binary;
```

neworderofjamie (changes requested):
There's definitely some scope for optimising this a bit more and at least one buffer overrun issue.

include/imgproc/dct_hash.h
alexdewar: I've just realised we don't actually need ``mask`` at all here:
```c++
unsigned long long hash = 0;
for (size_t i = 0; i < 64; i++) {
    if(reinterpret_cast<float *>(low_dct.data)[i] > medianv) {
        hash |= 1 << i;
    }
}
```

neworderofjamie: Excellent point

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,82 @@
...
+class DCTHash {
+
+    private:
+    static unsigned int countSetBits(unsigned long long int n) { 
+        unsigned int count = 0; 
+        while (n) { 
+            count += n & 1; 
+            n >>= 1; 
+        } 
+        return count;
```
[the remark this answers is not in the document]
alexdewar: Good idea. I suppose an alternative would be to use ``std::bitset<64>`` instead of ``uint64_t`` for the hashes, then you've got the ``.count()`` method etc.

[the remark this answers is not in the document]
neworderofjamie: That is an even better idea - much more C++y, _probably_ implements the optimisations I suggest and, if you do want to extract a ``uint64_t``, you can call ``to_ullong()``

include/imgproc/dct_hash.h
neworderofjamie: Well, I think the idea is to throw away the high/low frequency components.

alexdewar: Ahh I see. Makes sense.

include/imgproc/dct_hash.h
tenxlenx: good point, the current implementation is just a hack with the strings and stuff

examples/dct_hash/dct_hash.cc
tenxlenx: I was going to do something with it, but then I deleted and forgot about this one :D

include/imgproc/dct_hash.h
alexdewar: Ignore this, I just noticed you're reusing ``vec``...

examples/dct_hash/dct_hash.cc
alexdewar: This still needs doing

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,82 @@
...
+#include <algorithm>
+
+// Standard C includes
+#include <cmath>
+
+//opencv
+#include <opencv2/opencv.hpp>
+
+
+class DCTHash {
```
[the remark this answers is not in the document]
alexdewar: This would be tidier

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,82 @@
...
+    } 
+
+    public:
+    
+    
+    //! computes the DCT hash 
+    static unsigned long long computeHash(const cv::Mat &in) {
+        
+		unsigned long long hash;
+        cv::Mat dct_mat;
```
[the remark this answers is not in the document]
alexdewar: Just right click and choose format document in vscode...

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,59 @@
...
+
+		// calculate median value
+		std::sort(std::begin(vec), std::end(vec));
+		float medianv;
+		const int n = 64;
+		if (n % 2 != 0) {
+			medianv = vec[n/2];
+		} else {
+			medianv = (vec[(n-1)/2] + vec[n/2]) / 2;
+		}
```
alexdewar, line 40: We *know* ``n`` is even here, so you don't need the if statement

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,59 @@
...
+        for (size_t i = 0; i < 64; i++) {
+            if(reinterpret_cast<float *>(low_dct.data)[i] > medianv) {
+                binary.set(i, 1);
+            }
+        }
+		return binary;
+    }
+
+    //! computes the hamming distance between 2 image hashes
+    static int distance(const std::bitset<64> &hash1, const std::bitset<64> &hash2) {
```
alexdewar, line 52: These probably don't need to be references because they're basically going to be plain old 64-bit ints under the hood

examples/dct_hash/dct_hash.cc
alexdewar: This might be what's breaking the CI... If this doesn't work, then click on the red cross by the commit and go through the logs to see what's up

examples/dct_hash/dct_hash.cc
```diff
@@ -0,0 +1,29 @@
...
+#include <opencv2/core.hpp>
+#include <opencv2/imgcodecs.hpp>
+#include <opencv2/highgui.hpp>
+
+int bobMain(int, char **)
+{
+	std::string image_path1 = "library1.jpg";
+	std::string image_path2 = "library2.jpg";
+    cv::Mat img1 = imread(image_path1, cv::IMREAD_GRAYSCALE);
+	cv::Mat img2 = imread(image_path2, cv::IMREAD_GRAYSCALE);
```
alexdewar, line 11: Another tabs/spaces issue.

VSCode can automatically sort this out. If you install the editorconfig extension then your whitespace will be sorted out when you save a file.

alexdewar (approved):
This is looking much better! There are still a few small changes I'd like to see though.

alexdewar:
D'oh! I meant to click "Request changes" not "Approve". So consider my approval revoked :wink:

neworderofjamie (changes requested):
Everything @alexdewar said basically 😄 this is looking _way_ nicer though - well done

alexdewar:
We should also write automated tests for this for completeness. @tenxlenx Let me know when you've done everything else and I can help you with that.

alexdewar (approved):
Looks good! There are a couple of small changes that I'll do now:
* Fix whitespace (there's still a weird combination of tabs/spaces going on)
* Remove ``static`` attributes of functions -- we don't need them
* Write some simple unit tests

If you haven't already, would you mind installing this vs code extension: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

(I put it as a "recommendation" for the repo, so it should be listed when you go to the extensions tab.) That *should* automatically fix future whitespace problems.

alexdewar:
Something else I missed: This should be in the ``BoBRobotics::ImgProc`` namespace. I'll do that too.

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,43 @@
...
+{
+    //! computes the DCT hash
+    inline static std::bitset<64> computeHash(const cv::Mat &in) {
+
+        cv::Mat dct_mat;
+        cv::dct(in, dct_mat);
+        cv::Mat rect(dct_mat,{ 0, 0, 8, 8});                // we only need the low 8x8 frequencies
+        std::sort(rect.begin<float>(),rect.end<float>());   // sorting to get median value
+        const float median= rect.at<float>(-1+rect.size().height/2,rect.size().width-1); 
+        rect = {dct_mat,{ 0, 0, 8, 8}};                     // re-assigning so we get the correct order back
```
alexdewar, line 24: I've only just noticed this, but this is a major bug! You *won't* get the correct order back if you do this. When you change ``rect`` you also change ``dct_mat``: OpenCV does not make a copy of the matrix.

include/imgproc/dct_hash.h
```diff
@@ -0,0 +1,43 @@
...
+namespace DCTHash
+{
+    //! computes the DCT hash
+    inline static std::bitset<64> computeHash(const cv::Mat &in) {
+
+        cv::Mat dct_mat;
+        cv::dct(in, dct_mat);
+        cv::Mat rect(dct_mat,{ 0, 0, 8, 8});                // we only need the low 8x8 frequencies
+        std::sort(rect.begin<float>(),rect.end<float>());   // sorting to get median value
+        const float median= rect.at<float>(-1+rect.size().height/2,rect.size().width-1);
```
alexdewar, line 23: The median is calculated wrongly here. There are 64 values here, so the median will be (element[31] + element[32]) / 2 (with zero-based indexing)

alexdewar (changes requested):
Those two bugs I spotted need fixing.

alexdewar:
@tenxlenx I did fix up the whitespace. Do a ``git pull`` to get my changes before you do anything.

include/imgproc/dct_hash.h
alexdewar: Instead of sorting ``rect`` in place, I'd suggest this:
```c++
std::array<float, 33> sorted;
std::partial_sort_copy(rect.begin<float>(), rect.end<float>(), sorted.begin(), sorted.end());
```

``sorted`` is declared on the stack so you don't have unnecessary heap allocations.

tests/dct.cc
```diff
@@ -0,0 +1,32 @@
...
+    EXPECT_EQ(dct(TestImages[1]), 17858815827266642603ULL);
+    EXPECT_EQ(dct(TestImages[2]), 1975389874025492227ULL);
+}
+
+TEST(DCT, distance)
+{
+    EXPECT_EQ(distance(0, 0), 0);
+    EXPECT_EQ(distance(0, 1), 1);
+    EXPECT_EQ(distance(std::numeric_limits<uint64_t>::max(), 0), 64);
+}
```
alexdewar, line 32: @tenxlenx This is how you write tests for GTest.

PR #203, opened 2020-10-26
Add Ofstad arena and misc fixes

alexdewar:
So this is half PR, half support request :wink:. I've converted the Ofstad arena (for @jembennettuk) to a .obj file so we can use it with the ant world code:
ofstad2

Hopefully it'll be handy as a small test environment apart from anything else.

The only issue is that @jembennettuk 's student is using a Mac and I'm struggling to get the ant world code working on it... I *think* the issue is that glew is not initialising properly (despite the call to ``glewInit()`` seemingly succeeding). A bunch of OpenGL functions seem to just be null pointers, suggesting they haven't been loaded properly. Any thoughts? I thought I'd ask you as our resident OpenGL expert :smile: 

@jembennettuk It is working on Linux at the moment though, so if you'd still like me to build things for you and talk you through how to use it, I'm happy to do that. I suppose you could always get Leo to run things in an Ubuntu VM, but that would maybe be a bit of a pain for him!


neworderofjamie:
So I think the answer is that glew isn't required on mac os. At least according to https://stackoverflow.com/a/11213354

alexdewar:
Ahh that makes sense!

Unfortunately that was only part of the problem though :cry:. Basically OpenGL support is horrid on macOS (it's actually deprecated so they're going to drop it at some point) and you can only use OpenGL v2 with the SFML graphics module: https://www.sfml-dev.org/tutorials/2.5/window-opengl.php#creating-an-opengl-window

We do use some OpenGL v3 features, so I think that basically there's no easy way to get this working on macOS. Sorry, @jembennettuk !

neworderofjamie:
God I hate Mac OS

tnowotny:
but I think we bought a non-mac laptop for @jembennettuk?

alexdewar:
@tnowotny It's more that his student who's currently using my old Ofstad matlab code is on macOS.

tnowotny:
oh ... I see ... but then the student still has a Sussex login I believe, so before you guys spend tons of time on this, we could have him work remotely from one of our linux boxes?

alexdewar:
Or he could run Ubuntu in a VM I guess...

alexdewar:
So I think this PR is done then, whenever you want to take a look at it @neworderofjamie :smiley:

cmake/bob_robotics.cmake
```diff
@@ -486,11 +486,6 @@ macro(BoB_build)
     # Link threading lib
     BoB_add_link_libraries(${CMAKE_THREAD_LIBS_INIT})
 
-    # Clang needs to be linked against libm and libstdc++ explicitly
-    if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
-        BoB_add_link_libraries(m stdc++)
-    endif()
-
```
neworderofjamie, line 493: I feel we do this with GeNN too - is it not required any more?

neworderofjamie (approved):

cmake/bob_robotics.cmake
alexdewar: It doesn't seem to be. Perhaps it's been fixed in newer versions of clang? I've tested with clang on macOS and Linux and it seems fine without this.

This also forces clang to use GNU's libstdc++ rather than LLVM's libc++, which could cause problems if you don't have libstdc++. macOS was telling me that libstdc++ is deprecated so it might not work in the future.

PR #205, opened 2020-10-30
fixed  bug where reading data was skipped, removed velicity as it mad…

tenxlenx:
-fixed a bug where nmea_parser said it can't parse the sentence, when it was a perfectly normal data
- removed velocity from readings as it was quite unreliable and seemed like it wasn't always available and made the gps unstable


alexdewar:
Let's fix these issues on #210 instead.

Issue #204, opened 2020-10-28
Build issues with CMake < 3.13


alexdewar:
This affects Ubuntu 18.04 and older. Annoyingly this includes the current version of the NVIDIA JetPack for Jetsons. On macOS you can get a newer version of CMake from homebrew, but you have to make sure that it is on your path, otherwise you may be using an outdated version.

Issues:
- The OpenCV cmake package forces the C++ standard to be C++11 with older versions of CMake, which breaks all the C++14 code we have
- At some point I apparently broke the linking process for older versions of CMake, so you get linking errors with e.g. OpenCV

Workaround:
- Install a newer version of CMake

You can either build from source or use Kitware's APT repository: https://apt.kitware.com/
(though I doubt they provide builds for ARM64)

neworderofjamie:
When did this break? I was building BoB robotics on a Jetson with one of the Ubuntu 16.04 Jetpacks the other day? Is this maybe only an issue with newer OpenCVs (I suspect this had 3.something)

alexdewar:
I've done some more probing and I think it might be. Maybe I'll submit an issue upstream if I can figure out exactly what the problem is.

neworderofjamie:
Found another delightful cmake issue on Ubuntu 16.0.4:
```cmake
if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.13.0")
```
The ``VERSION_GREATER_EQUAL`` operator isn't supported by cmake < 3.7.2 so maybe this needs turning into a VERSION GREATER "3.12.9" or however cmake versions work...

PR #206, opened 2020-11-02
Add support for extra fields in ImageDatabase

alexdewar:
This just adds support for having extra columns in the CSV file for the ``ImageDatabase`` class, e.g. for if we want to save readings for extra sensors. The problem at the moment is that if you have e.g. an IMU you'd have to save the readings to a separate file.

I've also stuck in examples for using the ``ImageDatabase`` class because I feel they may be useful.

Hopefully this will be useful for @tenxlenx and @efkag .


examples/image_database/write_example.cc
```diff
@@ -0,0 +1,59 @@
...
+#include "third_party/units.h"
+
+using namespace BoBRobotics;
+using namespace units::length;
+using namespace units::literals;
+
+int bobMain(int, char **argv)
+{
+    // Make a new image database; make folder with random name in current folder
+    const auto programPath = filesystem::path{ argv[0] }.parent_path();
```
neworderofjamie, line 16: This is more an FYI than anything I think should be fixed here but,  ``argv[0]`` is not very good for this - it's never an absolute path. Sadly the solutions are messy - https://stackoverflow.com/questions/933850/how-do-i-find-the-location-of-the-executable-in-c - different for mac, linux and windows - yey

include/navigation/image_database.h
```diff
@@ -73,78 +79,245 @@ class ImageDatabase
     //! The metadata for an entry in an ImageDatabase
     struct Entry
     {
-        Vector3<millimeter_t> position;
-        degree_t heading;
+        Vector3<millimeter_t> position = Vector3<millimeter_t>::nan();
+        degree_t heading{ NAN };
         filesystem::path path;
         std::array<size_t, 3> gridPosition; //! For grid-type databases, indicates the x,y,z grid position
+        std::unordered_map<std::string, std::string> extraFields;
```
neworderofjamie, line 86: I'm in two minds on this one but, especially as the ``record`` API is based on adding elements in order, it might be a lot simpler to just make this a nested class and have:

```c++
std::array<std::string, NumExtraFields> extraFields;
```
Then, if you wanted to,  you could always make an application-side enum or whatever.

include/navigation/image_database.h
```diff
@@ -73,78 +79,245 @@ class ImageDatabase
...
+            m_ImageDatabase.writeImage(filename, image);
+
+            Entry newEntry{
+                position,
+                heading,
+                m_ImageDatabase.m_Path / filename,
+                gridPosition,
+                {}
+            };
+            m_NewEntries.emplace_back(std::move(newEntry));
```
neworderofjamie, line 195: Does this not work here:
```c++
m_NewEntries.emplace_back(
    position,
    heading,
    m_ImageDatabase.m_Path / filename,
    gridPosition,
    {});
```
?
Otherwise I think you should ``push_back`` here. ``newEntry`` is on the stack so it can't be 'moved' in any useful way to a heap data structure. I'm a little surprised you don't get a warning here (that may only be a clang-level OCD thing)

include/navigation/image_database.h
```diff
@@ -73,78 +79,245 @@ class ImageDatabase
...
+            setExtraFields(std::forward<Ts>(extraFieldValues)...);
+        }
+
+        template<class... Ts>
+        void setExtraFields(std::string value, Ts&&... otherValues)
+        {
+            const auto &key = *(m_ExtraFieldNames.cend() - 1 - sizeof...(otherValues));
```
neworderofjamie, line 202: This is cool 😃

include/navigation/image_database.h
```diff
@@ -73,78 +79,245 @@ class ImageDatabase
...
+            Vector3<millimeter_t> position;
+            for (size_t i = 0; i < position.size(); i++) {
+                position[i] = (m_Separation[i] * gridPosition[i]) + m_Begin[i];
+            }
+            return position;
+        }
 
         //! Get a vector of all possible positions for this grid
-        std::vector<Vector3<millimeter_t>> getPositions();
+        std::vector<Vector3<millimeter_t>> getPositions()
```
neworderofjamie, line 263: const?

tests/test_headers/string.h
```diff
@@ -0,0 +1,19 @@
+#include "common.h"
```
neworderofjamie, line 1: Not 100% sure what your reason for having this in a header way but, due to the horror/magic within gtest, it can find tests across multiple executable units so there's no harm having tests in a bunch of cc files (this is how GeNN's unit tests are structured)

examples/image_database/write_example.cc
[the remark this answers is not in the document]
alexdewar: Heh, TIL. C++ never ceases to surprise with its weird edge cases :thinking:.

I might add a little helper function seeing as we do this fairly often...

include/navigation/image_database.h
alexdewar: This did occur to me after I'd submitted the PR... I think it's probably the more sensible way of doing it tbh. You could still have the helper function to get the value from its string name.

We probably shouldn't rely on fields being in a particular order though. I've deliberately made the CSV parsing code agnostic to column order now because there were a few hacks that made assumptions about the data based on number of columns and it's not v extensible that way.

include/navigation/image_database.h
alexdewar: > Does this not work here:
> 
> ```c++
> m_NewEntries.emplace_back(
>     position,
>     heading,
>     m_ImageDatabase.m_Path / filename,
>     gridPosition,
>     {});
> ```
> 
> ?
For some reason this didn't work for me, though I feel it should work too.

> Otherwise I think you should `push_back` here. `newEntry` is on the stack so it can't be 'moved' in any useful way to a heap data structure. I'm a little surprised you don't get a warning here (that may only be a clang-level OCD thing)
I *think* ``Entry`` will have a default move ctor, which just moves its moveable members (i.e. the filepath and the ``std::unordered_map``) and copies the others. So it might avoid a copy for the filepath this way.

include/navigation/image_database.h
alexdewar: Template metaprogramming FTW :+1:

examples/image_database/write_example.cc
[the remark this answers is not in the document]
neworderofjamie: Yeah, doing it right is annoying and it is helpful so that might be a good idea. Give me a shout if you want any windows testing

include/navigation/image_database.h
alexdewar: Good spot!

include/navigation/image_database.h
neworderofjamie: Ahh maybe you need to use ``std::piecewise_construct`` and ``std::forward_as_tuple`` to deal with the {}s

tests/test_headers/string.h
alexdewar: My reason for having it in a header is not really understanding gtest properly :wink:. It's cool that you can have multiple cc files though... I'll tidy this up then.

include/navigation/image_database.h
alexdewar: That sounds sensible... but I can't seem to get it to work here :thinking:. I've tried various permutations of brackets/not-brackets etc. but the compiler keeps complaining. Any thoughts?

neworderofjamie: Ahh not worth it then!

include/navigation/image_database.h
alexdewar: ...although after fiddling with the code I now realise we'd need ``Entry``s to hold a reference to the ``ImageDatabase`` and then use that to figure out which field corresponds to a given key. So it's probably cleaner as is.

examples/image_database/write_example.cc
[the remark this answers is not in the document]
alexdewar: I've pushed it to this branch (it's in ``src/common/path.cc`` if you want to take a look). Tested on Linux + Windows + macOS.

And with that, I think we're ready to merge!

alexdewar:
I think this PR and #204 are ready to merge, if you agree @neworderofjamie ?

include/common/path.h
```diff
@@ -5,6 +5,14 @@
 
 namespace BoBRobotics {
 namespace Path {
+//! Get the path of the folder containing the currently running program
+filesystem::path
+getProgramFolder();
```
neworderofjamie, line 10: ``getProgramDirectory`` might be a better name - folder always feels a bit 90s

src/common/path.cc
```diff
@@ -11,6 +22,35 @@
...
+filesystem::path
+getProgramPath()
+{
+#ifdef __linux__
+    char path[PATH_MAX + 1];
+    ssize_t len = readlink("/proc/self/exe", path, PATH_MAX);
+    BOB_ASSERT(len >= 0);
+    path[len] = '\0';
+#endif
+#ifdef _WIN32
```
neworderofjamie, line 40: maybe an 
```c++
#ifdef __linux__
    // linux
#elif defined(_WIN32)
    // windows
#elif defined(__APPLE__)
    // mac
#else
    #error "Unsupported platform"
#endif
```
might be good

src/navigation/image_database.cc
```diff
@@ -452,6 +455,21 @@ ImageDatabase::getFilename(const Vector3<millimeter_t> &position,
     return ss.str();
 }
 
+bool
+ImageDatabase::fileNameCompare(const std::string &fn1, const std::string &fn2)
+{
+    std::regex regex{ R"(^([^0-9]*)([0-9]+)\.(.+)$)" };
```
neworderofjamie, line 461: some comments here would be nice - I find nothing is quite as confusing as last month's regex 😃

tests/common.h
```diff
@@ -6,11 +6,6 @@
 // Google Test
 #include "gtest/gtest.h"
 
-// Standard C++ includes
-#include <string>
-#include <utility>
-
-// Put these using declarations here so we don't have to bother putting it in the test headers
 namespace BoBRobotics {}
 using namespace BoBRobotics;
```
neworderofjamie, line 10: using namespace in a header is pretty nasty....

tests/tests.cc
```diff
@@ -1,7 +1,5 @@
-#include "test_headers/common.h"
-
-// Automatically generated includes for the files in test_headers
-#include ".include_test_headers.h"
+// Google Test
+#include "gtest/gtest.h"
 
 int bobMain(int argc, char **argv)
```
neworderofjamie, line 4: you don't **generally** need this with gtest, you can just link in gtest_main.cc

include/common/path.h
alexdewar: True... I must have been feeling nostalglic for Windows 95 for a moment.

src/common/path.cc
alexdewar: Hmm... Yeah it is a bit clunky how it is.

src/navigation/image_database.cc
alexdewar: Good point! I also swapped around the logic of the function to make it clearer.

tests/common.h
alexdewar: Good point. I've dropped it now.

tests/tests.cc
alexdewar: Ah TIL! I think we will need it though as we already have a definition of ``main()`` that would conflict with the one in ``gtest_main.cc``.

neworderofjamie: oh yeah, I guess main() is defined in the common module

include/common/path.h
neworderofjamie: Why do you think XFCE is my favourite window manager....

src/navigation/image_database.cc
```diff
@@ -458,16 +458,23 @@ ImageDatabase::getFilename(const Vector3<millimeter_t> &position,
...
+        && m1[1] == m2[1] && m1[3] == m2[3]) {
+
+        // ...then use the numbers for comparison
+        return std::stoul(m1[2]) < std::stoul(m2[2]);
     }
 
-    // Compare using the number in the file name
-    return std::stoul(m1[2]) < std::stoul(m2[2]);
+    // ...else fall back on alphabetic comparison
+    return fn1 < fn2;
```
neworderofjamie, line 477: string comparison is a much better fallback behaviour 👍

src/navigation/image_database.cc
```diff
@@ -458,16 +458,23 @@ ImageDatabase::getFilename(const Vector3<millimeter_t> &position,
 bool
 ImageDatabase::fileNameCompare(const std::string &fn1, const std::string &fn2)
 {
+    // Find file names like (string)(number)(extension), e.g. "image45.png"
```
neworderofjamie, line 461: What's the purpose of the group around everything? Can't that be just be accessed with ``m[0]``?

neworderofjamie (approved):

src/navigation/image_database.cc
alexdewar: Those brackets aren't part of the regex -- they show where the raw string begins and ends. It's a weird syntax.

neworderofjamie: Oh yeah - **those** brackets. I'm sure there's some reason why they're necessary but I have never understood it....

PR #207, opened 2020-11-02
CMake v3.5 doesn't have VERSION_GREATER_EQUAL

alexdewar:
Addresses this problem: https://github.com/BrainsOnBoard/bob_robotics/issues/204#issuecomment-720525630

Tested with Ubuntu 16.04 docker image.


alexdewar:
Also, I think this should fix your CMake issue @neworderofjamie

PR #209, opened 2020-11-27
Mecanum motor test program

neworderofjamie:
Just made this to figure out Stathis's mecanum issues - do you think it's worth tidying this up (i.e. making the print out say what motors are 'supposed' to be turning') and keeping it in BoB robotics? The mecanum example is currently a bit pointless as it does exactly the same as joystick_robot so we could just replace it


alexdewar (approved):
I think this is a really sensible idea. It's so much harder to figure out what's not working with a robot by just trying to drive it around.

neworderofjamie:
Great, I will change the messages so they print what the robot should be doing and add some more comments etc

neworderofjamie:
@alexdewar  what do you think?

alexdewar (approved):

PR #210, opened 2020-11-27
Gps imu camera recorder

tenxlenx:
Hi Both,

Here is the recorder program for outdoor datasets.  Please tell me the issues so I can fix them :)


alexdewar:
@neworderofjamie Just to get you up to speed, I just had a chat with @tenxlenx and the plan for this PR is to get it merged fairly soon as this code has already been used for data recording and it would be good to have it in master. The idea was then to open issues for things that are outstanding. So if you could take a look that'd be great, but don't feel you have to spend too long on it!

alexdewar:
A couple of things to start with:
- CI is currently failing because of compiler warnings. I'd fix these up first
- See the suggestions on #205 for things to do with removing the velocity field etc.

include/common/nmea_parser.h
```diff
@@ -113,29 +113,30 @@ class NMEAParser {
...
+        const char delimiter = '\n';      // sentences separated by [$]
         const char w_delimiter = ',';    // elements separated by  [,]
 
         bool found = false;
-        string sentence;
+        string sentence, foundSentence;
         istringstream split(textToParse);
         while (getline(split, sentence, delimiter)) {
-            if (sentence.find(NMEA_sentence_id) != string::npos) {
+            if (std::strstr(sentence.c_str(), NMEA_sentence_id)) {
```
alexdewar, line 128: :+1: This should be faster

include/common/nmea_parser.h
```diff
@@ -160,7 +161,7 @@ class NMEAParser {
 
         try {
             if (toParse.empty()) throw GPSError("Emtpy serial output");
-            vector<string> elements= parseNMEAstring(toParse, "GNGGA"); // parse string
+            vector<string> elements= parseNMEAstring(toParse, "$GNGGA"); // parse string
```
alexdewar, line 164: See comments on #205

include/common/serial_reader.h
```diff
@@ -55,21 +57,21 @@ class SerialReader {
             }
             tcflush(fileDescriptor, TCIFLUSH);
             return serialString;
-        }
+        /*}
         catch(const char *error) {
             std::cout << "[SerialReader:readData()] : " << error << std::endl;
         }
-        return "";
+        return "";*/
```
alexdewar, line 64: I know this is technically my fault! But would you mind dropping the commented lines here...

include/common/serial_reader.h
```diff
@@ -89,13 +91,13 @@ class SerialReader {
             /* now clean the modem line and activate the settings for the port */
             tcflush(fileDescriptor, TCIFLUSH);
             tcsetattr(fileDescriptor,TCSANOW,&settings);
-        }
+/*        }
         catch(const char *s) {
             std::cout << "[SerialReader:connect()]: " << s <<  std::endl;
-        }
+        }*/
```
alexdewar, line 97: ...and here

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+
+#include <sys/stat.h>
+#include <ctime>
+
+
+
+
+
+
+// Standard C++ includes
```
alexdewar, line 27: That's a whole lot of whitespace...

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+{
+
+    //auto unwrapper = cam->createUnwrapper(cv::Size(WIDTH,HEIGHT));
+
+    while(!shouldQuit) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(5));
+        std::lock_guard<std::mutex> lock(mutex);
+        cam->readFrame(img);
+
+    }
```
alexdewar, line 55: I think you probably already know my opinion on this threading business by now :wink:. I feel we probably could remove it, but as this isn't a priority we can look at this later.

Longer term I think we want to establish whether or not it actually fixes camera problems and if it does, we should do this threading malarkey in ``Video::OpenCVCamera`` and if it isn't we should drop it.

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+
+
+}
+
+void saveThread(cv::Mat img, std::string fileName) {
+    cv::imwrite(fileName, img);
```
alexdewar, line 61: Again I think we should drop this thread, but this can also be done later.

``cv::imwrite()`` doesn't throw an error if it fails to save the file, it returns ``false`` instead. So I think we should check for this and print a warning if it hasn't saved properly.

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+
+    time_t rawtime;
+    struct tm * timeinfo;
+    char buffer[80];
+
+    time (&rawtime);
+    timeinfo = localtime(&rawtime);
+
+    strftime(buffer,sizeof(buffer),"%d-%m-%Y %H:%M:%S",timeinfo);
+    std::string str(buffer);
```
alexdewar, line 90: We've talked about the Jetson's clock issue, but let's leave this bit as is for now and raise it as a Github issue after it's merged.

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+            }
+        } catch(...) {
+            std::cout << " measuring failed, trying again in 1 second " << "[" << maxTrials - numTrials << "/" << maxTrials << "]" << std::endl;
+            std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+            numTrials--;
+        }
+    }
+    if (numTrials == 0) {
+        std::cout << " There is no valid gps measurement, please try waiting for the survey in to finish and restart the program " << std::endl;
+        exit(1);
```
alexdewar, line 116: It's better practice to do ``return EXIT_FAILURE;`` instead because that will call your objects' destructors. ``exit()`` just kills your program without cleaning anything up.

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+    int gps_minute = gps_time.minute;
+    int gps_second = gps_time.second;
+
+    std::string folderName;
+    //std::cin >> folderName;
+
+    folderName = str;
+
+    const char *c  = folderName.c_str();
+    mkdir(c, 0777);
```
alexdewar, line 130: I'd check the return code for this to make sure the directory has actually been created, e.g.:
```c++
BOB_ASSERT(mkdir(c, 0777) == 0);
```

I'm guessing that you're making this world-writable because you're running your program as root?

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
+// standard includes
```
neworderofjamie, line 1: Can you remove the binary from git?

include/common/serial_reader.h
```diff
@@ -55,21 +57,21 @@ class SerialReader {
...
-        return "";
+        return "";*/
 
     }
 
 
     void connect(const char *serial_device_path) {
     
-        try {        
+//        try {
```
neworderofjamie, line 71: can you remove these rather than just commenting them out?

include/common/serial_reader.h
```diff
@@ -31,9 +33,9 @@ class SerialReader {
 
     //! reads the serial port (USB) and returns the string from it
     std::string readData() {
-        try {
+//        try {
```
neworderofjamie, line 36: can you remove these rather than just commenting them out?

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+    time (&rawtime);
+    timeinfo = localtime(&rawtime);
+
+    strftime(buffer,sizeof(buffer),"%d-%m-%Y %H:%M:%S",timeinfo);
+    std::string str(buffer);
+
+    int seconds_to_run = 30; // default value
+    if (argc > 1) {
+        seconds_to_run = stoi(std::string(argv[1]));
+        std::cout << "running for " << seconds_to_run << " seconds" << std::endl;
```
neworderofjamie, line 95: This should be written to LOGI/LOGD rather than std::cout so it can be easily turned on and off

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+
+    }
+
+
+
+    // check to see if we get valid coordinates by checking GPS quality
+    while (numTrials > 0) {
+        try {
+            if (gps.getGPSData().gpsQuality != BoBRobotics::GPS::GPSQuality::INVALID) {
+                std::cout << " we have a valid measurement" << std::endl;
```
neworderofjamie, line 105: This should be written to LOGD rather than std::cout so it can be easily turned on and off

alexdewar (changes requested):
Cheers for posting this :smile:. I've made a few suggestions.

Would you mind also removing the ``data_recorder`` binary from the repo? (You could add a .gitignore entry for it.)

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+                break;
+            }
+        } catch(...) {
+            std::cout << " measuring failed, trying again in 1 second " << "[" << maxTrials - numTrials << "/" << maxTrials << "]" << std::endl;
```
neworderofjamie, line 109: This should be written to LOGE rather than std::cout so it can be easily turned on and off

projects/dataset_recorder/data_recorder.cc
neworderofjamie, line 116: ``return EXIT_FAILURE;`` is better here - I fear this probably long jumps over all @alexdewar 's careful safety wrapper around bobMain...

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+            std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+            numTrials--;
+        }
+    }
+    if (numTrials == 0) {
+        std::cout << " There is no valid gps measurement, please try waiting for the survey in to finish and restart the program " << std::endl;
```
neworderofjamie, line 115: LOGE

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+
+    folderName = str;
+
+    const char *c  = folderName.c_str();
+    mkdir(c, 0777);
+    std::cout << "directory " << c << " created" << std::endl;
+
+    auto cam = getPanoramicCamera();
+    const auto cameraRes = cam->getOutputSize();
+    std::cout << "camera initialised" << std::endl;
```
neworderofjamie, line 135: LOGD

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+    int gps_second = gps_time.second;
+
+    std::string folderName;
+    //std::cin >> folderName;
+
+    folderName = str;
+
+    const char *c  = folderName.c_str();
+    mkdir(c, 0777);
+    std::cout << "directory " << c << " created" << std::endl;
```
neworderofjamie, line 131: LOGD

projects/dataset_recorder/data_recorder.cc
neworderofjamie, line 130: ``filesystem::create_directory()`` is better as it's cross platform (``#include "third_party/path.h"``)

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+
+            // get gps data
+            BoBRobotics::GPS::GPSData data = gps.getGPSData();
+            BoBRobotics::MapCoordinate::GPSCoordinate coord = data.coordinate;
+            BoBRobotics::GPS::TimeStamp time = data.time;
+            int gpsQual = (int) data.gpsQuality; // gps quality
+
+            // output results
+    	   // if (i%5 == 0) { // only output once ever sec
+            std::cout << std::setprecision(10) << "GPS quality: " << gpsQual << " latitude: " << coord.lat.value()<< " longitude: " <<  coord.lon.value()  << " num sats: " << data.numberOfSatellites << " time: " << static_cast<units::time::millisecond_t>(sw_timestamp.elapsed()).value()/1000 << std::endl;
```
neworderofjamie, line 189: LOGD

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+
+    cv::Mat imgOrig;
+    //--------------------------------->>>>>>>>>>> START RECORDING <<<<<<<<<<<<----------------------------
+
+
+    // start camera thread
+    cv::Mat tmpImage, unwrappedImage, resized;
+    cv::Mat outputImage(cv::Size(WIDTH,HEIGHT), CV_8UC3);
+    std::thread readCameraThread(&readCameraThreadFunc, cam.get(), std::ref(tmpImage), std::ref(shouldQuit), std::ref(mex));
+    std::this_thread::sleep_for(std::chrono::milliseconds(3000));
```
neworderofjamie, line 149: Presumably this sleep is to wait for the camera to come alive? I think it would be much better to have the camera in the main thread (as I assume it's faster) and have a worker thread for the GPS

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+    	   // if (i%5 == 0) { // only output once ever sec
+            std::cout << std::setprecision(10) << "GPS quality: " << gpsQual << " latitude: " << coord.lat.value()<< " longitude: " <<  coord.lon.value()  << " num sats: " << data.numberOfSatellites << " time: " << static_cast<units::time::millisecond_t>(sw_timestamp.elapsed()).value()/1000 << std::endl;
+
+	        std::ostringstream fileString, folderString;
+            fileString << folderName << "/image" << i << ".jpg";
+	        folderString << folderName << "/coordinates.csv";
+            std::string fileName = fileString.str();
+
+            //  headers:   gpsQuality|latitude|longitude|altitude|roll|pitch|yaw|imagename|timestamp
+            coordinates.open (folderString.str(), std::ofstream::app); // open coordinates.csv file and append
```
neworderofjamie, line 197: why are you re-opening the file here? can't you open it once outside of the loop?

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+            coordinates.close();
+
+            // poll from camera thread
+            cv::Mat originalImage;
+            mex.lock();
+            tmpImage.copyTo(originalImage);
+            mex.unlock();
+
+            if (originalImage.size().height > 0) {
+               // cv::resize(originalImage, originalImage, cv::Size(), 0.5, 0.5);
```
neworderofjamie, line 209: ?

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+               // cv::imshow("orig",resized);
+            }
+
+            std::thread imwriteThread{ &saveThread, std::move(originalImage), std::move(fileName) };
```
neworderofjamie, line 213: I don't think this is a good idea - if you want to save every image just do it in the same thread as the camera (as I suggest above, I think it should be the main thread). If these worker threads are slower than this loop they'll be writing ``originalImage`` when it's in an ambiguous state as the next loop iteration will have started to mess with it. ``std::move`` is definitely not what you want here either

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+
+// Standard C++ includes
+#include <iostream>
+
+using namespace BoBRobotics;
+using namespace BoBRobotics::ImgProc;
+using namespace BoBRobotics::Video;
+
+
+#define WIDTH 192
```
neworderofjamie, line 35: ``constexpr int width = 192;`` is much nicer

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+               // cv::resize(originalImage, originalImage, cv::Size(), 0.5, 0.5);
+               // cv::imshow("orig",resized);
+            }
+
+            std::thread imwriteThread{ &saveThread, std::move(originalImage), std::move(fileName) };
+            imwriteThread.detach();
+
+            int count = (std::chrono::milliseconds(UPDATE_INTERVAL) - sw.elapsed()).count();
+            //std::cout << " count " << count/1000000 << " " << std::endl;
+            count /= 1000000;
```
neworderofjamie, line 218: why not have count in whatever units you actually want it in?

i.e. 
```c++
std::chrono::milliseconds count = std::chrono::milliseconds(UPDATE_INTERVAL) - sw.elapsed();
std::this_thread::sleep_for(count);
```

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,246 @@
...
+            std::this_thread::sleep_for(std::chrono::milliseconds(count));
+
+
+
+
+            if ( (static_cast<units::time::millisecond_t>(sw_timestamp.elapsed()).value()/1000)>seconds_to_run ) {
```
neworderofjamie, line 224: likewise, of seconds to run with in ``std::chrono::seconds``  you could just go:

```c++
if (sw_timestamp.elapsed() >seconds_to_run) {
```

neworderofjamie (changes requested):

include/common/nmea_parser.h
neworderofjamie: why?

alexdewar: Hmm, actually maybe not :laughing:. @tenxlenx why did you change ``NMEA_sentence_id`` from a ``std::string``? I assume it was to avoid a type conversion....?

projects/dataset_recorder/data_recorder.cc
alexdewar: In a similar vein to @neworderofjamie's suggestion above, you could also do something like:
```c++
constexpr auto UpdateInterval = 200ms;
```

You would then need a ``using namespace std::literals`` somewhere above.

projects/dataset_recorder/data_recorder.cc
neworderofjamie: you should _*never*_ have to run stuff like this as root! For serial stuff, you need to add the user to dialout and for I2C, you need to follow these instructions https://forums.developer.nvidia.com/t/accessing-gpio-without-using-sudo-su/38774

alexdewar: Yeah, it's a bit of a hack. Adding a udev rule is definitely the way to go.

include/common/nmea_parser.h
tenxlenx: This method is correctly finding the string now, the previous one was buggy and did not work properly

projects/dataset_recorder/data_recorder.cc
tenxlenx: okay, but what is the reason? Compiler likes it more?

alexdewar: It's more modern. That way your constant has a type and a namespace etc. and it won't clash with some value in a C header that's also called ``WIDTH``.

And the type thing can be useful for finding errors at compile time: e.g. with your ``UPDATE_INTERVAL``, if you do it as I suggested below then it'll have a type of ``std::chrono::milliseconds`` which will stop you from accidentally using it as a value in seconds.

projects/dataset_recorder/data_recorder.cc
tenxlenx: Yes, the sleep is for waiting for the camera.
I was thinking about the camera in the main thread and the gps as the worker thread, but I could not figure out how to do it. 
With the current method,  it is easy to just get the current image when the main loop get to the line.
Doing the other way around, I would have to start a stopwatch in the main thread for timing the gps and also to write to the csv which I found a bit clunky. What do you think?

projects/dataset_recorder/data_recorder.cc
alexdewar: @tenxlenx I think this change is worth implementing before we go out (in case there's an error writing to disk).

Replace with:
``c++
if (!cv::imwrite(fileName, img))) {
    LOGW << "Could not save image file";
}
``

src/robots/rc_car_bot.cc
```diff
@@ -86,9 +86,21 @@ RCCarBot::stopMoving()
...
 }
 
+void
+RCCarBot::updateState()
+{
+    uint8_t buffer[2];
+    m_I2C.read(buffer);
+    m_speed = 2*float(int(buffer[0])-127) / 255;
+    m_turningAngle = degree_t( 90 - int(buffer[1]));
+    std::cout << m_speed << " " << m_turningAngle << std::endl;
```
alexdewar, line 96: This should be a ``LOGD`` (or even a ``LOGV``). I'm guessing atm it spams the terminal when you run it...?

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,267 @@
...
+            coordinates << std::setprecision(10) << std::fixed << bot_speed << "," <<  turn_ang.to<double>() << "," << gpsQual << "," << coord.lat.value() << "," << coord.lon.value() << "," << data.altitude.value()  << "," << roll << "," << pitch << "," << yaw  << ",";
+            coordinates << "image" << i << ".jpg" << "," << static_cast<units::time::millisecond_t>(sw_timestamp.elapsed()).value() <<"\n";
+            coordinates.close();
+
+
+
+
+
+
+
```
alexdewar, line 234: *Definitely* not important for today, but just a reminder that you can get VS code to clean up your whitespace by right-clicking and choosing Format document.

arduino/i2c_slave_code_rccar/i2c_slave_rccar/i2c_slave_rccar.ino
```diff
@@ -89,14 +99,48 @@ bool checkRemote() {
...
 // called when data is sent through i2c to arduino
 // to update the motor values
-void receiveEvent() {
+void receiveEvent(int bytes) {
     // check incoming bytes from master
     uint8_t read_array[2];
+    int i = 0;
     while(Wire.available()) {
         read_array[i] = Wire.read();
+        i++;
```
neworderofjamie, line 108: this clearly fixes *one* issue but it could still start trampling the stack if more than two bytes are available. this is safer and nicer:
```c++
for(int i = 0; Wire.available() && i < 2; i++) {
    read_array[i] = Wire.read();
}
```

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,267 @@
...
+    // start camera thread
+    cv::Mat tmpImage, unwrappedImage, resized;
+    cv::Mat outputImage(cv::Size(WIDTH,HEIGHT), CV_8UC3);
+    std::thread readCameraThread(&readCameraThreadFunc, cam.get(), std::ref(tmpImage), std::ref(shouldQuit), std::ref(mex));
+    std::this_thread::sleep_for(std::chrono::milliseconds(3000));
+
+    // create and open file to write gps coordinates
+    std::ofstream coordinates;
```
alexdewar, line 162: Fstreams don't check for errors by default, so add this to get exceptions when something goes wrong:
```c++
coordinates.exceptions(std::ios::failbit | std::ios::badbit);
```

projects/dataset_recorder/data_recorder.cc
neworderofjamie: I think this is an important one - it's a very simple fix and constantly opening and closing files will prevent buffering which will really thrash the poor Jetson's shitty filesystem...

arduino/i2c_slave_code_rccar/i2c_slave_rccar/i2c_slave_rccar.ino
```diff
@@ -69,6 +75,10 @@ bool checkRemote() {
     // read PWM from receiver
     remoteControlSpeed = pulseIn(THROTTLE, HIGH);
     remoteControlSteering = pulseIn(STEERING, HIGH);
+    
+    
+    
+    Serial.println("speed " + String(remoteControlSpeed) + " steering " +  String(remoteControlSteering));
```
neworderofjamie, line 81: are you reading this anywhere? If not you should at least comment it out as that's going to really slow down your main loop.

neworderofjamie (changes requested):
There's a few issues that really should be fixed before this is merged but it's generally alright. As @alexdewar says the threading should also be tidied up but, as that's in the robot code, it's ok to merge without doing that

arduino/i2c_slave_code_rccar/i2c_slave_rccar/i2c_slave_rccar.ino
alexdewar: Good spot!

alexdewar:
> There's a few issues that really should be fixed before this is merged but it's generally alright. As @alexdewar says the threading should also be tidied up but, as that's in the robot code, it's ok to merge without doing that

Thanks for looking at this. I just pushed @tenxlenx 's most recent changes so I could have a quick look-through before we go out collecting data today, but I totally agree! Let's get this sucker merged soon.

projects/dataset_recorder/data_recorder.cc
neworderofjamie: i.e. your motors...

neworderofjamie:
> > There's a few issues that really should be fixed before this is merged but it's generally alright. As @alexdewar says the threading should also be tidied up but, as that's in the robot code, it's ok to merge without doing that
> 
> Thanks for looking at this. I just pushed @tenxlenx 's most recent changes so I could have a quick look-through before we go out collecting data today, but I totally agree! Let's get this sucker merged soon.

Also that image writing thread - at some point something _will_ get faster or slower and images _will_ get corrupted. Enjoy the "collecting" 😄

neworderofjamie:
this looks so much better @tenxlenx - nice work

neworderofjamie (approved):
this looks so much better @tenxlenx - nice work

alexdewar:
Hmm... This is currently getting a green tick from Jenkins even though there are warnings present. So it should be failing the tests :-/

alexdewar:
Plus clang-tidy isn't being run at all. Ugh.

alexdewar:
Gah, apologies about this @tenxlenx. Just lemme fiddle with Jenkins a bit more and see if I can get these tests running...

projects/dataset_recorder/data_recorder.cc
alexdewar: Would you mind removing this thread too? I'm pretty sure the implementation is correct, but we still don't need it and removing it will make the code cleaner. The fewer threads we have, the better.

If you're doing it so that you can initialise the hardware in parallel, use ``std::async`` (see here for example: https://github.com/BrainsOnBoard/bob_robotics/blob/master/projects/snapshot_bot/snapshot_bot.cc)

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,256 @@
...
+            numTrials--;
+        }
+    }
+
+    if (numTrials == 0) {
+        LOGW << " There is no valid gps measurement, please try waiting for the survey in to finish and restart the program ";
+        return EXIT_FAILURE;
+    }
+    std::stringstream ss;
+    ss << "imgdataset_"  << str << "_" << gps_hour << "-" << gps_minute << "-" << gps_second;
```
alexdewar, line 118: On Jenkins, there's a warning about these variables possibly being used uninitialised. Do you get a compiler warning too? (The issue is that if there is no GPS signal, the variables never get set and then they're used here anyway.)

FYI: Jenkins does test if there are compiler warnings and it'll mark things as failing if there are. You *do* want to fix them because a) they might be indicative of a real problem or b) they are harmless, but they obscure other warning messages that are real problems.

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,256 @@
...
+        tmpImage.copyTo(originalImage);
+        mex.unlock();
+        std::ostringstream fileString;
+        fileString << folderName << "/image" << i << ".jpg";
+        std::string fileName = fileString.str();
+
+
+        if (!cv::imwrite(fileName, originalImage)) {
+            LOGW << "Could not save image file";
+        }
```
alexdewar, line 166: :partying_face:

Doesn't this look so much cleaner now? :wink:

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,256 @@
...
+        // get imu data
+        float roll,pitch,yaw;
+        try {
+            angles = imu.getEulerAngles();
+            yaw = angles[0].value();
+            pitch = angles[1].value();
+            roll = angles[2].value();
+        } catch(...) {
```
alexdewar, line 174: I'd do:
```c++
} catch (std::exception &e) {
    LOGE << "Could not read from IMU: " << e.what();
   // ...
```

The problem with not printing anything is that afterwards you can't distinguish the case where there's a bug in ``getEulerAngles()`` from the case where the hardware is playing up. I know it'll probably always be the latter, but it could be frustrating if we have a CSV file of NaNs and can't diagnose what's gone wrong.

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,256 @@
...
+        degree_t turn_ang;
+        try {
+            bot.updateState();
+            bot_speed =  bot.getSpeed();
+            turn_ang = bot.getTurningAngle();
+        } catch(...) {
+            // if we can't read speed or angle, we just write nan values
+            bot_speed = NAN;
+            turn_ang = 0_deg;
+        }
```
alexdewar, line 193: Ditto above

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,256 @@
...
+            BoBRobotics::MapCoordinate::GPSCoordinate coord = data.coordinate;
+            int gpsQual = (int) data.gpsQuality; // gps quality
+
+            // output results
+            LOGI << std::setprecision(10)
+                << "GPS quality: " << gpsQual
+                << " latitude: " << coord.lat.value()
+                << " longitude: " <<  coord.lon.value()
+                << " num sats: " << data.numberOfSatellites
+                << " time: " << static_cast<units::time::millisecond_t>(sw_timestamp.elapsed()).value()/1000 << std::endl;
```
alexdewar, line 208: This could be ``LOGD`` if you don't want to show it all the time

projects/dataset_recorder/data_recorder.cc
```diff
@@ -0,0 +1,256 @@
...
+                << gpsQual << ","
+                << coord.lat.value() << ","
+                << coord.lon.value() << ","
+                << data.altitude.value()  << ","
+                << roll << ","
+                << pitch << ","
+                << yaw  << ","
+                << "image" << i << ".jpg" << "," << static_cast<units::time::millisecond_t>(sw_timestamp.elapsed()).value() <<"\n";
+            // calculating time to wait if loop was done faster thann update time
+            std::this_thread::sleep_for(std::chrono::milliseconds(UPDATE_INTERVAL) - sw.elapsed());
```
alexdewar, line 222: This could now simply be:
```c++
std::this_thread::sleep_for(UPDATE_INTERVAL - sw.elapsed());
```

alexdewar (changes requested):
I've made some really minor suggestions. The most important of these is to fix the failing tests (i.e. the compiler warnings), because we can't merge with broken tests!

alexdewar (changes requested):
I've made a few small suggestions, but on the whole this is looking excellent. Good job! This is *much* nicer than it was before.

The most important thing to fix is the test failures (i.e. the compiler warnings) because we can't merge while Jenkins is unhappy....

alexdewar:
> I've made a few small suggestions, but on the whole this is looking excellent. Good job! This is _much_ nicer than it was before.
> 
> The most important thing to fix is the test failures (i.e. the compiler warnings) because we can't merge while Jenkins is unhappy....

Now that the tests seem to be firing properly, Jenkins has also spotted an old problem in ``gps.h`` which will also need to be fixed: https://gen-ci.inf.sussex.ac.uk/job/BoB%20robotics/job/bob_robotics/job/gps_imu_camera_recorder/22/clang_tidy_INF900802/source.bc96b7ce-f0eb-4ee9-9399-e800b380c234/#25

(It's not a biggie though)

arduino/i2c_slave_code_rccar/i2c_slave_rccar/i2c_slave_rccar.ino
```diff
@@ -89,14 +99,44 @@ bool checkRemote() {
...
     // check incoming bytes from master
     uint8_t read_array[2];
-    while(Wire.available()) {
+    int i = 0;
+    for(int i = 0; Wire.available() && i < 2; i++) {
         read_array[i] = Wire.read();
     }
+    
     movement[0] = read_array[0]; // throttle value
     movement[1] = read_array[1]; // steering value
```
alexdewar, line 111: I've only just spotted this, but there is the possibility that you'll read garbage values from ``read_array``, if ``Wire.available()`` returns ``false`` unexpectedly. I guess in that case, we'd want to leave ``movement`` as is?

tenxlenx: I have added a condition now to check whether all bytes were read successfully

arduino/i2c_slave_code_rccar/i2c_slave_rccar/i2c_slave_rccar.ino
```diff
@@ -101,40 +101,43 @@ bool checkRemote() {
...
+    int read_byte = 0; // number of bytes read
+    for(read_byte = 0; Wire.available() && read_byte < bytes; read_byte++) {
+        read_array[read_byte] = Wire.read();
+    }
+
+    if (read_byte+1==bytes) { // check if we did read all the data
+        movement[0] = read_array[0]; // throttle value
+        movement[1] = read_array[1]; // steering value
+        updateMotors = true;
     }
```
alexdewar, line 114: Much better! I hadn't noticed that you weren't using ``bytes`` before either.

tnowotny: Are you sure it should be ```if (read_byte+1==bytes) ``` ? If the ```Wire.available()``` remains true throughout, the loop ends with  ```read_byte == bytes``` I would have thought?

alexdewar: You're totally right. D'oh! We've merged this branch now but I'll push a fix to master.

alexdewar: Scratch that. It occurs to me that this code *can't* have been tested, because at the moment, ``movement`` will never be updated. Untested code should *not* be going into the master branch.

@tenxlenx Is this the code you're running on the Arduino at present? If so, the speed and steering values won't be being saved. (If not, then why not...?) I'd appreciate it if you could open a PR with a fix and confirm that this has been tested properly before we push it to master.

tenxlenx: > Are you sure it should be `if (read_byte+1==bytes) ` ? If the `Wire.available()` remains true throughout, the loop ends with `read_byte == bytes` I would have thought?

You are totally right Thomas read_byte+1 == bytes, I'll fix that asap

tenxlenx: > Scratch that. It occurs to me that this code _can't_ have been tested, because at the moment, `movement` will never be updated. Untested code should _not_ be going into the master branch.
> 
> @tenxlenx Is this the code you're running on the Arduino at present? If so, the speed and steering values won't be being saved. (If not, then why not...?) I'd appreciate it if you could open a PR with a fix and confirm that this has been tested properly before we push it to master.

@alexdewar I uploaded this to the robot, but this is only to control the robot from the Jetson not saving values to our csv file, so maybe we should move this issue somewhere else,  as it is not directly effecting the recorder program. For now I could just remove the whole receiveEvent() function and re-add it in another pull-request.

alexdewar: Ahh ok. In that case feel free to push a fix directly to master for it. It doesn't really need a separate PR.

PR #211, opened 2021-02-08
Fixes to infomax code

alexdewar:
Hey,

This is mostly just a couple of small fixes for the infomax code + adding some tests, with a small amount of refactoring.

I've fixed the following bugs:
* The weight matrix was ``numPixels * (numPixels - 1)`` in size by default rather than ``numPixels * numPixels`` as we usually report it
* The weights were being normalised by row-wise variance rather than SD as they should have been (though as they're still random I suspect it won't make any difference)
* With a high learning rate the weights will end up too high and become NaNs, which silently breaks things, so I've added a check for this

(This does mean that it *could* have been an issue in previous experiments, but I think it's really unlikely, though we should definitely have this sanity check in there. I only noticed it when I was fiddling with parameters.)

Also, CC @aphilippides FYI.


examples/infomax/infomax_matlab_comparison.cc
```diff
@@ -39,14 +39,14 @@ runTest(const filesystem::path &dataPath, int num)
 
     // Load matrices of weights
     const auto pref = "test"s + snum + "_"s;
-    const auto initWeights = readData<>(dataPath / (pref + "weights_init.bin"s));
-    const auto matlabOutputWeights = readData<>(dataPath / (pref + "weights_out.bin"s));
-    const auto matlabOutputWeightsMany = readData<>(dataPath / (pref + "weights_out_many.bin"));
-    const auto matlabU = readData<>(dataPath / (pref + "u.bin"s));
-    const auto matlabY = readData<>(dataPath / (pref + "y.bin"s));
+    const auto initWeights = readMatrix<double>(dataPath / (pref + "weights_init.bin"s));
```
neworderofjamie, line 42: what do the ``s`` do?

examples/infomax/infomax_matlab_comparison.cc
```diff
@@ -39,14 +39,14 @@ runTest(const filesystem::path &dataPath, int num)
...
+    const auto matlabOutputWeights = readMatrix<double>(dataPath / (pref + "weights_out.bin"s));
```
neworderofjamie, line 43: I _seem_ to remember matlab binary files are column-major  - is that also the default for Eigen?

include/common/serialise_matrix.h
```diff
@@ -0,0 +1,50 @@
...
+namespace BoBRobotics {
+//! Read a matrix of type T from a binary file
+template<class T>
+auto
+readMatrix(const filesystem::path &filepath)
+{
+    // Open file
+    std::ifstream ifs;
+    ifs.exceptions(std::ios::badbit | std::ios::eofbit);
+    ifs.open(filepath.str(), std::ios::in | std::ios::binary);
```
neworderofjamie, line 24: **really** not important but ``std::ifstream`` always |s ``std::ios::in`` with the flags that are passed (weird behaviour but still) so it's not necessary to repeat it

include/common/serialise_matrix.h
```diff
@@ -0,0 +1,50 @@
...
+    using namespace Eigen;
+    Matrix<T, Dynamic, Dynamic> data(size[0], size[1]);
+    ifs.read(reinterpret_cast<char *>(data.data()), sizeof(T) * data.size());
+
+    return data;
+}
+
+template<class T>
+void
+writeMatrix(const filesystem::path &filepath, const T &matrix)
```
neworderofjamie, line 40: Might be a little safer to template this on:
```c++
template<class T, int R, int C>
void 
writeMatrix(const filesystem::path &filepath, const Matrix<T, R, C> &matrix)
```
(or whatever template arguments Matrix has)

include/common/serialise_matrix.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace BoBRobotics {
+//! Read a matrix of type T from a binary file
+template<class T>
+auto
+readMatrix(const filesystem::path &filepath)
+{
+    // Open file
+    std::ifstream ifs;
+    ifs.exceptions(std::ios::badbit | std::ios::eofbit);
```
neworderofjamie, line 23: what's the intention of this? it's a new unitialised stream so no flags should be set - checking ``ifs.good()`` after opening might be a good idea though - otherwise, I believe, it'll fail silently

include/navigation/infomax.h
```diff
@@ -81,6 +88,42 @@ class InfoMax : public VisualNavigationBase
...
+        std::generate_n(weights.data(), weights.size(), [&]() { return distribution(generator); });
+
+        LOG_VERBOSE << "Initial weights" << std::endl
+                    << weights;
+
+        // Normalise mean and SD for row so mean == 0 and SD == 1
+        const auto means = weights.rowwise().mean();
+        LOG_VERBOSE << "Means" << std::endl << means;
+
+        weights.colwise() -= means;
```
neworderofjamie, line 111: Is normalising the columns by the rows (I _think_ that's what this does) intentional? with square matrices I think it'd work but probably not provide very good normalisation

include/navigation/infomax.h
```diff
@@ -90,7 +133,21 @@ class InfoMax : public VisualNavigationBase
...
+         * weight matrix, which will silently muck up subsequent calculations.
+         * I don't *think* this will ever be an issue with a sensibly small
+         * learning rate, but if so, the learning rate could be reduced at this
+         * point instead of just bailing out.
+         */
+        const auto ptr = m_Weights.data();
+        const auto check = [](auto val) { return std::isnan(val); };
+        if (std::any_of(&ptr[0], &ptr[m_Weights.size()], check)) {
+            throw WeightsBlewUpError{};
+        }
```
neworderofjamie, line 150: By randomly googling, I saw this nan check:
```c++
return ((x.array() == x.array())).all();
```
which should get vectorized/parallelised by Eigen so likely to be faster

tests/infomax.cc
```diff
@@ -0,0 +1,53 @@
+#include "common.h"
```
neworderofjamie, line 1: If you want to run these automatically from Jenkins you can get the google tests to emit 'junit':
```bash
tests --gtest_output="xml:test_results.xml"
````
And then add a stage like this to the Jenkins file
```groovy
buildStep("Gathering test results (" + env.NODE_NAME + ")") {
    dir("genn/tests") {
         // Process JUnit test output
         junit "**/test_results*.xml";
     }
}
```

neworderofjamie (commented):
Looks cool - weight handling was always the most irritating aspect of the InfoMax code and this really improves it. I _think_ I spotted a bug but the rest is just suggestions/random C++ary 😄

examples/infomax/infomax_matlab_comparison.cc
alexdewar: Makes it into a ``std::string`` rather than ``const char *``.

examples/infomax/infomax_matlab_comparison.cc
alexdewar: Yeah, although I wasn't trying to make them compatible with matlab here. We're also writing the matrix size into the file first so I think that'll mess matlab up.

include/common/serialise_matrix.h
alexdewar: Heh TIL. C++ is odd.

include/common/serialise_matrix.h
alexdewar: Good idea. It would also be more descriptive that way.

alexdewar: Actually it seems like this won't really work because Eigen's templated types can end up really complicated, e.g. I'm wrapping around another data structure somewhere and end up with a ``Eigen::Map<Eigen::Matrix<float, -1, 1>, 0, Eigen::Stride<0, 0> >`` which won't work with your suggestion :cry:. (You'd think it'd end up being implicitly casted, but apparently not!)

include/common/serialise_matrix.h
alexdewar: This sets up the exception mask for any *future* operations you do with ``ifs`` (including ``open()`` and ``read()``), otherwise they just fail silently (I just ran a little test to check this is true). I think the issue is that fstreams were added to C++ before exceptions were, so this was tacked on as an afterthought.

This does mean that we're not properly checking for errors elsewhere in BoB robotics, but I guess once a file is open then it's probably not too likely to fail on a ``read()`` or ``write()`` so it probs won't cause many problems.

include/navigation/infomax.h
alexdewar: I did a double-take here because it does *look* wrong, but what it's actually doing is getting the mean for each column (as a vector which will be ``numColumns`` long) and then subtracting that from each of the columns. So in the end your columns end up having a mean of approx 0.

include/navigation/infomax.h
alexdewar: Ah this does look nicer than my hack. Thanks!

tests/infomax.cc
alexdewar: This is really cool! I might attempt this in a separate PR. We could probably also do with building things in ``tools`` and ``projects`` as part of our CI.

include/navigation/infomax.h
neworderofjamie: No worries - I dunno about nicer but perhaps idiomatic Eigen rather than idiomatic C++

alexdewar:
Thanks for looking at this @neworderofjamie. Are we ready to merge...?

PR #213, opened 2021-02-13
Serial and gps updates

alexdewar:
Hey,

I've had a go at reworking the code for the GPS with the aim of:
* fixing the weird parsing errors we're getting
* adding support for non-blocking reads

As part of this, I've changed the code over to using the same ``SerialInterface`` class we're using for the mecanum robot (rather than the separate ``SerialReader`` class) and I've tried to implement the proper line-reading functionality. Btw if someone with a mecanum robot (@efkag? @ajc158) could test this quickly to check I haven't broken anything that'd be awesome :smile:

Obvs the GPS code needs retesting before we merge, so, @tenxlenx , would you mind checking that the GPS example still works and gives sensible output when you run it, please? I've also added an option to test the non-blocking functionality to the example (just run ``./gps nonblock``) -- could you please check this too? I haven't yet switched over the ``data_recorder`` to use non-blocking GPS, but that's the eventual aim.


include/common/nmea.h
```diff
@@ -0,0 +1,205 @@
...
+    arcminute_t latitudeMinutes, longitudeMinutes;
+    char latDirection, longDirection;
+    meter_t altitude;
+    double horizontalDilution;
+    int numberOfSatellites;
+    int gpsQualityIndicator;
+    GPSQuality qualityOfGps;
+
+    // Let's reuse our data structure so we don't have to reallocate with every call
+    static array<string, 9> elements;
```
neworderofjamie, line 157: I think this will be entirely stack allocated and this makes it very un thread safe so I don't think this is a good idea

include/common/nmea.h
```diff
@@ -0,0 +1,205 @@
...
+    if (!splitString(toParse, elements, "$GNGGA")) {
+        return false;
+    }
+
+    try {
+        TimeStamp time{ elements[0] };
+        latitude = degree_t(stod(elements[1].substr(0, 2)));
```
neworderofjamie, line 164: why not just declare these here rather than above?

src/robots/mecanum.cc
```diff
@@ -15,7 +15,28 @@ namespace BoBRobotics {
 namespace Robots {
 Mecanum::Mecanum(const char *path, bool alternativeWiring)
   : m_Serial(path), m_AlternativeWiring(alternativeWiring)
-{}
+{
+    auto tty = m_Serial.getAttributes();
```
neworderofjamie, line 19: most of this stuff is pretty boilerplatey - it seems a bit of a shame to expose all this posix horror t all users of the class.

neworderofjamie (commented):
haha, if you want something done right do it yourself no eh? 😄 this is all way better - multiple serial interfaces was obviously a bad idea. My only concern is with the whole exposing of all the termios attributes. most of them are boilerplate, probably all that really needs exposing is baud, stop bits and parity

examples/gps/gps.cc
```diff
@@ -1,60 +1,78 @@
...
+using namespace BoBRobotics::GPS;
+using namespace std::literals;
 
-int bobMain(int, char **)
+int
+bobMain(int argc, char **argv)
 {
-    const char *path_linux = "/dev/ttyACM0";       // path for linux systems
-   // const char *path_mac = "/dev/cu.usbmodem141401"; // the path for mac is often different!
+    const char *path_linux = "/dev/ttyACM0"; // path for linux systems
```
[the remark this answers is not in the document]
alexdewar: Good idea!

include/common/nmea.h
```diff
@@ -0,0 +1,205 @@
...
+    double horizontalDilution;                            // horizontal dilution - lower value is better
+    GPSQuality gpsQuality;                                // GPS quality indicator
+    TimeStamp time;                                       // time of measurement
+};
+
+namespace NMEA {
+
+template<size_t N>
+bool
+splitString(const std::string &line,
```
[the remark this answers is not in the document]
alexdewar: Good point. That would certainly be a bit tidier.

include/common/nmea.h
alexdewar: Hmm yeah. I was thinking that this way we could avoid heap allocations but I suppose that's not really likely to be much of a problem (we're only reading at 5Hz anyway).

include/common/nmea.h
alexdewar: True! There's quite a lot of boilerplate going on here.

include/common/nmea.h
neworderofjamie: I think at 5Hz, a few string allocations'll be alright (I was clearly being dense before - I see what you were trying to do now) but, if it does become slow, turning this into an ``NMEAParser`` class or something would be nicer.

include/common/nmea.h
```diff
@@ -144,57 +144,30 @@ parse(const std::string &toParse, GPSData &data)
...
+        data.coordinate.lon = degree_t(stod(elements[3].substr(0, 3))) +
+                              arcminute_t(stod(elements[3].substr(3, 9)));
+        if (elements[4][0] == 'S') {
+            data.coordinate.lon = -data.coordinate.lon;
+        }
+        data.gpsQuality = static_cast<GPSQuality>(stoi(elements[5]));
+        data.numberOfSatellites = stoi(elements[6]);
+        data.horizontalDilution = stod(elements[7]);
+        data.coordinate.height = meter_t{ stod(elements[8]) };
+    } catch (invalid_argument &e) {
```
neworderofjamie, line 168: oh wow, there was even more boilerplate than I thought 😄

neworderofjamie (approved):

src/robots/mecanum.cc
alexdewar: Yeah, I know :-(. The problem is that the two users of the class (``Robots::Mecanum`` and ``GPSReader``) set a bunch of different options besides and I wasn't sure which were the ones that I could safely get rid of without breaking things.

neworderofjamie: I suspect all that's different between the two devices (and 99.9% of others) is the baud rate, the number of data and stop bits and the parity. Looking at the code (which I suspect I copy-and-pasted), as I memset the whole structure to zero, the "non-canonical" flag setting has no effect. Sounds like the VMIN and VTIME business (http://unixwiz.net/techtips/termios-vmin-vtime.html) might be worth exposing as well though

neworderofjamie: Actually, remembering more and googling:
"Canonical Mode: In canonical mode, read on the serial port will not return until a new line, EOF or EOL character is received by the Linux Kernel. ... Non-Canonical Mode: In non-canonical mode, read on the serial port will not wait until the new line. Number of bytes read depends on the MIN and TIME settings."

So I suspect the mecanum should be non-canonical and the GPS canonical so...maybe...just exposing all the settings is best

PR #214, opened 2021-02-15
projects/dataset_recorder: Call sync() before program exit

alexdewar:
This ensures that the data is physically written to disk and not just
lurking in OS memory somewhere. Should make things a bit safer for when we
next get a random power-off.


alexdewar:
Given that this is fairly trivial, I'm just going to merge it into your branch @tenxlenx

PR #215, opened 2021-02-16
Extra CI tests

alexdewar:
I'd been meaning to do this for ages but never got round to it. This PR adds the projects in ``tools`` and ``projects`` to the build tests we run and also checks the gtest results.

A couple of issues seem to be preventing things from building on your computer for the time being: it seems like CMake can't find GeNN or Google Test. Are these installed? (There are also some other warnings/errors to fix but this seems like a good place to start!)


Jenkinsfile
```diff
@@ -21,6 +21,42 @@ void setBuildStatus(String message, String state) {
...
+            }
+
+            // Parse test output for GCC warnings
+            // **NOTE** driving WarningsPublisher from pipeline is entirely undocumented
+            // this is based mostly on examples here https://github.com/kitconcept/jenkins-pipeline-examples
+            // **YUCK** fatal errors aren't detected by the 'GNU Make + GNU C Compiler (gcc)' parser
+            // however JENKINS-18081 fixes this for
+            // the 'GNU compiler 4 (gcc)' parser at the expense of it not detecting make errors...
+            def parserName = ("mac" in nodeLabel) ? "Apple LLVM Compiler (Clang)" : "GNU compiler 4 (gcc)";
+            step([$class: "WarningsPublisher",
```
neworderofjamie, line 50: It might be worth following this sorry tale and trying out "GNU Make + GNU C Compiler (gcc)". However, the warnings plugin is deprecated and shit so you _could_ do me a **massive** favour and have a go with the snazzily named "Warnings Next Generation" plugin 😄 It is installed on the jenkins machine and documented [here](https://www.jenkins.io/doc/pipeline/steps/warnings-ng/). I would guesstimate that the ``scanForIssues`` command is what you want.

Jenkinsfile
```diff
@@ -62,40 +98,26 @@ for(b = 0; b < builderNodes.size(); b++) {
...
+            runBuild("projects", nodeLabel);
+            runBuild("tools", nodeLabel);
+            runBuild("tests", nodeLabel);
+
+            stage("Running tests (" + env.NODE_NAME + ")") {
+                dir("tests") {
+                    // Generate unique name for message
+                    def uniqueMsg = "msg_test_results_" + env.NODE_NAME;
+                    def runTestsCommand = "./tests --gtest_output=xml:test_results.xml 1>> \"" + uniqueMsg + "\" 2>> \"" + uniqueMsg + "\"";
+                    def runTestsStatus = sh script:runTestsCommand, returnStatus:true;
```
neworderofjamie, line 111: you should probably do something with this status here e.g. fail the build after you've archived the output

neworderofjamie (commented):
Regarding GeNN and gtest, you shouldn't really rely on them being installed somewhere particularly. You should probably checkout GeNN master with something like https://github.com/genn-team/pynn_genn/blob/master/Jenkinsfile#L113-L130 and download a gtest release like https://github.com/genn-team/genn/blob/master/Jenkinsfile#L128-L139 (ideally a newer one 😄 )

Jenkinsfile
alexdewar: Good point!

projects/ardin_mb/CMakeLists.txt
```diff
@@ -4,13 +4,19 @@ include(../../cmake/bob_robotics.cmake)
...
 else()
     BoB_project(EXECUTABLE ardin_mb
-                SOURCES ardin_mb.cc state_handler.cc mb_memory.cc opencv_texture.cc mb_memory_ardin.cc vector_field.cc visual_navigation_ui.cc
+                SOURCES ardin_mb.cc state_handler.cc mb_memory.cc
+                        opencv_texture.cc mb_memory_ardin.cc vector_field.cc
+                        visual_navigation_ui.cc
                 BOB_MODULES common antworld navigation
-                THIRD_PARTY imgui
-                INCLUDE_GENN_USERPROJECTS)
+                GENN_MODEL model_ardin.cc
```
neworderofjamie, line 17: These weren't included for a reason here - this example uses runtime dynamic linking so it has a plugin style architecture where it loads dynamic libraries previously built by GeNN. That clearly doesn't play nice with CI so, personally, I'd just skip this one. The stone CX model is a much simpler static linking against a GeNN model where this automagical stuff should work. In GeNN I just add files called "skip" to directories with tests I don't wanna run and detect them in [bash](https://github.com/genn-team/genn/blob/master/tests/run_tests.sh#L61-L62)

projects/ardin_mb/model_ardin.cc
```diff
@@ -57,7 +57,7 @@ void modelDefinition(NNmodel &model)
 {
     using namespace BoBRobotics;
     model.setDT(MBParamsArdin::timestepMs);
-    model.setName("mb_memory_ardin");
+    model.setName("ardin_mb");
```
neworderofjamie, line 60: this definitely breaks the runtime dynamic linking as it looks to load a model called mb_memory_ardin...

cmake/bob_robotics.cmake
```diff
@@ -94,13 +94,17 @@ macro(BoB_project)
...
+            # If the user just wants the userprojects headers...
+            if(PARSED_ARGS_INCLUDE_GENN_USERPROJECTS)
                 # On *nix link dl
                 if(NOT WIN32)
                     BoB_add_link_libraries(dl)
                 endif()
 
+                # If GeNN is installed to /usr the above method for finding
+                # userprojects headers won't work
+                BoB_add_include_directories(/usr/include/genn)
```
neworderofjamie, line 106: I'm guessing our /usr install (which sadly I seldom use/test) doesn't do anything sensible with the userprojects includes?

projects/ardin_mb/CMakeLists.txt
alexdewar: Ahh I see... Ok I won't test this project then

cmake/bob_robotics.cmake
alexdewar: Nope :disappointed:. For my Arch package I just copy over the ``userprojects`` folder and symlink the headers into /usr/include: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=genn

neworderofjamie: next major version I'll tidy the mess of userprojects up...I kinda deprecated ``/userprojects/include`` and then resurrected it, forgetting that it wasn't really installable

neworderofjamie: Do arch packages get rebuilt automatically with releases btw?

alexdewar: I have to update the package and push it back to the AUR but if there is anyone out there who is also using the package then they should just get updates through their package manager (though it will build it from source).

Jenkinsfile
alexdewar: I tried this and it *seemed* to work (I deliberately put a bug in a file and Jenkins detected it) but now I seem to be getting this warning:

```

java.lang.IllegalStateException: ID gcc is already used by another action: io.jenkins.plugins.analysis.core.model.ResultAction for GNU C Compiler (gcc)

	at io.jenkins.plugins.analysis.core.steps.IssuesPublisher.ensureThatIdIsUnique(IssuesPublisher.java:117)
	at io.jenkins.plugins.analysis.core.steps.IssuesPublisher.attachAction(IssuesPublisher.java:92)
	at io.jenkins.plugins.analysis.core.steps.IssuesRecorder.publishResult(IssuesRecorder.java:623)
	at io.jenkins.plugins.analysis.core.steps.IssuesRecorder.record(IssuesRecorder.java:556)
	at io.jenkins.plugins.analysis.core.steps.IssuesRecorder.perform(IssuesRecorder.java:521)
	at io.jenkins.plugins.analysis.core.steps.RecordIssuesStep$Execution.run(RecordIssuesStep.java:942)
	at io.jenkins.plugins.analysis.core.steps.RecordIssuesStep$Execution.run(RecordIssuesStep.java:907)
	at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:47)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

Any ideas?

neworderofjamie: Maybe try turning off ``aggregatingResults`` as you only use a single tool (based on https://issues.jenkins.io/browse/JENKINS-54027)

neworderofjamie: you may also be able to set ``id: env.NODE_NAME`` and get actual per-node error reports rather than the weird blend we have on GeNN

alexdewar: YESSSSS! Setting the ``id`` totally solved it.

I've tried deliberately breaking tests and adding warnings and it all seems to work :partying_face:

alexdewar:
Now *this* looks like a good idea: https://pspdfkit.com/blog/2018/using-clang-tidy-and-integrating-it-in-jenkins/

It's pretty trivial to get the necessary ``compile_commands.json`` file from CMake then run clang-tidy over your code. The preliminary messing around I've done with it found a few issues. What do you think?

neworderofjamie:
Sounds good to me! The question would be where to run it - the Jenkins machine itself would make sense but getting clang installed on it may be a little tricky as it runs ancient CentOS.

alexdewar:
Maybe it's just easier to run it on my machine then? I guess it doesn't make much sense to run it on multiple machines...

Alternatively we could just build all the necesarry LLVM stuff from source on the Jenkins machine.

neworderofjamie:
what exactly is required?

alexdewar:
It looks like it's included with clang on my machine. So I guess we'd need to build llvm, then clang if we build from source.

alexdewar:
Unless... I don't suppose the Jenkins machine has docker on it?

neworderofjamie:
What sort of version of llvm are we looking for? I'll poke the jenkins
machine.

Docker's a whole other story. I think the next big upgrade of this whole
system would be to run all the tests within docker containers so we don't
have to mess with manually installing dependencies which is definitely a
mode of operation jenkins encourages.

alexdewar:
I've no idea, but I suppose the newer the better...?

alexdewar:
And :+1: for using docker!

neworderofjamie:
Ok, I've installed clang 5 which you can enable with:
```bash
source /opt/rh/llvm-toolset-7/enable
```
Each Jenkins shell block is it's own shell btw so the sourcing will only work within that

alexdewar:
Thanks! Although I've just realised that we may have to run clang-tidy on our own machines anyway so that it has access to all the headers we include from external libraries :worried:

neworderofjamie:
That is a good point - up to you whether you try adding that now or in a future PR - INF900801 has clang installed so it should be all good to go

alexdewar:
> That is a good point - up to you whether you try adding that now or in a future PR - INF900801 has clang installed so it should be all good to go

I think you might also have to do an ``apt install clang-tidy``, unfortunately

alexdewar:
Woooooo it's working! I'm only doing the default checks for now but there are a bunch of others which could be useful...

neworderofjamie:
NOICE! those virtual call during destructor things are a good automatic-shout as well..

alexdewar:
@neworderofjamie I think this is almost ready to go :partying_face:. I've picked what I think are a sensible set of tests and fixed up all the warnings.

The only issue is that ``clang-tidy`` doesn't seem to be installed on your desktop machine. So I guess either you could install it, or we could just run those tests on my machine (I don't suppose there's much advantage to doing two passes of the codebase!).

The tests are currently taking about 8 min to run, which really isn't too bad. If it starts to get annoying at some point (or it eats up all the CPU) we can always restrict the tests to newly changed files.

alexdewar:
One issue is that old build/test logs don't seem to be being cleaned up, so they get appended to (e.g. https://gen-ci.inf.sussex.ac.uk/job/BoB%20robotics/job/bob_robotics/job/extra_ci_testing/65/artifact/msg_clang_tidy_INF900802/*view*/).

It doesn't affect the output of the junit plugin though, so perhaps it's not a biggie.

cmake/bob_robotics.cmake
```diff
@@ -80,77 +80,88 @@ macro(BoB_project)
...
+            endif()
+        endif()
+
+        # Remove filename to get path to GeNN bin directory
+        get_filename_component(GENN_BIN_PATH ${GENN_BUILDMODEL} DIRECTORY)
+
+        # Get absolute path to userproject include
+        get_filename_component(GENN_USERPROJECT ${GENN_BIN_PATH}/../userproject/include ABSOLUTE)
+        message("GeNN found in ${GENN_USERPROJECT}")
+        BoB_add_include_directories(${GENN_USERPROJECT})
```
neworderofjamie, line 105: this could probably go in the ``else`` clause above - if you have ``GENN_PATH``, getting ``GENN_USER_PROJECT`` is trivial 😄

bin/run_clang_tidy_check
```diff
@@ -0,0 +1,237 @@
...
+    'openmp-use-default-none',
+    'performance-type-promotion-in-math-fn',
+)
+
+# Create a `ErrorDescription` tuple with all the information we want to keep.
+ErrorDescription = collections.namedtuple(
+    'ErrorDescription', 'file line column error error_identifier description')
+
+# This class was adapted from here: https://github.com/PSPDFKit-labs/clang-tidy-to-junit
+class ClangTidyConverter:
```
neworderofjamie, line 65: This is cool but presenting the clang-tidy output via junit seems weird to me as they appear in the tests section of the jenkins report - wouldn't it make more sense to show these through warnings NG (which claims to support clang-tidy output)

src/common/threadable.cc
```diff
@@ -11,7 +11,7 @@ Threadable::Threadable()
 {}
 
 Threadable::~Threadable()
-{}
+= default;
```
neworderofjamie, line 14: I tend to do the previous thing we had here too - what is the difference?

src/robots/ev3/ev3.cc
```diff
@@ -26,8 +26,8 @@ checkMotor(const EV3::MotorType &motor, const std::string &label)
     }
 }
 
-EV3::EV3(const ev3dev::address_type leftMotorPort,
-         const ev3dev::address_type rightMotorPort)
+EV3::EV3(const ev3dev::address_type &leftMotorPort,
```
neworderofjamie, line 29: aren't these literally pointers or some other small type that probably shouldn't be passed by reference? (they probably should not have been ``const`` if that's true but hey)

src/robots/gazebo/uav_plugin.cc
```diff
@@ -345,13 +351,13 @@ GazeboQuadCopterPlugin::ApplyMotorForces(const double _dt)
 {
     MotorMixing(_dt);
     // update velocity PID for rotors and apply force to joint
-    for (size_t i = 0; i < m_Rotors.size(); ++i) {
-        const double velTarget = m_Rotors[i].m_Multiplier *
-                           m_Rotors[i].m_Cmd /
-                           m_Rotors[i].m_RotorVelocitySlowdownSim;
-        const double vel = m_Rotors[i].m_Joint->GetVelocity(0);
+    for (auto &rotor : m_Rotors) {
```
neworderofjamie, line 354: oh wow - did clang-tidy spot that this loop could be achieved without messing about with indices? 😮

src/video/see3cam_cu40.cc
```diff
@@ -209,10 +208,10 @@ See3CAM_CU40::calculateImageEntropy(const cv::Mat &mask)
 
     // Sum together entropy for each colour channel
     float entropy = 0.0f;
-    for (unsigned int c = 0; c < 3; c++) {
+    for (auto &c : hist) {
```
neworderofjamie, line 211: this could be ``const &c`` right?

src/video/v4l_camera.cc
```diff
@@ -157,7 +157,7 @@ Video4LinuxCamera::open(const std::string &device,
 }
 
 void
-Video4LinuxCamera::enumerateControls(std::function<void(const v4l2_queryctrl &)> processControl)
+Video4LinuxCamera::enumerateControls(const std::function<void(const v4l2_queryctrl &)> &processControl)
```
neworderofjamie, line 160: ``std::function`` is pretty lightweight - I think it's ok to pass by value, does clang-tidy disagree?

tools/unwrapconfig/unwrapconfig.cc
```diff
@@ -22,7 +22,7 @@ const int BIG_PX_JUMP = 5;
 inline void
 drawCalibrationLine(cv::Mat &imorig, cv::Point p1, cv::Point p2)
 {
-    cv::line(imorig, p1, p2, cv::Scalar(0x00, 0xff, 0x00), 2);
+    cv::line(imorig, std::move(p1), std::move(p2), cv::Scalar(0x00, 0xff, 0x00), 2);
```
neworderofjamie, line 25: why would you move these? ``cv::Point`` has no internal resources that moving would prevent the need to copy

neworderofjamie (commented):
This looks really cool - when I'm slightly less busy I am definitely going to hook some of this awesome up for GeNN as well. There are a few clang-tidy decisions I disagree with that might be worth looking at/figuring whether I'm wrong or the tests are a bit false-positivey to bother with

cmake/bob_robotics.cmake
alexdewar: Good point!

bin/run_clang_tidy_check
alexdewar: Hmm... I didn't notice that warnings NG supported it! I basically stole the idea and most of the code from here: https://pspdfkit.com/blog/2018/using-clang-tidy-and-integrating-it-in-jenkins/

I'm guessing that they were converting to junit files either a) because warnings NG wasn't a thing when they did it or b) the output format is more useful.

I'll maybe try pushing a random error to the tree to see how warnings NG formats it. But if it looks ok, then that's definitely more sensible.

src/common/threadable.cc
alexdewar: Nothing -- it's just a style thing. Now that you point it out, I see that it's a pretty dubious improvement at best :thinking:.

Let's just disable this warning.

src/robots/ev3/ev3.cc
alexdewar: These are actually just ``typedef``'d ``std::string``s (though why they bothered with that I don't know), so in this particular case it does make sense.

src/robots/gazebo/uav_plugin.cc
alexdewar: Yes! And it even had a go at writing a patch for it. I did clean it up a bit, but it's still nice!

src/video/see3cam_cu40.cc
alexdewar: True!

src/video/v4l_camera.cc
alexdewar: It does, though it may be being overzealous here.

Basically the options are:
1. disable the warning altogether (but I do like it, because it *did* find some places where we really should have been passing by reference)
2. add a magic comment to tell clang-tidy to ignore this particular case
3. turn it into a ``const`` reference

What do you think?

tools/unwrapconfig/unwrapconfig.cc
alexdewar: That's true. This is definitely not an improvement! Alternatively we could pass ``p1`` and ``p2`` by reference or suppress the warning here.

src/video/v4l_camera.cc
alexdewar: I've just checked this, because I wasn't sure whether it made sense to take a ``const`` reference to a ``std::function`` in particular and it seems like you probably should: https://stackoverflow.com/questions/8711391/should-i-copy-an-stdfunction-or-can-i-always-take-a-reference-to-it?lq=1

It *could* be any arbitrary callable object, not just a function pointer.

alexdewar:
@neworderofjamie I think this is all ready now, if you're happy to merge...?

I've made it skip running ``clang-tidy`` on your machine. I'm not sure there's too much point in running it twice anyways...

Jenkinsfile
```diff
@@ -62,39 +87,52 @@ for(b = 0; b < builderNodes.size(); b++) {
...
+                    }
+                }
+            }
+
+            stage("Gathering test results (" + env.NODE_NAME + ")") {
+                junit '**/*_test_results.xml'
+            }
+
+            // Jamie's machine doesn't have clang-tidy installed, but we probably don't need to run this on multiple machines anyhow...
+            if(env.NODE_NAME == "INF900802") {
```
neworderofjamie, line 125: you could make this a little nicer by testing:
```java
if("clang_tidy" in nodeLabel) {
```
(I added the clang_tidy tag to your machine on Jenkins)

bin/run_clang_tidy_check
```diff
@@ -0,0 +1,128 @@
...
+    proc = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stdout = proc.communicate()[0].decode('utf-8')
+    if stdout != '':
+        print(stdout)
+
+if __name__ == '__main__':
+    all_json = []
+    all_files = set()
+    exclude_path = os.path.join(os.getcwd(), 'third_party')
+    for root, _, files in os.walk('.'):
```
neworderofjamie, line 97: I think you could get rid of quite a lot of this logic with ``glob.glob("**/compile_commands.json", recursive=True``

alexdewar: So you can! I should have guessed there'd be a more pythonic way to do that...

neworderofjamie: Hah, for any given piece of code and a personal view on pythonicness, there _always_ is....

alexdewar:
All done! Thanks for the feedback. If we can get this merged then I'll be able to run the new tests over #210 ...

PR #217, opened 2021-03-05
Option to record image databases as video files

alexdewar:
This is based on #212, so we can either merge that first, or just deal with those (small) changes here.

This PR adds support for recording route-type image databases as video files, which is much more space and IO efficient. (We're probably going to hit our git LFS quota again sometime fairly soon...) The idea would be that e.g. a large dataset recorded with the RC car could be uploaded in ``MP4`` format and then anyone wishing to use it for data analysis would just use the ``image_database_unwrapper`` tool to extract and process it first.

I've also converted over @tenxlenx's ``dataset_recorder`` to use this new functionality and fixed a couple of issues along the way:
* I removed the separate image-reading thread as it's not doing anything useful
* I removed the sleep we were doing inside the main loop, because it's just going to pointlessly slow down writing (@tenxlenx *This* is presumably the reason you weren't able to record at a consistent 15 fps)

@tenxlenx Would you mind testing this branch for me when you get a chance?


alexdewar:
Ok so now I've stopped fiddling with it :laughing: so it should be good to review...

alexdewar:
Hi both,

If you get a chance to have a look at this at some point that'd be great! It's not super urgent though...

include/navigation/image_database.h
```diff
@@ -412,47 +479,118 @@ class ImageDatabase
...
+    hertz_t getFrameRate() const;
+
     //! Get the resolution of saved images
     cv::Size getResolution() const;
 
     //! Check if this database has any saved metadata (yet)
     bool hasMetadata() const;
 
+    template<class Func>
+    size_t forEachImage(const Func &func, bool greyscale = true) const
```
neworderofjamie, line 513: I'm not really that familiar with this bit of the code but might it be worth templating ``ImageDatabase`` on a ``FrameReader`` policy/strategy class like ``Recorder`` does rather that having a monolithic thing (or just derive off separate classes)?

projects/dataset_recorder/dataset_recorder.cc
```diff
@@ -0,0 +1,172 @@
...
+                 << "[" << maxTrials - numTrials << "/" << maxTrials << "]";
+            std::this_thread::sleep_for(1s);
+            continue;
+        }
+
+        if (gpsData.gpsQuality != GPS::GPSQuality::INVALID) {
+            LOGI << " we have a valid measurement";
+            break;
+        }
+    }
```
neworderofjamie, line 67: might it be worth putting this logic into either  a method in ``GPS::GPSData`` or a free function in ``GPS`` - it seems pretty standard to most of the code I've seen go by and it seems synonymous with functions like ``See3CAM_CU40::autoExposure``

projects/dataset_recorder/dataset_recorder.cc
```diff
@@ -0,0 +1,172 @@
...
+    // Use the system clock to get date (might be wrong!)
+    time_t rawtime;
+    time(&rawtime);
+    std::tm currentTime = *localtime(&rawtime);
+
+    /*
+     * Use GPS to get time, because Jetsons don't remember the system time
+     * across boots.
+     */
+    currentTime.tm_hour = gpsData.time.hour;
```
neworderofjamie, line 83: Hah, "just use a GPS module" seems like the solution to the Jetson clock woes

projects/dataset_recorder/dataset_recorder.cc
```diff
@@ -0,0 +1,172 @@
...
+                                                   "Speed",
+                                                   "Steering angle [degrees]",
+                                                   "UTM zone",
+                                                   "GPS quality",
+                                                   "Timestamp [ms]");
+
+    cv::Mat frame;
+    catcher.trapSignals();
+    sw.start();
+    for (auto timestamp = 0_ms; timestamp < runTime; timestamp = sw.elapsed()) {
```
neworderofjamie, line 112: to my mind timesteps are integers - maybe rename this to ``time``

neworderofjamie (commented):
All looks good but, at a higher level, I do have some concerns about using video for image databases and subsequent image-matching type algorithms. Aside from at periodic "i-frames", most video formats don't store actual frames. Are sequences of low-res jpegs really problematic in the world of 32gb SD cards?

alexdewar:
> All looks good but, at a higher level, I do have some concerns about using video for image databases and subsequent image-matching type algorithms. Aside from at periodic "i-frames", most video formats don't store actual frames. Are sequences of low-res jpegs really problematic in the world of 32gb SD cards?

Thanks for looking at this! The amount of storage it takes up on the SD card is not *such* a big issue (although the Jetson doesn't seem to have a whole lot of available internal storage once all the NVIDIA crap is installed on there), but it is more annoying for uploading/downloading to GitHub. It should also make it faster and less CPU intensive to save the database to the SD card in the first place, as it will be less IO intensive and the GPU *should* be leveraged (which it won't be for single JPEGs). The idea would be that the raw route databases could be saved in this format, but we would almost certainly want to turn it into a series of unwrapped JPEGs before we actually do anything with it. So I've just added enough support that we can view and unwrap video-type databases with our existing tools. I definitely wouldn't want to use MP4s everywhere for image databases and it would be really silly for e.g. a simulated environment.

neworderofjamie:
Oh yeah the internal storage is pretty minimal - I _think_ all our carrier boards have SD card slots though. Lots of small files is indeed a bit filesystem-crushing but maybe using MJPEG compression rather than MPEG for the videos would be a nice, easy solution. I'm also not convinced OpenCV uses the Jetson's hardware MPEG decoder - I think you would have to build OpenCV with GStreamer support and I've never had a great deal of success with that.

include/navigation/image_database.h
```diff
@@ -159,22 +185,27 @@ class ImageDatabase
             imageDatabase.m_IsRoute = isRoute;
 
             // Get current date and time
-            std::time_t now = std::time(nullptr);
-            char timeStr[sizeof("0000-00-00 00:00:00")];
-            BOB_ASSERT(0 != std::strftime(timeStr, sizeof(timeStr), "%F %T", std::localtime(&now)));
+            const auto &creationTime = imageDatabase.getCreationTime();
+            std::stringstream ss;
+            ss << std::setfill('0')
```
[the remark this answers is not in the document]
alexdewar: No. The problem is that the Jetson's clock resets with every boot, so we want to be able to use the time given by the GPS instead. So I refactored ``ImageDatabase`` so you can pass in your own ``creationTime``.

include/navigation/image_database.h
alexdewar: I think you're right. In fact, I think it'd make sense to split out the reading and writing functionality into separate classes too by essentially turning the ``Recorder`` types into ``ImageDatabaseWriter``s. I'll have a go at refactoring

alexdewar: Hmm... The problem is that we don't know the database type at load time and we want to be able to load either a video-type or multi-file-type database transparently. We *could* do this by making ``forEachImage`` virtual in some base class, but then we wouldn't be able to have ``Func`` as a template parameter. We could pass in a ``std::function`` instead and I think in this case the performance hit wouldn't be too much of a problem. What do you think?

projects/dataset_recorder/dataset_recorder.cc
alexdewar: Good point

projects/dataset_recorder/dataset_recorder.cc
alexdewar: Ok cool

alexdewar:
> Oh yeah the internal storage is pretty minimal - I _think_ all our carrier boards have SD card slots though. Lots of small files is indeed a bit filesystem-crushing but maybe using MJPEG compression rather than MPEG for the videos would be a nice, easy solution. I'm also not convinced OpenCV uses the Jetson's hardware MPEG decoder - I think you would have to build OpenCV with GStreamer support and I've never had a great deal of success with that.

Though I *think* OpenCV will use the ffmpeg libraries if GStreamer isn't present and those should also make use of the GPU. MJPEG is in general a nice simple format to use, but we wouldn't see any space savings over just having multiple JPEG files, which seems like a shame. I don't really want to spend hours uploading/downloading to GitHub for every dataset ideally.

projects/dataset_recorder/dataset_recorder.cc
alexdewar: Hmm... maybe it's better to address GPS issues in #213 ? I'd still like to merge that at some point, but it's pending testing by @tenxlenx .

neworderofjamie:
Oh yes! FFMPEG does indeed support NVIDIA GPU acceleration - I _bet_ you have to compile from source to get it though 😢

neworderofjamie:
Also, I would suspect that the problem with JPEG datasets and git (LFS I assume) is lots of small files (which MJPEG would address) rather than the actual size.

include/navigation/image_database.h
neworderofjamie: Ahh ok, in that case, it's totally fine as is

projects/dataset_recorder/dataset_recorder.cc
neworderofjamie: Sure

alexdewar:
It seems like you have to use GStreamer to get hardware acceleration on Jetsons :cry:: https://forums.developer.nvidia.com/t/ffmpeg-using-hardware-gpu-cuda/72312

I *think* this should already be installed on our Jetsons though and optimistically CMake might have noticed it was installed and linked OpenCV against it accordingly.

alexdewar:
I'd at least like to have the *option* to use MP4 because I do already have some videos in that format (e.g. from collaborators). I feel like it would also be an appropriate choice for the kind of datasets @tenxlenx is collecting, but we perhaps could give users the option? Maybe we shouldn't artificially restrict users to MP4.

alexdewar:
On another note, have you had a chance to try this on the RC car @tenxlenx?

neworderofjamie:
I think giving users a choice of codec would be a good solution

alexdewar:
Cool. I'll do that then

alexdewar:
Hmm... I have tried implementing that but OpenCV seems to choke on using MJPEG as a format :cry:. It *does* work if you explicitly set GStreamer to be the backend, but I'm not sure if we generally compile in support for that when building OpenCV on Jetsons...?

We probs still should have the option to let users pick their own codec in any case, but it might be that getting MJPEG to work is more trouble than it's worth.

neworderofjamie:
What container format are you trying to write your video in? [this](https://www.pyimagesearch.com/2016/02/22/writing-to-video-with-opencv/) suggests MJPEG maybe only works with AVI containers (sorta makes sense mp4 or mpeg wouldn't work tbh)

alexdewar:
That did the trick! I was just writing them as ``.mjpg`` files, but I'm not sure if that's a proper format (though GStreamer seems to accept it...).

So I've now made the codec optional, but defaulting to MJPEG.

include/navigation/image_database.h
```diff
@@ -102,18 +102,18 @@ class ImageDatabase
     class ImageFileWriter
       : public FrameWriter {
     public:
-        ImageFileWriter(const ImageDatabase &) {}
-        void setImageFormat(std::string);
+        ImageFileWriter(const ImageDatabase &, std::string imageFormat);
```
neworderofjamie, line 105: ``const std::string&`` maybe?

alexdewar: I move it into ``m_ImageFormat`` so we can potentially avoid a copy (not that it'll matter with the 3-char strings we're using here anyway!).

include/navigation/image_database.h
```diff
@@ -510,57 +438,63 @@ class ImageDatabase
...
                 try {
-                    func(i, m_Entries[i].load(greyscale));
-                } catch (std::exception &e) {
-                    LOGE << "Error occurred in forEachImage: " << e.what();
-
-                    #pragma omp atomic
-                    errors++;
+                    func(i, m_Entries[i * frameSkip].load(greyscale));
+                } catch (...) {
+                    eptr = std::current_exception();
```
neworderofjamie, line 464: woah! I never knew you could do this 😨

alexdewar: Neat huh? It's handy for punting exceptions between threads (like we do with ``BackgroundExceptionCatcher``).

include/navigation/image_database.h
```diff
@@ -510,57 +438,63 @@ class ImageDatabase
...
-
-                if (greyscale) {
-                    cv::cvtColor(img, img, cv::COLOR_BGR2GRAY);
-                }
+        for (size_t i = 0; i < m_Entries.size() / frameSkip; i++) {
+            /*
+             * It is possible to explicitly jump to a given frame with OpenCV,
+             * but that turns out to be reeeeeeeaaaally slow.
+             */
+            for (size_t j = 1; j < frameSkip; j++) {
```
neworderofjamie, line 487: assuming frameSkip is relatively small, I'd hypothesise this could be symptomatic of my concerns with MPG. i.e. a potential skip implementation might be:

1. Goto i-frame before desired frame (might be way behind current frame)
2. Apply all subsequent p-frames

Whereas, for small frameSkip, your code would just apply a few (mostly) p-frames
Whereas

src/navigation/image_database.cc
```diff
@@ -132,6 +237,9 @@ ImageDatabase::ImageDatabase(const std::tm *creationTime,
                              bool overwrite)
   : m_Path{ std::move(databasePath) }
 {
+    memset(&m_CreationTime, 0, sizeof(m_CreationTime));
```
neworderofjamie, line 240: I think ``m_CreationTime{{0}}`` in the member initialiser list will achieve the same

neworderofjamie (approved):
This looks really cool - sorry for my video format pedantry 😄

include/navigation/image_database.h
alexdewar: That's a good idea, but unfortunately it seems like it can't be achieved with OpenCV :disappointed:. You *can* figure out which frames are i-frames with ffmpeg but this doesn't seem to be exposed through OpenCV and writing some custom code for this is probably overkill. (It really seems like something OpenCV should do internally when you explicitly seek to a frame, but however they're doing it it's horribly slow.)

This isn't a particularly fast implementation in any case -- we could achieve some speedup by e.g. reading frames on one thread and doing the processing on another -- but I only really want to be able to convert these databases to the "multiple image file" type anyway (e.g. when we unwrap them), so I'm not too worried about performance.

src/navigation/image_database.cc
alexdewar: Good point!

alexdewar:
I've made some fixes and changed the code in this branch to use tbb. I've dropped the changes to the ``dataset_recorder`` until we can test them on the actual robot (after which we can do a ``git revert 638779d``).

neworderofjamie:
Did you mean to push the docs change (I guess they should live on github io)? My browser really isn't happy with the 5000 file review...

alexdewar:
Ohh that's weird. I didn't push a docs change to this branch per se but I did merge in ``master`` so I guess github is treating the docs as changed... I don't think I've made any particularly radical changes so feel free not to review it if it's a faff.

src/navigation/image_database.cc
```diff
@@ -744,20 +784,20 @@ ImageDatabase::unwrap(const filesystem::path &destination,
...
-
     // Finally, unwrap all images and save to new folder
     forEachImage([&](size_t i, const cv::Mat &image) {
+        /*
+         * *If* forEachImage runs in parallel, then each thread will need its own
+         * copy of these scratch variables.
+         */
+        static thread_local cv::Mat unwrapped;
+        static thread_local std::string outPath;
+
```
neworderofjamie, line 795: ouch - that was a nasty bug

neworderofjamie (approved):
Ahh ok - I filtered it down and it all looks good

alexdewar:
Cool. Thanks!

PR #219, opened 2021-05-10
Extra checks for experiment-grade code

alexdewar:
Hi! This idea was (sort of) suggested by @aphilippides a while ago and I thought it might be useful.

Essentially I've just added a few more checks to CMake for code we mark as ``IS_EXPERIMENT=TRUE``, which ideally we'd want to do for any simulation/data collection programs where the data is going to be published somewhere.

Currently the main checks are:
  * Whether there are uncommitted changes in the git working tree
  * Whether the current git commit is in the master branch

If these checks fail, you get a big red error message whenever you start your program (though otherwise it should run as normal).

This is mostly just to catch the case where e.g. you've accidentally checked out the wrong branch before you run your data collection program.

### DIFF
--- a/cmake/bob_robotics.cmake
+++ b/cmake/bob_robotics.cmake
@@ -15,7 +15,7 @@ macro(BoB_project)
     # Parse input args
     include(CMakeParseArguments)
     cmake_parse_arguments(PARSED_ARGS
-                          "INCLUDE_GENN_USERPROJECTS;GENN_CPU_ONLY"
+                          "INCLUDE_GENN_USERPROJECTS;GENN_CPU_ONLY;IS_EXPERIMENT"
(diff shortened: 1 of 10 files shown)


alexdewar:
> Hah, I feel a lot of my 'experiments' would result in angry warnings...
> 
Mine too :laughing:, though I figured that I'd rather have the warning and then have the option of ignoring than not notice that I was doing something daft at all...


> Only real question I have is whether all the hardcore cmakary works on the version of cmake on our Ubuntu 16.04 Jetsons (at some point we should update some/all to the new Ubuntu 18-based Jetpack)

I've just checked and things do still build with Ubuntu 16.04 (in docker), so I think we should be fine. I don't think I've used any fancy new CMake features in any of the new code.

If we hit issues because of Ubuntu 16.04's old version of CMake (I have in the past), there is also the option of just building it from source rather than reflashing the Jetsons.

PR #223, opened 2021-05-26
Fix OpenMP support for CMake < 3.9

alexdewar:
This should fix the OpenMP issue you were having on the Jetson. I've tested in docker but I suppose we should probably test it on an actual Jetson just to be sure...

### DIFF
--- a/cmake/bob_robotics.cmake
+++ b/cmake/bob_robotics.cmake
@@ -337,6 +337,16 @@ macro(always_included_packages)
             macos_find_homebrew_openmp()
         endif()
         find_package(OpenMP QUIET)
+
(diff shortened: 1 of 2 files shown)


neworderofjamie:
Seems like it's working on this TX2 - NICE!
```
16:19:23.567 [BoBRobotics::Navigation::VisualNavigationBase::VisualNavigationBase@29] Running on 4 threads
```

neworderofjamie (approved):
Yuck but seems to work

neworderofjamie:
I'll tell you tomorrow - currently it's sitting on my desk so can't robot around

PR #225, opened 2021-05-26
Removed trailing whitespace from TEGRA_CHIP_ID

### LINKED ISSUE #224: Numpy issue with cmake 3.5.1
When running cmake I see:
```
-- Found PythonLibs: /usr/lib/aarch64-linux-gnu/libpython2.7.so (found version "2.7.12") 
WARNING: Could not find numpy; will build without
```
But when I run make I get:
```
BoB/modules/viz/CMakeFiles/bob_viz.dir/flags.make:8: *** missing separator. Stop.
```
Looking at the makefile it looks like the -DWITHOUT_NUMPY is preceded by a newline. The code in  matplotlibcpp.cmake _looks_ ok so no idea why ``add_definitions`` is doing this.

neworderofjamie:
Because the file ended in a carriage return, this added a line break into the makefile it got written to (REALLY cmake...) which broke any subsequent ``add_definitions`` calls.

Fixes #224

### DIFF
--- a/cmake/bob_robotics.cmake
+++ b/cmake/bob_robotics.cmake
@@ -534,9 +534,10 @@ macro(BoB_build)
     endif()
 
     # Different Jetson devices have different user-facing I2C interfaces
-    # so read the chip ID and add preprocessor macro
+    # so read the chip ID and add preprocessor macro, stripping trailing whitespace (carriage return)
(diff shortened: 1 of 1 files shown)


alexdewar (approved):
Looks sensible to me...

CMake in general is pretty crap at dealing with newlines at the end of strings. I've ended up adding a few ``STRIP``s to work round things before.

Issue #221, opened 2021-05-18
``Navigation::InfoMax`` currently silently ignores the specified image mask


alexdewar:
This is *potentially* a problem for e.g. the ``snapshot_bot`` code as we are tacitly assuming that the mask is being applied. Note that, if anything, it makes the problem *harder* for the algorithm, so we shouldn't worry that we've been "cheating".

The fix is conceptually very easy (we just need to select only the unmasked pixels for training/testing images), so I'll see if I can sort it while I'm refactoring the navigation code.

alexdewar:
Tagging people FYI: @tnowotny @aphilippides @neworderofjamie @tenxlenx 

NB this is the code we used in e.g. the Alife paper.

neworderofjamie:
I may well be wrong but I don't _think_ this is correct. Wouldn't that be equivalent to providing a black input to the masked pixels (which isn't the same thing)?

alexdewar:
Wait, you mean the fix (a06fdc1)?

In the end, the way I did it was to select only the unmasked pixels as inputs and reduced the number of inputs to the network accordingly. I *think* this is the right way to do it.

alexdewar:
> Wait, you mean the fix ([a06fdc1](https://github.com/BrainsOnBoard/bob_robotics/commit/a06fdc1bcfc196f1cdb86db489041f91a71ac22b))?
> 
> In the end, the way I did it was to select only the unmasked pixels as inputs and reduced the number of inputs to the network accordingly. I _think_ this is the right way to do it.

Ok, on reflection it isn't :laughing:. It gets complicated when a masked image is being rotated (e.g. with an RIDF).

I suppose we probably ought to additionally zero out the masked pixels in the current view. I think we should still drop the inputs to the weights matrix corresponding to the masked pixels in the unrotated mask (as by definition the snapshots aren't providing inputs here).

What do you think, @aphilippides ?

alexdewar:
More annoyingly, we have issue #222

tnowotny:
would it be worth to get together when Jamie is back and have an in-depth discussion about this (I am not sure I can make any meaningful contributions without some prepping from you guys)?

alexdewar:
It's probably not worth a whole meeting by itself but it'd be worth having a chat about I think.

tnowotny:
we should get together anyway to take stock of some things and plan the summer - but we should wait until Jamie is back. Next Wed or Friday I would suggest.

[...]

PR #226, opened 2021-05-27
Navigation code updates

### LINKED ISSUE #222: Broken image masking code in for perfect memory algos
It seems that there's a lazy evaluation bug in the loop where we're checking whether pixels are masked or not: https://github.com/BrainsOnBoard/bob_robotics/blob/fec6550cd8b3af93409a0ac10a31175b47dc50b9/include/navigation/perfect_memory_store_raw.h#L96

If ``*imageMaskPtr`` is non-zero, then ``snapshotMaskPtr`` won't be incremented.

I introduced it in 1d672712 and the code was later copied over to ``PerfectMemoryStore::RawImage`` during a refactor.

I only just noticed it when I was writing some new unit tests (which is probably a good argument for doing test-driven development...).

The fix is trivial and I'll sort it in the branch I'm working on now.

alexdewar:
I initially started off by working on a correlation coefficient differencer, as we discussed, but ended up reworking the code for the other differencers a bit for correctness/tidiness. I also wanted to bundle more scratch data into the differencer classes, partly so that there is less state in the algo classes, but also because that means that only the differencer object needs to be made OpenMP safe, hopefully reducing the chance of multithreading-related snafus. In adding some tests I found some issues to do with masking (#221 and #222) and so I did some more fixing/test-adding for that. I've fixed #222 in this PR, but I feel like #221 should be its own PR.

So apologies if there are too many changes going on here, but I'm hoping that it shouldn't be too controversial. Really I've just tried to make the code tidier and safer -- I haven't radically altered any APIs.

One bonus is that in changing the differencers to use OpenCV's built-in ``cv::sum()`` and ``cv::mean()``, performance is *way* better (at least on my machine):
benchmark_cmp

Change summary:
* Refactor differencers (as discussed)
* Add a "correlation coefficient differencer" -- it's somewhat slow (see benchmark data above; probs because it uses floating-point numbers) but might be useful as a comparator
* Encapsulate mask-related functionality in a class ``ImgProc::Mask``, which gives added type safety and means that masks are always properly checked for correctness -- also makes it easier to test
* Add more tests for navigation including for your windowing code

Things left for future work:
* Refactor out the ``InSilicoRotater`` class -- we only have one rotater now so it's unnecessary
* Change ``getHeading()`` et al. to return data as ``struct``s cf. ``std::tuple``s as discussed
* Add tests for ``DynamicBestMatchGradient`` -- it's not obvious to me how to do this at the moment

(If you have any ideas about the last one, I'm happy to do it in this PR. I feel it'd be nice to have tests for this code too before we do experiments using it. It's just that the current sole user is in ``projects/ardin_mb`` and I didn't want to go pulling bits out of there.)


include/navigation/differencers.h
```diff
@@ -87,28 +94,101 @@ class AbsDiff
...
     {
-        return sqrt(sum / n);
-    }
+    public:
+        float operator()(cv::InputArray &src1, cv::InputArray &src2,
+                         const ImgProc::Mask &mask1 = {},
+                         const ImgProc::Mask &mask2 = {})
+        {
+            // Don't support masks for now
+            BOB_ASSERT(mask1.empty() && mask2.empty());
```
alexdewar, line 164: Hmm, I forgot to do this part... I'll have a go now.

include/navigation/differencers.h
```diff
@@ -87,28 +94,101 @@ class AbsDiff
...
+                auto beg = m.datastart;
+                auto end = m.dataend;
+                return std::all_of(beg + 1, end,
+                                   [beg](uint8_t val) { return val == *beg; });
+            };
+            if (allSame(src1) || allSame(src2)) {
+                throw std::invalid_argument("Vectors src1 and src2 must have "
+                                            "more than one unique value (e.g. "
+                                            "they cannot be all zeros)");
+            }
```
neworderofjamie, line 178: This check seems a bit heavyweight for something to do every frame. Not really sure what the alternative is though - accepting that if you put garbage in you get garbage out would be my vote

include/navigation/perfect_memory_store_hog.h
```diff
@@ -78,25 +80,26 @@ class HOG
...
+                                 const ImgProc::Mask &) const
     {
         BOB_ASSERT(imageMask.empty());
 
+        /*
+         * Workaround for a really longstanding gcc bug:
+         *      https://gcc.gnu.org/bugzilla/show_bug.cgi?id=27557
+         */
+        static thread_local typename Differencer::template Internal<std::vector<float>> differencer;
+
```
neworderofjamie, line 95: This is definitely for another PR but, with all these OpenMP hacks (before we even start on Windows), might it not be worth just using ``std::thread``? I parallelised some slow bits of GeNN the other day and it was _slightly_ more painful than OpenMP but at least explicit as to what was going on....

src/imgproc/mask.cc
```diff
@@ -0,0 +1,115 @@
...
+{
+    if (mask.empty()) {
+        // Clears mask
+        m_Mask = cv::Mat{};
+        return;
+    }
+
+    // The user has requested a specific size of mask
+    if (!size.empty()) {
+        cv::resize(mask, mask, size);
```
neworderofjamie, line 88: I think you might want to explicitly specify the ``cv::INTER_NEAREST`` interpolation mode if we're working with binary images

neworderofjamie: Also, I might _still_ be missing something but if you took a const reference here you could just specify ``m_Mask`` as the target for resize/convert?

tests/common.h
```diff
@@ -6,3 +6,38 @@
...
+            compare(m1(i, j), m2(i, j));
+        }
+    }
+}
+
+template<class T, class U>
+void compareFloatMatrices(const T &m1, const U &m2)
+{
+    const auto expect = [](float a, float b) {
+        EXPECT_FLOAT_EQ(a, b);
```
neworderofjamie, line 31: Having been swamped with billions of test errors after writing code like this for GeNN, I'd maybe thing on expecting the result of ``std::all_of`` or something

tests/differencers.cc
```diff
@@ -0,0 +1,146 @@
...
+        ASSERT_EQ(m1.type(), CV_8UC1);                                                \
+        ASSERT_EQ(m2.type(), CV_8UC1);                                                \
+        ASSERT_EQ(expected.type(), CV_8UC1);                                          \
+        ASSERT_EQ(m1.size(), m2.size());                                              \
+        bool doesMatch;                                                               \
+        float mean;                                                                   \
+        std::tie(doesMatch, mean) = compare<AbsDiff>(m1, m2, expected, mask1, mask2); \
+        EXPECT_TRUE(doesMatch);                                                       \
+        EXPECT_TRUE(mean == expectedMean);                                            \
+    } while (false);
```
neworderofjamie, line 25: The idiomatic version of the ``do{}while(false)`` macro trick leaves off the trailing semicolon meaning that it _needs_ to be called with a semicolon afterwards (like a _real_ function)

neworderofjamie (approved):
This looks really cool - just a couple of minor comments really

include/imgproc/mask.h
```diff
@@ -0,0 +1,55 @@
...
+{
+//! A class providing extra checks and type safety for image masks
+class Mask
+{
+public:
+    //! Create an empty mask (i.e. treat all pixels as unmasked)
+    Mask() = default;
+
+    //! Construct a mask based on a cv::Mat where zero pixels are treated as masked
+    explicit Mask(cv::Mat mask, const cv::Size &resizeTo = {});
```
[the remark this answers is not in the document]
alexdewar: I *think* this is the right way to pass objects into setters/constructors (there is actually a clang-tidy option to check for this which I didn't enable because it seems a bit overzealous).

The idea is that if you then move the object into the member variable, you can avoid a copy in the case that ``mask`` is passed as an rvalue.

include/navigation/differencers.h
alexdewar: This did occur to me too... I think the performance penalty shouldn't be too bad in general though as ``allSame()`` will return false when it hits upon the first non-matching pixel, which is likely to happen fairly early on.

An alternative would be to only do these additional checks in the case that the correlation coefficient comes out as zero (which seems to be what happens in this error case).

I'm a little concerned about relying on us not passing garbage in though as I can easily imagine a case where an all-black frame is passed in by mistake...

include/navigation/perfect_memory_store_hog.h
alexdewar: Another option would be Intel's tbb library which looks really nice: https://software.intel.com/content/www/us/en/develop/documentation/onetbb-documentation/top.html

Plus, it's a dependency for OpenCV so users will have it installed anyway.

src/imgproc/mask.cc
alexdewar: > I think you might want to explicitly specify the `cv::INTER_NEAREST` interpolation mode if we're working with binary images

Good point!

> Also, I might _still_ be missing something but if you took a const reference here you could just specify `m_Mask` as the target for resize/convert?

As we're ``std::move``ing ``mask`` at the end, effectively it doesn't make any difference, since we're operating on the same chunk of memory. It potentially saves us doing a separate allocation for the output of e.g. ``resize()``.

tests/common.h
alexdewar: Yeah, I've had similarly verbose error messages :grimacing:. The problem is that we're doing floating-point comparisons so we don't want to test for exact equality.

We could specify the desired precision manually and check that ``a - b < precision`` for all elements, but I'm a bit on the fence about this, because I think Google's default comparison for ``EXPECT_FLOAT_EQ()`` does something clever.

neworderofjamie: Good point and yes, I just looked at the source and it does indeed do _something_ clever 😄

tests/differencers.cc
alexdewar: Ah, good spot!

PR #227, opened 2021-05-28
Remove units from VICON data when logging during testing

neworderofjamie:
One of us fixed this during training but, somehow, this code path slipped through the net (cue much searching through old scripts for the right ``np.loadtxt`` magic to strip off suffixes)

### DIFF
--- a/projects/snapshot_bot/snapshot_bot.cc
+++ b/projects/snapshot_bot/snapshot_bot.cc
@@ -394,7 +394,7 @@ private:
                     const auto &attitude = pose.attitude();
 
                     // Write extra logging data
-                    m_LogFile << ", " << objectData.getFrameNumber() << ", " << position[0] << ", " << position[1] << ", " << position[2] << ", " << attitude[0] << ", " << attitude[1] << ", " << attitude[2];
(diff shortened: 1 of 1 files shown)


alexdewar (approved):
Oh weird! I wonder if something happened when I merged branches...

Looks good anyway.

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -394,7 +394,7 @@ class RobotFSM : FSM<State>::StateHandler
                     const auto &attitude = pose.attitude();
 
                     // Write extra logging data
-                    m_LogFile << ", " << objectData.getFrameNumber() << ", " << position[0] << ", " << position[1] << ", " << position[2] << ", " << attitude[0] << ", " << attitude[1] << ", " << attitude[2];
+                    m_LogFile << ", " << objectData.getFrameNumber() << ", " << position[0].value() << ", " << position[1].value() << ", " << position[2].value() << ", " << attitude[0].value() << ", " << attitude[1].value() << ", " << attitude[2].value();
```
alexdewar, line 397: I feel this could probably use a line break somewhere :laughing:

neworderofjamie: Fine point

neworderofjamie:
I think you fixed the training case ages ago but, probably because we haven't used this with the Vicon for a while, the testing bit never got fixed - I'm doubtful this is a merging issue.

PR #233, opened 2021-06-01
Two fixlets

alexdewar:
These should be in ``master`` but the first *may* also help your build error....


alexdewar:
Sorry, I only intended to include the last two commits. I've force-pushed over the top

neworderofjamie (approved):
I have no idea how that fixes my cmake problems but it does :smile:

PR #234, opened 2021-06-01
Image masks for ``CorrCoefficient`` differencer

alexdewar:
This is something I forgot to do in my last navigation-related PR, but it's a fairly simple change.

I also changed ``CorrCoefficient`` so that it could be used with HOG, for completeness.


include/navigation/differencers.h
```diff
@@ -136,6 +137,14 @@ class RMSDiff
     };
 };
 
+template<class T>
+bool allSame(cv::InputArray &arr)
+{
+    const cv::Mat_<T> m = arr.getMat();
+    return std::all_of(m.begin() + 1, m.end(),
+                       [&m](T val) { return val == *m.begin(); });
+}
```
neworderofjamie, line 146: Is it worth trying to use e.g. an OpenCV histogram for this? They tend to have more hard-core optimizations than the normal C++ compiler will attempt.

include/navigation/differencers.h
```diff
@@ -149,36 +158,45 @@ class RMSDiff
...
+            /*
+             * NB: We could easily support other mat types but these are the
+             * only ones we care about for now.
+             */
+            bool nonUniqueInputArray;
+            switch (src1.type()) {
+            case CV_8UC1:
+                nonUniqueInputArray = allSame<uint8_t>(src1) || allSame<uint8_t>(src2);
+                break;
+            case CV_32FC1:
```
neworderofjamie, line 184: is this to handle HOG features? I seem to remember raw images are always ``CV_8UC1``

neworderofjamie (approved):

include/navigation/differencers.h
alexdewar: That's true. How would calculating a histogram instead work though? I feel I'm missing something

include/navigation/differencers.h
alexdewar: Yes it is

alexdewar:
Tests are now passing, but only because I've disabled some tests :cry:. Essentially it seems like there's a bug in OpenCV v4.5.1 and earlier which means that image masks won't work with correlation coefficients, so I've just disabled this functionality at compile time: b5db1f5

What do you think, @neworderofjamie ? We don't *need* this functionality all the time and the easiest solution is probably just to rebuild OpenCV if we find that we do at some point...

neworderofjamie:
That seems a perfectly acceptable solution. As you say, if we need it, we can build a suitable OpenCV.

alexdewar:
Cool. Thanks!

PR #235, opened 2021-06-03
Basic Opteran ODK2 integration

neworderofjamie:
This PR integrates the Opteran ODK2 as an ``BoBRobotics::Video::Input`` which returns pitch and roll stabilized 256x65 RGB images and an IMU (there's currently no interface for this, but the signature for ``ODK2::getEulerAngles`` matches ``BN055::getEulerAngles``. The ODK2 streams data at a high frame-rate over UDP so this code uses a background thread to ensure data isn't lost. Aside from the issues still todo, this should be a drop in replacement for the PixPro + BN055 in the mecanum robot. There should also be a firmware update at the end of the month which provides high-level collision avoidance which I will expose in a subsequent PR.

When I have a suitable robot to mount the ODK2 on I will integrate it with the stone_cx project and snapshot_projects in subsequent PRs after we figure out a solution to the "dynamic masking" problem discussed below.


src/video/odk2/odk2.cc
```diff
@@ -0,0 +1,191 @@
...
+                // Create pixel into OpenCV image
+                const cv::Vec3b pixel(m_RawFrameBuf.image[fourpiIdx * 3],
+                                      m_RawFrameBuf.image[(fourpiIdx * 3) + 1],
+                                      m_RawFrameBuf.image[(fourpiIdx * 3) + 2]);
+
+                // Write pixel to OpenCV image
+                *outPixel++ = pixel;
+
+                // If pixel is black, clear mask otherwise set
+                *outMask++ = (pixel[0] == 0 && pixel[1] == 0 && pixel[2] == 0) ? 0 : 255;
```
alexdewar, line 105: I'll stick my comments about the image masking here because it's as good a place as any!

Out of interest, are you currently using this mask somewhere? The reason I ask is that if the goal is to integrate it with the navigation code then I feel like it's not the right approach. One problem is that the navigation algos currently assume that the mask is set once (before training begins) and then remains unchanged, whereas I guess ``m_Mask`` will change with each frame (though probably only slightly).

I agree that it might be nice to add a virtual function to ``Video::Input`` for this case. How about ``loadMask()`` or something? It could return an ``ImgProc::Mask`` (not a reference) and the name sort of implies that an object is being created... In general my idea for the ``ImgProc::Mask`` class was that the underlying ``cv::Mat`` is mostly immutable: there's deliberately no way to get a non-const reference to it and if you want to update it you have to call ``set()`` again which ensures that a) it is of type ``CV_8UC1`` and b) all the values are either 0 or 255 (or it's empty). Then in theory we can just pass the object around as we like without having to do any checks where we're using it.

alexdewar: I've just realised something: does the mask change on every frame because of the devkit's built-in image stabilisation? If so then this is probably a harder problem than I thought :grimacing:

alexdewar (changes requested):
I've added a few comments but it looks cool!

I think the only thing that really needs changing is the handling of errors on the background thread for ``Video::ODK2``.

The image masking stuff needs more thought if we're going to integrate it with the navigation code but we could always deal with that later.

src/video/odk2/odk2.cc
```diff
@@ -0,0 +1,191 @@
...
+        using namespace units::literals;
+
+        // Update last timestep
+        m_LastRawFrameTimestep = m_RawFrameBuf.timestampUs;
+
+        // Extract quaternion components from frame
+        const float qw = m_RawFrameBuf.quaternion.w;
+        const float qx = m_RawFrameBuf.quaternion.x;
+        const float qy = m_RawFrameBuf.quaternion.y;
+        const float qz = m_RawFrameBuf.quaternion.z;
```
alexdewar, line 73: I'd probably just do ``const auto &q = m_RawFrameBuf.quaternion;``

src/video/odk2/odk2.cc
```diff
@@ -0,0 +1,191 @@
...
+    const radian_t heading(std::atan2(2.0 * ((qw * qz) + (qx * qy)),
+                                      1.0 - (2.0 * ((qy * qy) + (qz * qz)))));
+
+    // Calculate roll from quaternion
+    const radian_t roll(std::atan2(2.0 * ((qw * qx) + (qy * qz)),
+                                   1.0 - (2.0 * ((qx * qx) + (qy * qy)))));
+
+    // Calculate pitch from quaternion
+    const double sinp = 2.0 * ((qw * qy) - (qz * qx));
+    const radian_t pitch((std::abs(sinp) >= 1.0) ? std::copysign(M_PI / 2.0, sinp) : std::asin(sinp));
```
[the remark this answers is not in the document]
neworderofjamie: I have that memory too and the devkit driver does joyously support windows so what _is_ a better PI constant?

src/video/odk2/odk2.cc
alexdewar: An additional point: even if we do want to derive a mask for each frame based on where the black pixels are I think it would make sense not to do it in this loop, to keep things simpler. There should be a relatively terse way to do this calculation after the fact using some vectorised OpenCV methods.

src/video/odk2/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+cmake_minimum_required(VERSION 3.1)
+include(../../../cmake/bob_robotics.cmake)
+BoB_module(SOURCES odk2.cc
+           BOB_MODULES video
+           THIRD_PARTY development_kit_odk2)
```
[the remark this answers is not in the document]
neworderofjamie: It should work on Linux and Windows but not Mac (like me)

src/video/odk2/odk2.cc
neworderofjamie: It does change every frame because of the stabilisation hence doing this in the loop. I couldn't actually find a build in method which did it (``cv::compare`` was close but worked per-channel) - quite how ``cv::threshold`` worked with multi-channel wasn't clear...

neworderofjamie: my other thought was to add a new ``set`` method to ``ImgProc::Mask`` which did whatever thresholding/other operation is required to make a mask (as that seems a reasonably general piece of functionality)

src/video/odk2/odk2.cc
```diff
@@ -0,0 +1,191 @@
...
+
+    // To match BN055, we want to return in heading, roll, pitch
+    return {heading, roll, pitch};
+}
+//------------------------------------------------------------------------
+void ODK2::readThread()
```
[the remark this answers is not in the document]
neworderofjamie: If a wire's loose you'll get a ``DEVKIT_DRIVER_RX_TIMEOUT`` which just logs a warning so, as BoB robotics generally uses exceptions for error handling I'll probably do that here. I feel it would be kinda nicer if the exceptions were returned on the main thread via ``readFrame``/``getEulerAngles`` rather than at a global level but that would be inconsistent with the rest of the library so I'll use the ``BackgroundExceptionCatcher``

src/video/odk2/CMakeLists.txt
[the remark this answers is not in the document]
alexdewar: Ok cool. ``PLATFORMS linux windows`` will also work...

src/video/odk2/odk2.cc
neworderofjamie: (maybe ``cv::inRange`` can do this actually)

alexdewar: I've had a thought about how to integrate this with the perfect memory code: we could just store a copy of the image mask with each stored snapshot instead of having a single mask for all of them. This is less awful than it sounds at first because ``cv::Mat``s are reference-counted and CoW so for the common case where the mask remains unchanged throughout training, all the stored masks would point to the same heap data.

neworderofjamie: or we could make the perfect memory have the option of treating (0,0,0) as mask and handle it all automagically.

neworderofjamie: I think from our previous discussions, as there's no biases, (0,0,0) already has no effect in InfoMax

alexdewar: I think it's ok to treat the ``ImgProc::Mask`` as mutable in this case and come up with some way to fiddle with its internal data structure. I just wanted to make it hard to do this!

One way of doing it would be to derive a ``MutableMask`` class with a ``getMutable()`` method or something. Then we just make sure the return type of ``getMask()`` is the immutable ``ImgProc::Mask`` type.

alexdewar: > or we could make the perfect memory have the option of treating (0,0,0) as mask and handle it all automagically.

Currently these methods only accept greyscale images so this would require some more hackery. I feel like being explicit about what we want our masks to be is probably the least error prone way of doing it. I've had enough trouble getting this explicit way to work correctly and I'd be worried if we have some implicit magic going on it'll make it really difficult to find bugs.

neworderofjamie: eugh, that is a good point. It could be triggered by a ``CV_8UC2`` input type where the second channel is mask?

neworderofjamie: that way, I think the change could be made at a pretty high level (in the perfect memory class itself)

neworderofjamie: So we'd implement:

1. A flag in the ODK2 constructor which made ``ODK2::readGreyscaleFrame`` return ``CV_8UC2`` with gray + mask.
2. Make sure the raw store is capable of storing ``CV_8UC2``
3. Add asserts to ``PerfectMemory`` to prevent fixed masks and ``CV_8UC2`` images being used together.
4. Make a change whenever we call ``getMask()`` in ``PerfectMemory`` to use the second channel of the stored image/input image instead

alexdewar: I've just installed the GitHub bot on Slack... let's see if it notices this message

alexdewar (approved):
Looking good!

PR #236, opened 2021-06-08
Swap out OpenMP for TBB

alexdewar:
This is a fairly simple change to use Intel's TBB instead of OpenMP, which should be less of a faff to get working. It may mean that users have to do an ``apt install libtbb-dev`` but it's only a small dependency really.

This does also mean that the parallel navigation code now works on Windows too.

### DIFF
--- a/examples/infomax/infomax_route_example.cc
+++ b/examples/infomax/infomax_route_example.cc
@@ -30,15 +30,14 @@ using namespace BoBRobotics::Navigation;
 using namespace std::literals;
 namespace plt = matplotlibcpp;
 
-using namespace units::literals;
-using namespace units::math;
-
 using FloatType = float;
 using InfoMaxType = InfoMaxRotater<InSilicoRotater, FloatType>;
 
 void doTesting(const InfoMaxType &infomax, const std::vector<double> &x,
                const std::vector<double> &y, const std::vector<cv::Mat> &images)
 {
+    using namespace units::math;
+
     std::vector<double> u, v;
     {
         // Test network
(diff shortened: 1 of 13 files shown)


examples/infomax/infomax_route_example.cc
```diff
@@ -30,27 +30,24 @@ using namespace BoBRobotics::Navigation;
...
         LOGI << "Testing network...";
         Timer<> t{ "Network testing took: " };
         for (const auto &image : images) {
             // Get heading and convert to vector
-            const units::angle::radian_t heading = std::get<0>(infomax.getHeading(image));
-            double sinx, cosx;
-            sincos(heading.value(), &sinx, &cosx);
-            u.push_back(cosx);
-            v.push_back(sinx);
+            const auto heading = std::get<0>(infomax.getHeading(image));
```
neworderofjamie, line 48: I feel ``sincos`` could be posix rather than standard C++

neworderofjamie, line 48: I feel ``sincos`` could be posix rather than standard C++

examples/infomax/infomax_route_example.cc
```diff
@@ -30,27 +30,24 @@ using namespace BoBRobotics::Navigation;
...
+            u.push_back(cos(heading));
+            v.push_back(sin(heading));
```
neworderofjamie, line 50: I feel ``sincos`` could be posix rather than standard C++

include/navigation/insilico_rotater.h
```diff
@@ -55,21 +55,19 @@ struct InSilicoRotater
...
-                    m_Mask.roll(scratchMask, index);
-
-                    func(scratchImage, scratchMask, distance(m_BeginRoll, i));
-                }
-            }
+            static thread_local cv::Mat scratchImage;
+            static thread_local ImgProc::Mask scratchMask;
+
+            tbb::parallel_for(tbb::blocked_range<size_t>(0, numRotations()),
+                [&](const auto &r) {
```
neworderofjamie, line 62: this is so much nicer than the pragma horror!

neworderofjamie (approved):
This looks so much nicer

alexdewar:
Yeah, it's a really nice library!

Would you mind installing ``libtbb-dev`` on your desktop machine? I think that's the reason the CI is currently failing

neworderofjamie:
I'm afraid it's already installed:
```libtbb-dev is already the newest version (2017~U7-8).```

alexdewar:
Arg! And so the debugging begins...

Issue #238, opened 2021-06-11
Add support for non-static image masks


alexdewar:
Currently, the visual navigation algorithms make the assumption that the image mask will essentially remain constant throughout the object's lifetime (i.e. during both training and testing phases). This works well for the common case where we are trying to mask out a part of the robot visible by the camera (e.g. an antenna), but notably doesn't work for the ``Video::ODK2`` camera, whose mask will change with each frame due to the on-board image stabilisation.

Hence, we need to update the navigation algos to be able to accept an ``ImgProc::Mask`` along with image for the ``train()`` method and to use this information appropriately when doing testing.

@neworderofjamie has volunteered for this!

PR #239, opened 2021-06-11
Add image mask support to InfoMax algo

alexdewar:
Closes issue #221. I *think* I've done this in the right way, but I'm not totally convinced so it would be good to have some feedback...The way I decided to implement this was:

1. Zero out the masked pixels to the input layer (which should be fine as there's no bias) for training and testing
2. Exclude the elements in the output layer which correspond to the masked pixels from the decision function (i.e. taking the mean)

I'm not sure if 2. makes sense or not...

It would be good to get this PR fixed up and merged before we add support for non-static image masks (#238) so that we can also use InfoMax for our demonstrator/experiments.


neworderofjamie:
I don't think 2 makes sense as the output layer is all-to-all connected to the input layer so each output unit receives input from all neurons. I think that, in our configuration, the fact that the two layers have the same number of neurons is a coincidence rather than because there's any correspondence between input and output neurons. I _think_ 2 should be replaced by some sort of scaling factor derived from the maths...

alexdewar:
@aphilippides just phoned me to say more or less the same thing :laughing:. I guess I should actually got back to the paper and figure out how to do it properly... It does seem like it will involve a scaling factor

aphilippides:
Hi Both

Nice that we Agreed! I also think the scaling is not going to make much difference but I also think we could get Srathis to check this.

Can we maybe have a chat on Monday to discuss and also plan out the data we need to gather for papers?

Andy

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

alexdewar:
Yep! Good plan.

alexdewar:
It turns out this was the wrong approach. In any case this isn't going to get merged before I leave and I don't think there's anything worth keeping on this branch.

PR #240, opened 2021-06-12
Protect BackgroundExceptionCatcher::ExceptionPtr with std::mutex

alexdewar:
As pointed out by @neworderofjamie, std::exception_ptr is not atomic, so we need to protect it from race conditions. Figuring out how to do the correct compare/exchange magic with a std::atomic_bool turned out to be beyond me, so let's just use a mutex.


neworderofjamie (approved):
Definitely no need to worry about performance in exception handling code

PR #242, opened 2021-06-17
Workaround for compiler error on Visual Studio

neworderofjamie:
I can't quite understand why this was occurring - adding brackets doesn't help so I don't think it's some sort of horrible macro operator precedence issue. Maybe a legit compiler bug?


alexdewar (approved):
Weird!

LGTM

PR #243, opened 2021-06-17
Runtime fixes

neworderofjamie:
After getting it working, the Visual Studio Debug runtime found a few nasty issues which I've tried to fix here.


alexdewar (approved):
Did these come from VS warnings?

neworderofjamie:
From the Visual Studio debug runtime which is very slow as it checks iterator bounds etc and seemingly also that functions actually obey the rules they are supposed to!

alexdewar:
That's awesome! Really handy.

PR #244, opened 2021-06-17
Dynamic masking refactor

neworderofjamie:
In this PR I have:
- Removed the ``VisualNavigationBase`` class - the only remnant I am unsure about is the trainRoute helper which, for now, I have stuck in the perfect memory example (where it's used)
- Masks are now owned by the application rather than the ``PerfectMemory`` class and provided to ``test``, ``train``, ``getHeading`` etc as appropriate
- Added functionality to ``ImgProc::Mask`` to build masks from range of RGB values e.g. to mask all black pixels for ODK2.
- ``PerfectMemoryStore::RawImage`` now stores pairs of images and their masks - masks are not deep copies so if dynamic masks are to be used they need to be cloned before passing to ``train`` method
- ``InSilicoRotater`` is only sensible form of rotater so it is now hard-coded - allows us to slightly re-order parameters for ``InfoMax`` and ``PerfectMemory`` classes
- Updated Ardin MB, snapshot bot, examples and tests to use new functionality and APIs


alexdewar:
FYI it seems like now that my machine is back online the CI is failing on the master branch because of a bunch of warnings/errors, which is probably why this branch is failing too. I'm just debugging it now.

neworderofjamie:
ooh - looking forward to seeing what GCC 11 makes of GeNN 😟

include/navigation/infomax.h
```diff
@@ -61,29 +60,27 @@ class InfoMax : public VisualNavigationBase
     {}
 
     //------------------------------------------------------------------------
-    // VisualNavigationBase virtuals
+    // Public API
     //------------------------------------------------------------------------
-    virtual void train(const cv::Mat &image) override
+    void train(const cv::Mat &image, const ImgProc::Mask& = ImgProc::Mask{})
```
alexdewar, line 65: It's a bit weird to have default values for unused args, but it's harmless and we'll change this when we fix #221 anyway.

include/navigation/perfect_memory.h
```diff
@@ -115,29 +103,41 @@ class PerfectMemory : public VisualNavigationBase
         return {0, getNumSnapshots()};
     }
 
+    //! Get the resolution of images
+    const cv::Size &getUnwrapResolution() const { return m_UnwrapRes; }
+
 protected:
+    static const Window defaultWindow;
```
alexdewar, line 110: Minor nit: but I feel like this should probably be public as it might be handy for users.

I guess you could probably also make it ``constexpr``.

include/navigation/perfect_memory.h
```diff
@@ -115,29 +103,41 @@ class PerfectMemory : public VisualNavigationBase
...
     {
         const auto &unwrapRes = getUnwrapResolution();
         BOB_ASSERT(image.cols == unwrapRes.width);
         BOB_ASSERT(image.rows == unwrapRes.height);
         BOB_ASSERT(image.type() == CV_8UC1);
+        if(!mask.empty()) {
+            BOB_ASSERT(mask.get().cols == unwrapRes.width);
+            BOB_ASSERT(mask.get().rows == unwrapRes.height);
+            BOB_ASSERT(mask.get().type() == CV_8UC1);
+        }
```
alexdewar, line 139: You could replace this block with ``BOB_ASSERT(mask.isValid(unwrapRes));`` (which will also work if ``mask`` is empty).

You don't need to check the type as this will always be ``CV_8UC1`` as long as the API has been used properly.

alexdewar (approved):
Looking good! Finally we're free of ``VisualNavigationBase``

include/navigation/infomax.h
neworderofjamie: yeah - I left that looking like a loose end semi on purpose

include/navigation/perfect_memory.h
neworderofjamie: this was a bit of a gross hack around not being able to initialise default parameters with a non-static method - we don't really want users initialising windows to this, they should use ``fullWindow()`` instead. Sadly wouldn't we need C++ 17 for inline constants rather than gross static initialisation syntax?

examples/odk2/odk2.cc
```diff
@@ -1,39 +1,74 @@
...
         // Get euler angles from ODK2
-        const auto euler = odk2.getEulerAngles();
-        std::cout << "Orientation: (" << euler[0].value() << ", " << euler[1].value() << ", "<< euler[2].value() << ") degrees" << std::endl;
+        //const auto euler = odk2.getEulerAngles();
+        //std::cout << "Orientation: (" << euler[0].value() << ", " << euler[1].value() << ", "<< euler[2].value() << ") degrees" << std::endl;
 
         // Pump events and stop if escape is pressed
-        if(cv::waitKey(1) == 27) {
+        const int key = cv::waitKey(1);
+        if(key == 27) {
```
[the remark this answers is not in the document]
neworderofjamie: it **should** work on Windows so this is indeed a good point

examples/odk2/odk2.cc
```diff
@@ -1,39 +1,74 @@
...
+        cv::imshow("Mask", mask.get());
+
+        if(pm.getNumSnapshots() > 0) {
+            std::cout << "No mask:" << pm.test(cameraImageGrayscale) << ", mask:" << pmMask.test(cameraImageGrayscale, mask) << std::endl;
+        }
         // Get euler angles from ODK2
-        const auto euler = odk2.getEulerAngles();
-        std::cout << "Orientation: (" << euler[0].value() << ", " << euler[1].value() << ", "<< euler[2].value() << ") degrees" << std::endl;
+        //const auto euler = odk2.getEulerAngles();
+        //std::cout << "Orientation: (" << euler[0].value() << ", " << euler[1].value() << ", "<< euler[2].value() << ") degrees" << std::endl;
```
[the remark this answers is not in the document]
neworderofjamie: I think this is a stale diff

include/navigation/perfect_memory.h
alexdewar: True: you would need a separate definition for ``defaultWindow`` somewhere. So it's probably not worth it. Let's just make users use ``fullWindow()``.

neworderofjamie: actually, the whole ``defaultWindow`` thing was gross - checkout my last commit

PR #245, opened 2021-06-18
CI fixes and clang-tidy updates

alexdewar:
It seems like only one of the Jenkins errors/warnings we're getting on master is due to gcc v11; the others were seemingly being emitted by ``clang-tidy`` -- I guess because there's a new version with additional warnings. So I disabled/suppressed the relevant warnings and CI started passing again.

While I was at it I made some additional changes:

1. Enable some additional ``readability-*`` and ``modernize-*`` warnings (which includes checking for whether functions can be made ``const``!)
2. Get Jenkins to bundle any fixes generated by ``clang-tidy`` into a zip file which can then be downloaded and applied with ``clang-apply-replacements [wherever you've unzipped the file to]``

I was in two minds about whether we want ``clang-tidy`` to pick up on some of these more stylistic-type problems, but as it generally seems to produce decent automated fixes for them, I figure it won't cause too much additional annoyance.


examples/collision_detector/collision_detector.cc
```diff
@@ -42,7 +42,7 @@ class Agent
         }
     }
 
-    virtual void draw(sf::RenderTarget &target, sf::RenderStates states) const override
+    void draw(sf::RenderTarget &target, sf::RenderStates states) const override
```
neworderofjamie, line 45: I know the ``override`` prefixes make this unnecessary but, personally I like having ``virtual`` at the start of virtual functions to make it extra-clear.

include/antworld/render_target.h
```diff
@@ -19,9 +19,9 @@ class RenderTarget
     //------------------------------------------------------------------------
     // Public API
     //------------------------------------------------------------------------
-    void bind();
-    void unbind();
-    void clear();
+    void bind() const;
+    static void unbind();
```
neworderofjamie, line 23: While these can technically be static (because OpenGL is a giant nightmareish, stateful blob), semantically it makes no sense i.e. calling ``RenderTarget::clear`` will clear whatever the active render target it

include/antworld/renderer.h
```diff
@@ -71,8 +71,8 @@ class Renderer
     // Private methods
     //------------------------------------------------------------------------
     void generateCubeFaceLookAtMatrices();
-    void applyFrame(meter_t x, meter_t y, meter_t z,
-                    degree_t yaw, degree_t pitch, degree_t roll);
+    static void applyFrame(meter_t x, meter_t y, meter_t z,
+                           degree_t yaw, degree_t pitch, degree_t roll);
```
neworderofjamie, line 75: Ditto

include/antworld/route_continuous.h
```diff
@@ -80,7 +80,7 @@ class RouteContinuous
     std::tuple<meter_t, meter_t> getDistanceToRoute(const Vector2<meter_t> &position) const;
     Pose2<meter_t, degree_t> getPose(meter_t distance) const;
 
-    void setWaypointFamiliarity(size_t pos, double familiarity);
+    void setWaypointFamiliarity(size_t pos, double familiarity) const;
```
neworderofjamie, line 83: Again, this makes no sense semantically. ``setWaypointFamilarity`` modifies buffers owned by this object via OpenGL C methods which the compiler doesn't understand but should only be done with a non-const reference to the route

neworderofjamie (changes requested):
It's clever but this auto-consting and auto-staticing totally breaks the semantics when we're using C++ to wrap underlying C APIs i.e. all the OpenGL and POSIX stuff so I don't really think it's a useful tool on this codebase as so much of it does that

examples/collision_detector/collision_detector.cc
alexdewar: So the point of this check is to mark functions as ``override`` if they need to be, but it does have the annoying side-effect of also removing the ``virtual`` prefixes when it does so... What do you think? I feel it is a useful thing to check for.

include/antworld/render_target.h
alexdewar: I think I agree with you on this one. We don't really gain anything by doing this (cf. automatically ``const``ifying methods)

Jenkinsfile
```diff
@@ -126,14 +132,26 @@ for(b = 0; b < builderNodes.size(); b++) {
...
 
                     archive uniqueMsg;
 
                     recordIssues enabledForFailure: true, tool: clangTidy(pattern: "**/msg_clang_tidy_" + env.NODE_NAME, id: "clang_tidy_" + env.NODE_NAME), qualityGates: [[threshold: 1, type: 'TOTAL', unstable: true]]
                     if(runClangTidyStatus != 0) {
                         setBuildStatus("Running clang-tidy (" + env.NODE_NAME + ")", "FAILURE")
                     }
+
+                    // If there are auto-generated fixes, archive these in a zip file
+                    if(!isDirEmpty("./clang_tidy_fixes")) {
```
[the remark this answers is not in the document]
alexdewar: I'll give it a go!

examples/collision_detector/collision_detector.cc
neworderofjamie: Marking functions as ``override`` _is_ clearly valuable. I am happy to have my slightly-archaic internal style guide overriden on the virtual issue in exchange 😄

alexdewar: Cool :-)

include/antworld/route_continuous.h
alexdewar: You can explicitly suppress ``clang-tidy`` in these places. How would you feel about a ``BOB_NON_CONST`` macro to tag these kinds of methods with?

src/common/bn055_imu.cc
```diff
@@ -77,14 +77,11 @@ Eigen::Vector3f BN055::getVector(VectorType vectorType)
     // Return vector, scaling correctly based on vector type
     if(vectorType == VectorType::MAGNETOMETER
         || vectorType == VectorType::GYROSCOPE
-        || vectorType == VectorType::EULER)
-    {
-        return Eigen::Vector3f((float)raw[0] / 16.0f, (float)raw[1] / 16.0f, (float)raw[2] / 16.0f);
-    }
-    else {
-        return Eigen::Vector3f((float)raw[0] / 100.0f, (float)raw[1] / 100.0f, (float)raw[2] / 100.0f);
+        || vectorType == VectorType::EULER) {
```
neworderofjamie, line 80: This is horrible formatting - I used an extra line break after a multi-line if for a reason

src/common/serial_interface.cc
```diff
@@ -59,7 +59,7 @@ SerialInterface::setup(const char *path)
 }
 
 void
-SerialInterface::setAttributes(int speed)
+SerialInterface::setAttributes(int speed) const
```
neworderofjamie, line 62: Can we just turn this const thing off entirely? So much of this codebase deals with wrapping C APIs with const-correct C++ wrappers that the number of special cases outweighs the advantages

src/common/bn055_imu.cc
alexdewar: It is indeed :laughing:

src/robots/bebop/video.cc
```diff
@@ -225,7 +225,7 @@ Bebop::VideoStream::decode(const ARCONTROLLER_Frame_t *framePtr)
      * http://developer.parrot.com/blog/2016/ARSDK-3-8-release/
      *
      * */
-    if (m_UpdateCodecParams && m_CodecData.size()) {
+    if (m_UpdateCodecParams && !m_CodecData.empty()) {
```
neworderofjamie, line 228: I like the automatic fixing of this though 😄

src/common/serial_interface.cc
alexdewar: Yeah ok.

src/robots/gazebo/uav.cc
```diff
@@ -37,12 +37,12 @@ void UAV::setPitch(float pitch)
     m_Pitch = pitch;
     sendCommand();
 }
-void UAV::setRoll(float roll)
+void UAV::setRoll(float right)
```
neworderofjamie, line 40: what is this doing!?

alexdewar: That's not an automated fix, though ``clang-tidy`` did flag it up. It has a warning that complains if the var names for the method declaration don't match those in the definition. It does sound a bit pedantic, but it can stop you doing dumb things like:
```c++
void rotate(size_t pixelsToLeft);
// ...
void rotate(size_t pixelsToRight)
{
    // ...
}
```

neworderofjamie: I think that's a reasonable warning but maybe, in this case, either you or it picked the wrong one 😄

alexdewar: Haha... Hmm yeah. I'll sort that

alexdewar: Done! Are we good to merge now?

neworderofjamie: All looks good apart from that one whitespace comment...

neworderofjamie (approved):

alexdewar:
Cheers! I guess you'll probs want to merge ``master`` back into your branch now to get the CI working again...

PR #246, opened 2021-07-01
Allow running Gazebo examples without joystick

alexdewar:
We could also just use a virtual joystick as a fallback, but in the meantime let's at least get the examples to run without a joystick to make life easier for @nitik1998.

### DIFF
--- a/examples/gazebo_tank/gazebo_tank.cc
+++ b/examples/gazebo_tank/gazebo_tank.cc
@@ -1,10 +1,10 @@
 // BoB robotics includes
-#include "plog/Log.h"
 #include "hid/joystick.h"
-#include "robots/simulated_tank.h"
+#include "plog/Log.h"
 #include "robots/gazebo/camera.h"
 #include "robots/gazebo/node.h"
 #include "robots/gazebo/tank.h"
+#include "robots/simulated_tank.h"
 #include "video/display.h"
 
 // Third-party includes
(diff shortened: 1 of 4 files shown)


examples/gazebo_tank/run.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-trap 'kill %1' SIGINT
+trap 'kill %1' INT
```
neworderofjamie, line 2: What's the difference here? I don't think I've ever actually used ``trap`` in bash but googled examples seem to use ``SIGINT``

examples/gazebo_tank/gazebo_tank.cc
```diff
@@ -33,38 +34,48 @@ int bobMain(int argc, char **argv)
...
+    }
 
-    do {
-        // Check for joystick events
-        if (!joystick.update() && (!display || !display->update())) {
-            // A small delay so we don't hog CPU
-            std::this_thread::sleep_for(5ms);
-        }
-    } while (!joystick.isPressed(HID::JButton::B));
+    if (!display && !joystick) {
```
neworderofjamie, line 60: What happens in either of these cases when you have a display but no joystick?

examples/gazebo_tank/run.sh
alexdewar: Apparently they're equivalent, but the standard POSIX way of doing it is to use ``INT`` instead. Nitik is using zsh which doesn't like ``SIGINT``

examples/gazebo_tank/gazebo_tank.cc
alexdewar: If there's a display but no joystick, then you'll have an ``OpenCV`` window showing the current view and you can terminate the program by closing it.

neworderofjamie (approved):

Issue #241, opened 2021-06-16
RC car control is laggy when both remote control + software control used


alexdewar:

alexdewar:
Fixed in #248

PR #248, opened 2021-07-06
Rework RC car control code

alexdewar:
Pertains to issue #241 (though I think the *real* reason we were seeing the lagginess was because input from the RC controller was essentially overriding the values read over I2C).

This is very much a work in progress. I'm still getting lagginess/jerkiness when using RC control + I2C control together, *however* they can at least be properly used together now.

I've allowed for setting the Arduino into one of two states, depending on whether you want it to be controlled by the RC controller directly or by values sent over I2C. The state can be changed with an I2C command. Values from the RC controller can also be read over I2C.

Are you around this week to do some hacking together on this? I feel like your expertise would be handy here!


neworderofjamie:
Definitely be in the office tomorrow and Friday - would be more than happy to help hack at this

alexdewar:
It works :partying_face: !

It's now happily driving around the office. @tenxlenx I also added a separate class (``PassiveRCCarBot``) for driving the robot exclusively with the remote control.

include/robots/rc_car_common.h
```diff
@@ -0,0 +1,38 @@
...
+};
+
+struct __attribute__((packed)) Message
+{
+    Command command;
+    union
+    {
+        State state;
+        Movement move;
+    };
```
neworderofjamie, line 34: This packing seems a little sub-optimal as sizeof(State) == 1 but sizeof(Movement) == 2. Why not have:
```c++
enum class Command : uint8_t
{
    SetStateRemoteControl,
    SetStateI2CControl,
    ReadRemoteControl,
    Drive
};

struct __attribute__((packed)) Message
{
    Command command;
    Movement move;
}
```
save a byte **and** get rid of a union - what could be better :smile:

include/common/i2c_interface.h
```diff
@@ -40,25 +40,23 @@ class I2CInterface
...
     void write(const T (&data)[N])
     {
-        const size_t size = sizeof(T) * N;
-        if (::write(m_I2C, &data[0], size) != size) {
-            throw std::runtime_error("Failed to write to i2c bus");
-        }
+        write(&data[0], N * sizeof(T));
     }
 
+    void write(const void *data, size_t size);
```
neworderofjamie, line 58: This is a bit of a nasty Cish API. If you wanted to be (unnecessarily?) fancy, you could do:
```c++
template<typename T, 
         typename std::enable_if<std::is_pod<T>::value, T>::type* = nullptr>
void write(const T &data)
{
    if (::write(m_I2C, &data, sizeof(T)) != sizeof(T)) {
        throw std::runtime_error("Failed to write to i2c bus");
    }
}
```

include/robots/rc_car_bot.h
```diff
@@ -4,30 +4,58 @@
...
+    void writeMessage(const RCCar::Message &msg);
+}; // PassiveRCCarBot
+
 //----------------------------------------------------------------------------
 // BoBRobotics::Robots::RCCarBot
 //----------------------------------------------------------------------------
 //! An interface for 4 wheeled, Arduino-based robots developed at the University of Sussex
 class RCCarBot final
-  : public Ackermann
+  : public Ackermann, public PassiveRCCarBot
```
neworderofjamie, line 53: Why not solve this multiple inheritance and the forementioned slicing issue by having ``RCCarBot`` own a ``PassiveRCCarBot`` and prodding it rather than inheriting from it (you'll also need to a little befriending)

alexdewar:
Fixes #241. I just tested :-)

include/robots/rc_car_common.h
alexdewar: Good idea!

alexdewar:
Right... I think I've addressed everything now. Let me know if you're happy so we can merge this sucker

include/common/i2c_interface.h
```diff
@@ -43,7 +43,12 @@ class I2CInterface
         read(&data[0], N * sizeof(T));
     }
 
-    void read(void *data, size_t size);
+    template<typename T,
+             typename std::enable_if<std::is_pod<T>::value, T>::type* = nullptr>
+    void read(T &data)
+    {
+        read(&data, sizeof(T));
+    }
```
neworderofjamie, line 51: now that's some elegant C++11 - bring on C++20 and concepts 😟

src/robots/rc_car_bot.cc
```diff
@@ -33,54 +33,50 @@ void
...
+    m_Bot.setState(RCCar::State::RemoteControl);
 }
 
 //! Move the car with Speed: [-1,1], TurningAngle: [-35,35]
 void
 RCCarBot::move(float speed, degree_t left)
 {
     BOB_ASSERT(speed >= -1.f && speed <= 1.f);
-    BOB_ASSERT(left >= -TurnMax && left <= TurnMax);
+    BOB_ASSERT(units::math::abs(left) <= PassiveRCCarBot::TurnMax);
```
neworderofjamie, line 73: presumably ``units::math::abs`` is ``fabs`` in standard maths library speak?

neworderofjamie (approved):

alexdewar:
I've merged this but GitHub doesn't seem to have twigged...

PR #249, opened 2021-07-07
GCC 5.4.0 fixes

neworderofjamie:
First issue was just that the compiler was too dumb to figure out where ``toIndex`` was. Second one was much gnarlier and probably a compiler bug. It **looked** like the code to initialize thread local variables wasn't being linked correctly and I suspect that this was because the compiler was failing to figure out that the function static thread local variables were actually being used. I made them class static instead and it seemed appeased. Thread local variables still hurt my head a little but I think this is equivalent.

### DIFF
--- a/include/navigation/insilico_rotater.h
+++ b/include/navigation/insilico_rotater.h
@@ -55,17 +55,14 @@ struct InSilicoRotater
         template<class Func>
         void rotate(Func func) const
         {
-            static thread_local cv::Mat scratchImage;
-            static thread_local ImgProc::Mask scratchMask;
-
             tbb::parallel_for(tbb::blocked_range<size_t>(0, numRotations()),
-                [&](const auto &r) {
+                [this, func](const auto &r) {
                     for (size_t i = r.begin(); i != r.end(); ++i) {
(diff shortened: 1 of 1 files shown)


alexdewar:
I think it's equivalent too. As long as it has the ``thread_local`` attribute I think we're good!

alexdewar (approved):
The fix actually wasn't as gnarly as I expected! At least compared to ugliness we had to do for the sake of OpenMP.

I suspect there are a few other places with a similar pattern though. I guess going forward we might want to stick to Ubuntu 18.04 or newer. Gcc 5 in particular has long been a source of pain for us.

neworderofjamie:
Yeah, Ubuntu 18 would be nice but the new jetson installer is super shit - I couldn't figure out a way to not get it to mess with my laptop's (delicately configured cos it's got dual graphics) cuda install so I stuck with the last jetpack version to use the old installer

alexdewar:
Oh gross. It messes with the Cuda install on your laptop?!

neworderofjamie:
Yeah, it wants to install a whole load of crap

alexdewar:
Yikes. Nvidia sure do like making things painful sometimes...

PR #254, opened 2021-07-28
Don't use GAZEBO_CXX_FLAGS

alexdewar:
It seems like the only thing it does is force the C++ standard to C++11,
which we don't want.


neworderofjamie (commented):
My only concern would be that it may start doing something else in the future. You _could_ use ``string(REGEX_REPLACE...) `` to  remove any - -std= flags

alexdewar:
Good point. I've done it as a bit of a lazy hack here

neworderofjamie:
Totally up to you whether you care about the hackiness or not :smile:

alexdewar:
After careful consideration I have decided I don't care about the hackiness :wink:

PR #255, opened 2021-07-28
Support for robots using PCA 9685 motor controller

neworderofjamie:
This is basically a chip which sits on the I2C bus and controls 16 Pulse Width Modulation channels which can be used be used to drive motors/flashing LEDS/whatever. 

You can get boards containing the chip by itself (https://thepihut.com/products/adafruit-16-channel-12-bit-pwm-servo-driver-i2c-interface-pca9685) or with motor drivers attached (https://thepihut.com/products/adafruit-dc-motor-stepper-featherwing-add-on-for-all-feather-boards). In this PR we:
* Define a simple interface to the PCA9685 in common 
* Define a wrapper around this to control the featherwing 
* However, turns out, the featherwing motor drivers are particularly weedy and the mecanum motors burns them out. On the new mecanum we've instead hooked up 4 motor driver boards to the first 8 channels of a PCA9685 and this configuration is supported by the ``MecanumPCA9685`` robot type.


include/common/pca_9685.h
```diff
@@ -0,0 +1,104 @@
...
+        CHANNEL_7 = 7,
+        CHANNEL_8 = 8,
+        CHANNEL_9 = 9,
+        CHANNEL_10 = 10,
+        CHANNEL_11 = 11,
+        CHANNEL_12 = 12,
+        CHANNEL_13 = 13,
+        CHANNEL_14 = 14,
+        CHANNEL_15 = 15,
+    };
```
alexdewar, line 46: I feel like having this as an ``enum`` is a bit unnecessary, but it's harmless I guess

src/common/pca_9685.cc
```diff
@@ -0,0 +1,138 @@
...
+    // Setup I2C device
+    m_Device.setup(path, slaveAddress);
+    
+    reset();
+}
+//----------------------------------------------------------------------------
+void PCA9685::setDutyCycle(Channel channel, float dutyCycle)
+{
+    // Clamp duty cycle
+    dutyCycle = std::min(1.0f, std::max(0.0f, dutyCycle));
```
alexdewar, line 32: Elsewhere in BoB robotics I've generally used error checks rather than silently capping values. The danger is that your code ends up passing some nonsense value in and you never notice.

How about:
```c++
BOB_ASSERT(dutyCycle >= 0.f && dutyCycle <= 1.f);
```

src/robots/mecanum_pca_9685.cc
```diff
@@ -0,0 +1,85 @@
...
+}
+//----------------------------------------------------------------------------
+void MecanumPCA9685::driveMotors(float m1, float m2, float m3, float m4)
+{
+    // clamp values to be between -1 and 1 after resolving
+    const auto cap = [](float &val) { val = std::min(1.f, std::max(val, -1.f)); };
+    cap(m1);
+    cap(m2);
+    cap(m3);
+    cap(m4);
```
alexdewar, line 57: See comment above re capping (although I have a feeling we may actually do this in the old mecanum code)

src/robots/mecanum_pca_9685.cc
```diff
@@ -0,0 +1,85 @@
...
+
+    setMotorThrottle(Motor::MOTOR_1, m1);
+    setMotorThrottle(Motor::MOTOR_2, m2);
+    setMotorThrottle(Motor::MOTOR_3, m3);
+    setMotorThrottle(Motor::MOTOR_4, m4);
+}
+//----------------------------------------------------------------------------
+void MecanumPCA9685::setMotorThrottle(Motor motor, float throttle)
```
alexdewar, line 65: The problem with ``Motor`` being an ``enum`` rather than an ``enum class`` is that a user could accidentally pass e.g. a value of ``5`` for ``motor`` and the compiler would happily accept it. (You could also just check for this case with ``BOB_ASSERT()``).

alexdewar (commented):
LGTM aside from a couple of ``enum``-related niggles!

include/common/pca_9685.h
neworderofjamie: It just means you don't need to assert somewhere that channel >= 1 and channel <= 15 :man_shrugging:

include/common/dc_motor_featherwing.h
```diff
@@ -0,0 +1,66 @@
...
+    //------------------------------------------------------------------------
+    // Enumerations
+    //------------------------------------------------------------------------
+    enum MotorChannel
+    {
+        MOTOR_CHANNEL_ENABLE,
+        MOTOR_CHANNEL_POSITIVE,
+        MOTOR_CHANNEL_NEGATIVE,
+        MOTOR_CHANNEL_MAX
+    };
```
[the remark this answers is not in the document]
neworderofjamie: Because I'm using them (in a somewhat old school way) to index into ``motorChannels`` which enum classes make super irritating (``static_cast`` EVERYWHERE)

src/robots/mecanum_pca_9685.cc
neworderofjamie: I wonder here if there should be an assert here and capping in the omni2D instead as I fear the derived m1, m2, m3 and m4 values may well end up not being between -1 and 1

src/robots/mecanum_pca_9685.cc
neworderofjamie: True - as it's an old-style enum I should check - bleurgh

neworderofjamie:
I've added asserts to cover the various enum and floating point input limits and moved the capping in the mecanum. Assuming I haven't broken it on the robot, are you ok to merge?

alexdewar:
Yep!

I'd also like to get #257 merged as I'm using it as a base for something else I'm doing. Are you ok to merge that once I've devirtualised `MecanumPCA9685` in my branch?

neworderofjamie:
So you've already merged this into #257?

neworderofjamie:
Ok, joystick_robot doesn't hit any asserts so I'm gonna merge

PR #257, opened 2021-08-10
Devirtualise robot classes

alexdewar:
So I think this is ready for review now...

I've refactored the robot classes to try to solve a few issues. In general the base-type classes were getting rather bloated (e.g. the ``Tank`` class incorporated methods for control with a joystick and over the network). This also meant that e.g. the ``robots`` module had to depend on ``hid`` and ``net``, even where this functionality wasn't being used. Another problem is that there were various virtual methods for robot classes that were basically optional for derived classes and if they weren't implemented you would get an error at runtime, when really it would be better to have one at compile time. A lot of the code we have would work fine for any robot class that implements a ``tank()`` method and nothing else, so it seems sensible to just use templates (which also means that we don't have to build every possible type of tank robot along with whatever project is using it).

I opted to leave the base classes there for tank robots, Ackermann robots etc., though using CRTP rather than as virtual classes (for a few simple methods like ``stopMoving()``) but they could also probably be removed in some cases (e.g. the ``AckermannBase`` class is currently just a placeholder with no methods).

I've also made things more modular. Notably, the various robot types now live in their own submodules (e.g. ``robots/tank``, ``robots/ackermann`` etc.) which means that there's less unnecessary compiling. I've also put all the classes into namespaces of the same names (i.e. tank robots are in the ``Robots::Tank`` namespace) which means that CMake can easily work out which robot module is needed if you want to specify a ``ROBOT_TYPE`` at compile time. For example, to use a Mecanum robot you'd build your project with:
```sh
cmake -DROBOT_TYPE=Omni2D::Mecanum ..
# (etc.)
```
Then the ``robots/omni2d`` module would also be built.

To make things more modular, I've moved all the code for controlling robots with joysticks to ``include/hid/robot_control.h``, using type traits to figure out how to control the robot (e.g. for a class with a ``tank()`` method, use the left stick etc.).

Also, we aren't using surveyor robots anymore so I dropped support for them.

Let me know what you think!


cmake/bob_robotics.cmake
```diff
@@ -660,6 +631,43 @@ endfunction()
 
 function(BoB_modules)
     foreach(module IN LISTS ARGV)
+        if(module STREQUAL robots)
```
neworderofjamie, line 634: So, just to clarify, this logic only happens if you just specify the ``robots`` module? Otherwise it's up to you to manually include your robot header etc

cmake/bob_robotics.cmake
```diff
@@ -660,6 +631,43 @@ endfunction()
...
+            # are used, with :: replaced with _, e.g.: Namespace::RobotClass becomes NAMESPACE_ROBOTCLASS
+            string(TOUPPER ${ROBOT_TYPE} ROBOT_TYPE_UPPER)
+            string(REGEX REPLACE :: _ ROBOT_TYPE_UPPER ${ROBOT_TYPE_UPPER})
+            add_definitions(-DROBOT_TYPE_${ROBOT_TYPE_UPPER})
+
+            # Extra modules needed for some robot types. The namespace name tells us
+            # which module to add (e.g. all Ackermann robots live in robots/ackermann).
+            string(TOLOWER ${ROBOT_TYPE} ROBOT_TYPE_LOWER)
+            string(REGEX REPLACE :: / ROBOT_TYPE_LOWER ${ROBOT_TYPE_LOWER})
+            get_filename_component(ROBOT_TYPE_MODULE ${ROBOT_TYPE_LOWER} DIRECTORY)
```
neworderofjamie, line 665: eugh cmake functions are so inconsistent - the output argument being the first one totally confused my limited CMake parsing ability here

include/hid/robot_control.h
```diff
@@ -0,0 +1,222 @@
...
+namespace BoBRobotics {
+namespace HID {
+template<class RobotType>
+void
+drive(RobotType &robot, const JoystickBase<JAxis, JButton> &joystick)
+{
+    drive(robot, joystick, DEFAULT_DEAD_ZONE);
+}
+
+template<class AckermannType,
```
neworderofjamie, line 27: I wonder if this could be restructured a little to reduce the duplication of ``enable_if`` horror:
```c++
template <typename AckermannType,
    std::enable_if_t<Robots::isAckermann<AckermannType>::value, int> = 0>
struct RobotControl
{
     static constexpr auto useAxes();

     static void addJoystick(...);

     static void drive(...);
};
```

include/robots/ackermann/ackermann_base.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+namespace BoBRobotics {
+namespace Robots {
+namespace Ackermann {
+//----------------------------------------------------------------------------
+// BoBRobotics::Robots::AckermannBase
+//----------------------------------------------------------------------------
+//! A base class for robots with Ackermann-type steering
+template<class Derived>
```
neworderofjamie, line 10: What's the purpose of this base class? Couldn't you remove it altogether? At first I thought it might be for ``std::enable_if``` but it doesn't seem to be

include/robots/type_traits.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#define BOB_DEFINE_HAS_METHOD(traitName, methodName)      \
```
neworderofjamie, line 3: I think this is a bit too clever. There aren't THAT many types I would just manually define the type traits for each robot type (this could be done with some macro boilerplate)

cmake/bob_robotics.cmake
alexdewar: Yes. This is only for the case when you're using ``ROBOT_TYPE``. I should add a comment explaining this...

cmake/bob_robotics.cmake
alexdewar: I know :cry:

include/hid/robot_control.h
alexdewar: Good idea!

include/robots/ackermann/ackermann_base.h
alexdewar: Nothing, really. I left it as a placeholder but we could remove it entirely... unless we use it for type traits (see below)

include/robots/type_traits.h
alexdewar: Hmm I'm not sure exactly what you're suggesting. Do you mean that robots classes would each have e.g. a static ``IsTank`` member if they're of a tank type?

alexdewar: I suppose we could also probably just check whether a class inherits from ``TankBase``

neworderofjamie: I mean, as each of your robot types *has* a corresponding base class, that would definitely be one approach. What I was thinking of was:
```c++
template<typename T>
struct IsAckermann
{
    static constexpr bool value = false;
}

template<>
struct IsAckermann<RCCar>
{
    static constexpr bool value = true;
}
```
Where the specialisation could easily be added via a macro if you wanted

alexdewar: Ahh I see.

I just tried an alternative approach to see if it works and it does! This code just checks whether a given class inherits from e.g. ``TankBase`` to determine whether it's a tank or not:
```c++
#pragma once

#define BOB_DEFINE_TRAIT(TYPE)                                                        \
    namespace TYPE {                                                                  \
    template<class Derived>                                                           \
    class TYPE##Base;                                                                 \
    }                                                                                 \
    template<class T>                                                                 \
    struct Is##TYPE                                                                   \
    {                                                                                 \
        static constexpr bool value = std::is_base_of<TYPE::TYPE##Base<T>, T>::value; \
    };

namespace BoBRobotics {
namespace Robots {

BOB_DEFINE_TRAIT(Ackermann)
BOB_DEFINE_TRAIT(Tank)
BOB_DEFINE_TRAIT(Omni2D)
BOB_DEFINE_TRAIT(UAV)

} // Robots
} // BoBRobotics
```

It does mean that we'd have to keep the ``AckermannBase`` class though.

neworderofjamie: Awesome - as you have the base classes, this is probably the way to go! In this case, I'd probably just stick the ``std::is_base_of`` straight into the ``std::enable_if`` clauses where you use them and delete this header though.

alexdewar: I feel the header could be generically useful though. There is a bit of boilerplate you have to do (i.e. forward-declaring your base classes or including the relevant headers) so it would be a faff if we wanted to use it elsewhere.

Atm we do have places where we're doing things like ``#ifdef ROBOT_TYPE_MECANUM`` where we could express it more elegantly with a ``Robots::IsOmni2D<RobotType>``. Would be handy for new code.

neworderofjamie: In order to do anything with a robot object you'll need to have it's base class #included one way or other and the problem with replacing #ifdefs is that you often use them to guard code that won't compile e.g. linux-only robot types but....nonetheless this is probably generally useful 😄

alexdewar: That's true, but if you're only using tank-type robots you wouldn't want to have to include headers for Ackermann robots just for the sake of making the ``std::enable_if``s work. We def should be able to at least reduce *some* ifdeffery this way...

include/hid/robot_control.h
alexdewar: Although I've just realised that the compiler won't be able to infer the type from the arguments passed to these member functions, so you will end up with horrors like this:
```c++
        HID::RobotControl<ROBOT_TYPE>::drive(robot, joystick, 0.25f);
```

You *could* get round this somewhat by having a ``using RobotControl = HID::RobotControl<ROBOT_TYPE>;`` at the start of your method but it still seems a bit gross.

neworderofjamie: Oh yes - shit! That is always the problem with templated classes/structs - adding wrappers would undo any of the tidiness improvement many times over so probably best to leave as is

alexdewar: Cool. So are we good to merge then? I think I've addressed everything...

neworderofjamie: Totally fine to merge!

alexdewar:
Thanks! Sorry to bother you on your hols!

PR #258, opened 2021-08-12
Add std::optional header-only library

alexdewar:
std::optional is a rather handy interface that is sadly only available from C++17 onwards.

I'm working on some code where it would make sense to use this kind of interface and I feel like it could be useful elsewhere.

### DIFF
--- a/third_party/optional.hpp
+++ b/third_party/optional.hpp
@@ -1,0 +1,1066 @@
+// Copyright (C) 2011 - 2012 Andrzej Krzemienski.
+//
+// Use, modification, and distribution is subject to the Boost Software
+// License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+//
+// The idea and interface is based on Boost.Optional library
+// authored by Fernando Luis Cacciola Carballal
+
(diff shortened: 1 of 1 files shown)


third_party/optional.hpp
```diff
@@ -0,0 +1,1076 @@
...
+# undef TR2_OPTIONAL_REQUIRES
+# undef TR2_OPTIONAL_ASSERTED_EXPRESSION
+
+/*
+ * Import everything into BoBRobotics namespace.
+ *
+ * This is a bit of a hack, but prefixing everything with std::experimental is
+ * annoyingly verbose.
+ */
+namespace BoBRobotics {
```
neworderofjamie, line 1072: std::optional is definitely one of the most obviously useful bits of C++17! I kinda thing it should be up to the code that uses this to tersify ``std::experimental`` though

alexdewar: Hmm... You're probably right. I'll leave out the namespace malarkey then

alexdewar: Are you happy to merge otherwise?

neworderofjamie: yep! the nuance of a third-party backport of ``std::optional`` is beyond me 😄

alexdewar: Haha - me too!

Issue #261, opened 2021-08-14
``GPSReader`` doesn't return current date


alexdewar:
GPS devices provide the current date as well as the current time, though we are not presently parsing this information. We should implement this (e.g. in ``dataset_recorder``) as the clock on Jetsons does not tick between boots, so the system clock's date can be wrong.

PR #262, opened 2021-08-16
GPS updates and fixes

### LINKED ISSUE #260: dataset_recorder uses UTC cf. localtime for database timestamps
This isn't the biggest issue given our longitude, but it does mean that the timestamps will be one hour out for half the year when the UK is on BST.

Could be easily remedied by converting to localtime.

alexdewar:
The main point of this PR is to add support for getting date information from the GPS device, but I've also done some refactoring and bugfixing along the way. A related (minor) issue is that the GPS timestamps used by ``dataset_recorder`` are in UTC rather than localtime so they are currently one hour out due to DST.

Commit summary:
- Put GPS code into own module so it's only compiled when needed
- Read date and timezone info from GPS and use this info in ``dataset_recorder`` (fixes #260 and #261)
- Actually check the checksums for the NMEA messages (I suspect they're unlikely to be corrupted in transmission from the GPS device to the host, but it's a good sanity check for parsing and serial device bugs)
- Record horizontal dilution (i.e. error estimate in metres) in ``dataset_recorder``
- Fix idiotic bug I introduced that meant ``dataset_recorder`` wasn't actually recording GPS readings


src/gps/nmea_parser.cc
```diff
@@ -0,0 +1,137 @@
...
+
+    try {
+        parseTimeField(m_Fields[0], time);
+        time.tm_mday = stoi(m_Fields[1]);
+        time.tm_mon = stoi(m_Fields[2]) - 1;
+        time.tm_year = stoi(m_Fields[3]) - 1900;
+
+        // This field is available in GNU/BSD but is not strict ISO C
+        int hoursOff = m_Fields[4].empty() ? 0 : stoi(m_Fields[4]);
+        int minutesOff = m_Fields[5].size() <= 3 ? 0 : stoi(m_Fields[5].substr(0, 2));
```
neworderofjamie, line 34: what's going on with ``m_Fields[5]``? Also, what's GNU/BSD got to do with this? Aren't the fields coming straight from the GPS device? 😨

src/gps/nmea_parser.cc
```diff
@@ -0,0 +1,137 @@
...
+        time.tm_gmtoff = 3600 * hoursOff + 60 * minutesOff;
+    } catch (std::invalid_argument &e) {
```
neworderofjamie, line 36: I've never actually done it before but I _think_ the correct thing to do here would be to 
```c++
catch (const std::invalid_argument &e) {
    std::throw_with_nested(NMEAError());
}
catch (const std::out_of_range &e) {
    std::throw_with_nested(NMEAError());
}
```
Then wherever you catch it:
```c++
catch(const NMEAError &e) {
   LOGW << "NMEA parsing error: " << m_Line;
    try {
        std::rethrow_if_nested(e);
    } catch(const std::exception& e) {
        LOGW << "NMEA parsing error internal:" <<  e.what();
    }
}
```
That way you get to keep the single exception type to throw but you aren't throwing away the internals which might be helpful for your debugger.

src/gps/nmea_parser.cc
```diff
@@ -0,0 +1,137 @@
...
+        return false;
+    }
+
+    if (line.size() < 3 || line[line.size() - 3] != '*') {
+        throw NMEAError{ "Missing checksum" };
+    }
+
+    // Check that the checksum matches
+    char *p;
+    long reportedChecksum = std::strtoul(&line[line.size() - 2], &p, 16);
```
neworderofjamie, line 109: This is a pretty hardcore ``std::strtoul`` invocation (I definitely had to look at the manual) - maybe add a comment to explain that you're parsing in hex and that you use the p to detect failure (I figure you're doing it because 0 could also be a valid checksum but from the documentation this still seems suspect - I see no reference to this being a guaranteed way to signal an error). In short, I *think* you should use ``std::stoul`` instead (https://en.cppreference.com/w/cpp/string/basic_string/stoul) and just check for ``std::invalid_argument`` and ``std::out_of_range``

src/gps/nmea_parser.cc
```diff
@@ -0,0 +1,137 @@
...
+        std::stringstream ss;
+        ss << "Bad checksum (expected: " << std::hex << reportedChecksum
+           << ", got: " << std::hex << actualChecksum << ")";
+        throw NMEAError{ ss.str() };
+    }
+
+    // Extract the remaining fields
+    size_t count = 0;
+    for (; getline(lineStream, field, ',') && count < numFields; count++) {
+        m_Fields[count] = std::move(field);
```
neworderofjamie, line 128: I don't think this is a valid ``std::move``. ``field`` is being re-used so you can't/shouldn't steal it's heap-allocated memory!

tests/nmea_parser.cc
```diff
@@ -0,0 +1,65 @@
...
+    BOB_EXPECT_UNIT_T_EQ(data.coordinate.lon, degree_t{ 121 } + arcminute_t{ 18.85961 });
+    BOB_EXPECT_UNIT_T_EQ(data.coordinate.height, meter_t{ 1113 });
+
+    // Check self-reported accuracy
+    BOB_EXPECT_UNIT_T_EQ(data.horizontalDilution, meter_t{ 0.98 });
+
+    // Misc
+    EXPECT_EQ(data.gpsQuality, GPSQuality::GPSFIX);
+    EXPECT_EQ(data.numberOfSatellites, 12);
+}
```
neworderofjamie, line 40: some 'death' tests of the form:
```c++
 try {
        // **TODO** something
        FAIL();
    }
    catch(const NMEAError &) {
    }
```
would be nice to check the error handling works properly

include/gps/gps_reader.h
```diff
@@ -16,6 +19,9 @@ class GPSReader
     static constexpr const char *DefaultLinuxDevicePath = "/dev/serial/by-id/usb-u-blox_AG_-_www.u-blox.com_u-blox_GNSS_receiver-if00";
     GPSReader(const char *devicePath = DefaultLinuxDevicePath);
 
+    //! Note: This will probably be UTC
+    const std::tm &getCurrentDateTime() const;
```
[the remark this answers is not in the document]
alexdewar: In theory it can be a local time (but I've no idea why anyone would actually *want* their GPS device to do this). You get timezone info in the ``$GNZDA`` messages that we're now parsing, so the code should handle this properly.

It doesn't really matter because we'd likely always want to convert the returned time to the user's localtime anyway (to avoid bugs like #260).

src/gps/gps_reader.cc
```diff
@@ -0,0 +1,155 @@
...
+    /*
+     * B9600   : set baudrate to 9600
+     * CS8     : 8n1 (8bit,no parity,1 stopbit)
+     * CLOCAL  : local connection, no modem contol
+     * CREAD   : enable receiving characters
+     */
+    tty.c_cflag = B9600 | CRTSCTS | CS8 | CLOCAL | CREAD;
+
+    // ICANON  : enable canonical input
+    tty.c_lflag = ICANON;
```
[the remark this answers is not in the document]
alexdewar: Yup! It turns out this is much nicer than actually trying to figure out where lines begin and end manually.

src/gps/gps_reader.cc
```diff
@@ -0,0 +1,155 @@
...
+    // if GPS location is invalid, keep trying to get a valid one
+    // if failed x times we exit
+    int numTrials = 3;
+    GPSData data{};
+    while (numTrials--) {
+        read(data);
+        if (data.gpsQuality != GPSQuality::INVALID) {
+            LOGI << "Valid GPS fix found (" << data.coordinate.lat.value() << "°, "
+                 << data.coordinate.lon.value() << "°)";
+            break;
```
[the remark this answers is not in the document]
alexdewar: Good point!

include/gps/gps_reader.h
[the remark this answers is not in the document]
neworderofjamie: ahh gotcha

examples/gps/gps.cc
```diff
@@ -35,54 +36,27 @@ bobMain(int argc, char **argv)
...
-        return EXIT_FAILURE;
-    }
-
     // Let the user set to nonblocking mode for testing purposes
     if (argc > 1 && strcmp(argv[1], "nonblock") == 0) {
         gps.setBlocking(false);
     }
 
     // print for 100 timestep
+    GPSData data;
```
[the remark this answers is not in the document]
neworderofjamie: using ``std::optional`` would definitely be the more modern solution but there's plenty of other places in the code we returning using out var references so totally up to you

src/gps/nmea_parser.cc
```diff
@@ -0,0 +1,137 @@
...
+namespace GPS {
+
+void
+parseTimeField(const std::string &field, std::tm &time)
+{
+    time.tm_hour = stoi(field.substr(0, 2));
+    time.tm_min = stoi(field.substr(2, 2));
+    time.tm_sec = stoi(field.substr(4, 2));
+
+    // Indicates that information about DST is not available
```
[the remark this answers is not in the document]
alexdewar: Nope. It should be set to 0 if it isn't DST and 1 if it is. -1 just means we don't know (which is true because there's no way to signal this with NMEA messages).

I think in our case -1 will be functionally equivalent to 0, but the semantics are slightly different.

src/gps/nmea_parser.cc
```diff
@@ -0,0 +1,137 @@
...
+    time.tm_isdst = -1;
+}
+
+bool
+NMEAParser::parseDateTime(const std::string &line, std::tm &time)
+{
+    if (!parse(line, "$GNZDA", 6)) {
```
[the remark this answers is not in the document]
alexdewar: :+1:

src/gps/nmea_parser.cc
alexdewar: Apparently the ``std::move`` leaves it in a "valid, but undefined state": https://stackoverflow.com/questions/9168823/reusing-a-moved-container/9168917#9168917

So it's ok to reassign to it (as this operation does not assume the ``std::string`` is in a particular state).

tests/nmea_parser.cc
alexdewar: Good idea

src/gps/nmea_parser.cc
alexdewar: Ok the commenting is a bit fucked here. It's actually the ``tm_gmtoff`` field of the ``std::time`` which is a GNU/BSD thing, not the fields of the NMEA sentence!

We only use the first two chars of ``m_Fields[5]`` for ``minutesOff`` because the last three chars are used for the checksum (e.g. ``*A0``)

src/gps/nmea_parser.cc
alexdewar: Oh cool! I didn't know this was a thing

src/gps/nmea_parser.cc
neworderofjamie: but, semantically, moving makes no sense so why do it!?

src/gps/nmea_parser.cc
neworderofjamie: hah me neither - C++ exception-handling seems an infinitely deep mire of complexity

src/gps/nmea_parser.cc
alexdewar: I think doing it this way *might* save us a copy (because the allocated memory from ``field`` will always have been transferred when the loop ends). But there won't be much in it performance-wise :laughing:

tests/nmea_parser.cc
```diff
@@ -63,3 +66,46 @@ TEST(NMEAParser, SampleGNZDASentence)
     EXPECT_EQ(time.tm_gmtoff, 0);
     EXPECT_EQ(time.tm_isdst, -1);
 }
+
+/*
+ * Check that a broken NMEA sentence triggers an NMEAError. Note that we append
+ * the correct checksum so that the parsing doesn't fail on those grounds.
+ */
+#define TEST_BROKEN_NMEA(name, line)                                                       \
```
neworderofjamie, line 74: All the test magic also works inside functions so there's no need to use macros. Also, this functionality is built into gtest. You can do something like this:
```c++
class BrokenNMEATest : public ::testing::TestWithParam<std::string>
{
};

TEST_P(BrokenNMEATest, BrokenNMEA)
{
    NMEAParser nmea;
    GPSData data{};

    std::stringstream ss;
    const auto checksum = NMEAParser::computeChecksum(GetParam(), GetParam().length()); 
    ss << line << "*"
        << std::uppercase << std::setfill('0') << std::setw(2) << std::hex << checksum;

    EXPECT_THROW(nmea.parseCoordinates(ss.str(), data), NMEAError); 
}

INSTANTIATE_TEST_CASE_P(NMEAParser,
                        BrokenNMEA,
                        ::testing::Values(),);
```
Documented [here](https://github.com/google/googletest/blob/master/docs/advanced.md#value-parameterized-tests)

alexdewar: That is cool, but here I do want to have different names for the different tests, so I know which of them is failing.

neworderofjamie: Oh yeah - duh - I didn't grasp that aspect:
```c++
TEST(NMEAParser, TooFewFields){ nmeaDeathTest("$GNGGA,001043.00,4404.14036,N,12118.85961,W,1,12"); }
```
would probably still be nicer than a macro though

alexdewar: True. I've done that now

examples/gps/gps.cc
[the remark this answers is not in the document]
alexdewar: I decided to do this in the end btw. It's debatable as to whether the syntax *looks* nicer, but it does mean that you can do things like this:
```c++
        if (const auto optData = gps.read()) {
            const auto &gpsData = optData.value();
            // ...
```

And then we know ``gpsData`` always has valid data in the scope it exists in :smile:

neworderofjamie (approved):

src/common/map_coordinate.cc
```diff
@@ -0,0 +1,12 @@
+// BoB robotics includes
+#include "common/map_coordinate.h"
+
+namespace BoBRobotics {
+namespace MapCoordinate {
+const Ellipsoid WGS84::ellipsoid{units::length::meter_t(6378137), units::length::meter_t(6356752.314245)};
```
neworderofjamie, line 6: Is this finally the end of ``NO_HEADER_DEFINITIONS``? :partying_face: I feel this instance was definitely my fault...

alexdewar: Not quite:
```
alex @ ~/code/bob_robotics
 └─ gps_updates $ git grep -n NO_HEADER_DEFINITIONS -- :^docs
cmake/bob_robotics.cmake:268:    add_definitions(-DNO_HEADER_DEFINITIONS)
projects/snapshot_bot/image_input.cc:1:#define NO_HEADER_DEFINITIONS
projects/snapshot_bot/memory.cc:1:#define NO_HEADER_DEFINITIONS
```
:wink:

PR #263, opened 2021-08-20
Automatic data collection for RC car + visualisation updates

alexdewar:
This is based on #262, so we should merge that PR before this one.

Logically this should really be two PRs, but unfortunately these two things ended up a bit tangled together (I was using the visualisation code to test the robot positioning code and fixing it up as I went along). Sorry! I hope it's not too much of a mess.

The event handling for the SFML code was messy and broken as it was: keyboard events were returned from the ``update()`` method and mouse clicks were recorded and the information could then later be retrieved. The problem is that multiple events can occur between vsyncs, so it ends up being messy trying to pick out and handle these events specifically. Instead I swapped to an interface where you can pass a lambda for event handling (which is then run inside the loop) so the caller can do whatever it wants.

Commit summary:
- Put SFML visualisation code into own namespace
- Convert remaining users of SDL-based visualiser to SFML and drop SDL support
- Fix event handling for SFML
- Tidy the interface to the ``PurePursuitController``, which we're using for data collection (no functional changes intended)
- Add ``auto_dataset_recorder`` program for automatically collecting data by driving the robot along a previously recorded route

Still to do:
- Add a simple way to collect data at a fixed distance from a previously recorded route (e.g. 1m to the left)


alexdewar:
@tenxlenx The readme file for the new dataset_recorder stuff is here: https://github.com/BrainsOnBoard/bob_robotics/blob/rc_auto_route_collection/projects/dataset_recorder/README.md

neworderofjamie:
Can we maybe merge #262 first, merge master in this and then review? as is, it's pretty hard to separate

alexdewar:
Yup, deffo. That was my intention. I just wanted to open it as a separate PR because I wanted @tenxlenx to test this code on the robot.

examples/obstacle_circumnavigation/simulated_obstacle_avoiding_positioner.cc
```diff
@@ -80,34 +81,43 @@ int bobMain(int, char **)
...
+
+            if (runPositioner && event.type == sf::Event::MouseButtonPressed
+                && event.mouseButton.button == sf::Mouse::Left) {
+
+                // set a new goal position if user clicks in the window
+                const auto vec = display.pixelToVector(event.mouseButton.x, event.mouseButton.y);
 
                 // Set the goal to this position
-                const Pose2<meter_t, radian_t> goalPose{ mousePosition.x(), mousePosition.y(), 15_deg };
+                const Pose2<meter_t, radian_t> goalPose{ vec.x(), vec.y(), 15_deg };
```
neworderofjamie, line 106: is this an arbitrary angle?

examples/pure_pursuit_controller/pure_pursuit_controller.cc
```diff
@@ -16,187 +16,202 @@
...
+        if (state == Invalid) {
+            return true;
         }
 
-        // draw lines between waypoints to form a path
-        if (rekt_list.size() > 1) {
-            drawLinesBetweenRects(rekt_list, display.getRenderer());
-        }
+        switch (event) {
+        case Event::Enter:
```
neworderofjamie, line 120: Ifs would probably make sense here - imo swith is only preferable when each case is a one liner otherwise it's easy to lose track of the stupid fall-through

include/robots/control/pure_pursuit_controller.h
```diff
@@ -30,44 +31,40 @@ class PurePursuitController {
...
     //! sets the lookahead distance. Large values causes the car to cut corners,
     //! where small values makes the car follow the path more closely
-    void setlookAheadDistance(const millimeter_t distance);
+    void setLookAheadDistance(const millimeter_t distance);
 
     //! sets the distance between the car's wheel bases
     void setWheelBaseLength(const millimeter_t length);
 
     //! sets the stopping distance. If the car is within this distance, the controller stops
     void setStoppingDistance(const millimeter_t distance);
```
neworderofjamie, line 47: some pointless ``const`` here

include/viz/sfml/robot_control.h
```diff
@@ -0,0 +1,130 @@
...
+drive(RobotType &robot, const sf::Event &event, SpeedType maxSpeed = 1.f,
+      TurnType maxTurn = 1.f)
+{
+    /*
+     * We need to keep track of these values between calls. This is a slightly
+     * gross way of doing things, but making a templatable RobotControl class
+     * to manage state instead seems grosser.
+     */
+    static SpeedType velocity{};
+    static TurnType turn{};
```
neworderofjamie, line 77: What's the purpose of this? It seems a really bad idea as the values will be shared between multiple robots e.g. in a simulation. Why not just cache the velocity and turn values in the robot class (like tank does)? They're also useful for dead reckoning. Also, why doesn't the Ackermann robot just keep doing whatever the last command was if you don't call drive like all the other robot types?

projects/dataset_recorder/auto_dataset_recorder.cc
```diff
@@ -0,0 +1,174 @@
...
+                const auto data = m_GPS.read();
+                if (data && data->gpsQuality != GPS::GPSQuality::INVALID) {
+                    return MapCoordinate::latLonToUTM(data->coordinate);
+                }
+
+                std::this_thread::sleep_for(ReadDelay);
+            }
+
+            throw std::runtime_error{ "Could not get initial GPS fix" };
+        }();
```
neworderofjamie, line 114: why is this a lambda function? Just so you can make ``utmStart`` const?

projects/dataset_recorder/auto_dataset_recorder_simulated.cc
```diff
@@ -0,0 +1,91 @@
...
+
+using namespace BoBRobotics;
+using namespace units::angle;
+using namespace units::length;
+using namespace units::velocity;
+using namespace units::literals;
+using namespace units::math;
+namespace plt = matplotlibcpp;
+
+class RobotArrow
```
neworderofjamie, line 23: Why is this a class rather than just a function? ``plt::quiver`` is going to surely do more than enough memory allocating of its own to make any attempt to save allocations of ``m_X``, ``m_Y`` etc worthwhile. If you really care just have the vectors at ``bobMain`` scope and do this manually in the one place you actually draw an arrow!

projects/dataset_recorder/dataset_recorder.cc
```diff
@@ -1,100 +1,23 @@
...
-}
+using namespace units::literals;
 
 int
 bobMain(int argc, char *argv[])
 {
     CLI::App app{ "Record data with RC car robot" };
     auto *useSystemClock = app.add_flag("--use-system-clock",
                     "Use the system clock to get current date rather than GPS");
+    app.allow_extras();
```
neworderofjamie, line 20: why not just parse runtime as an optional argument?

projects/dataset_recorder/robot_recorder.h
```diff
@@ -0,0 +1,225 @@
...
+        using namespace units::angle;
+        using namespace units::length;
+        using namespace units::time;
+
+        // Read image from camera (synchronously)
+        m_Camera->readFrameSync(m_Frame);
+
+#ifdef DUMMY_CAMERA
+        // RandomInputs don't block, so throttle the framerate to 30 fps
+        std::this_thread::sleep_for(std::chrono::milliseconds(33));
```
neworderofjamie, line 75: maybe this would be a good feature to include in ``Video::RandomInput`` instead?

projects/dataset_recorder/robot_recorder.h
```diff
@@ -0,0 +1,225 @@
...
+        try {
+            std::tie(botSpeed, turnAngle) = robot.readRemoteControl();
+        } catch (std::exception &e) {
+            // if we can't read speed or angle, we just write nan values
+            LOGE << "Could not read speed and steering angle from robot : " << e.what();
+        }
+
+        // Poll for GPS data
+        if (const auto optData = m_GPS.read()) {
+            const auto &gpsData = optData.value();
```
neworderofjamie, line 104: should this be ``m_GPSData = optData.value()``? If not, I don't understand where ``m_GPSData`` is getting set

projects/dataset_recorder/robot_recorder.h
```diff
@@ -0,0 +1,225 @@
...
+        return std::make_unique<Video::RandomInput<>>(cv::Size{ 360, 100 });
+#else
+        return Video::getPanoramicCamera();
+#endif
+    }
+
+    units::frequency::hertz_t getFrameRate() const
+    {
+#ifdef DUMMY_CAMERA
+        return units::frequency::hertz_t{ 30 };
```
neworderofjamie, line 165: Again, it would be much cleaner to just add this to ``Video::RandomInput`` (or make a derived version just used here or something) what with this already being a virtual method

projects/dataset_recorder/robot_recorder.h
```diff
@@ -0,0 +1,225 @@
...
+     */
+    const std::tm &getCurrentDateTime(bool useSystemClock) const
+    {
+        using namespace std::literals;
+
+        auto currentTime = m_GPS.getCurrentDateTime();
+
+        // We already have current date from GPS
+        if (currentTime.tm_year) {
+            goto out;
```
neworderofjamie, line 189: Why not just add:

```c++
std::tuple<int, int, int> getCurrentDate(bool useSystemClock)
{
     if (useSystemClock) {
        LOGW << "Using system clock to get current date. This may not be correct!";
        const auto rawTime = std::time(nullptr);
        const auto &systemTime = *gmtime(&rawTime);
        return std::make_tuple(systemTime.tm_mday, 
                               systemTime.tm_mon,
                               systemTime.tm_year);
    }
    else {
        Stopwatch sw;
        sw.start();
        while (sw.elapsed() < 5s) {
            currentTime = m_GPS.getCurrentDateTime();
            if (currentTime.tm_year) {
                return std::make_tuple(currentTime.tm_mday, 
                                       currentTime.tm_mon,
                                       currentTime.tm_year);
            }
        }

        // ...and if we didn't, give up
        throw std::runtime_error{ "Timed out waiting for date from GPS" };
    }
}
```
and then, in ``getCurrentDateTime``:
```c++
if(!currentTime.tm_year) {
    std::tie(currentTime.tm_mday, 
             currentTime.tm_mon,
             currentTime.tm_year) = getCurrentDate(useSystemClock);

}
```
You could also go a bit less modern C++ and pass ``currentTime`` by reference and set the day, month and year fields directly 🤷

src/robots/control/pure_pursuit_controller.cc
```diff
@@ -1,179 +1,186 @@
...
-                    lookaheadPoint.x() = x2 + x;
-                    lookaheadPoint.y() = y2 + y;
-                    didGetIntersection = true;
-                }
-            }
+    BOB_ASSERT(m_wayPoints.size() > 1);
+
+    std::experimental::optional<Vector2<millimeter_t>> lookAheadPoint;
+    for (unsigned int i = 0; i < m_wayPoints.size() - 1; i++) {
+        // path points
```
neworderofjamie, line 112: It would be really nice to have a little bit more explanation of this maths

neworderofjamie (commented):
All minor comments aside from one bug I _think_ I spotted

arduino/i2c_slave_code_rccar/i2c_slave_code_rccar.ino
```diff
@@ -78,16 +78,17 @@ drive(volatile Movement &m)
 void
 readRemote()
 {
-  int curTurn = pulseIn(STEERING, HIGH, 0);
-  int curSpeed = pulseIn(THROTTLE, HIGH, 0);
+  // Timeout for reading pulses (in microsecs)
+  constexpr auto Timeout = 500000;
 
-  // curTurn is zero if controller not connected
```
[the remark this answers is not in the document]
alexdewar: No, it is (and I should probably put it back in). The 0 in this case indicates that ``pulseIn()`` timed out

examples/obstacle_circumnavigation/simulated_obstacle_avoiding_positioner.cc
alexdewar: Yeah. I didn't want it to be exactly 0° because it's harder to tell if things are broken that way

examples/pure_pursuit_controller/pure_pursuit_controller.cc
alexdewar: That's less of a problem with recent versions of gcc though. They give you a warning if you don't have ``// falls through`` or similar in there. So our CI will pick up these sorts of problems.

projects/dataset_recorder/robot_recorder.h
alexdewar: Good plan

projects/dataset_recorder/robot_recorder.h
alexdewar: Oooops! Good spot

projects/dataset_recorder/robot_recorder.h
alexdewar: True! Done

projects/dataset_recorder/robot_recorder.h
alexdewar: The only problem with this is that it doesn't get the timezone info. The timezone will almost certainly be UTC anyway, but we do at least parse that info...

src/robots/control/pure_pursuit_controller.cc
alexdewar: There is *some* description in the body of the loop, but tbh I just tweaked @tenxlenx's code here so I didn't write it from scratch... @tenxlenx do you have a reference to the paper you used for this that we could stick in a comment?

projects/dataset_recorder/robot_recorder.h
alexdewar: Actually I could just ensure that it is UTC with a ``BOB_ASSERT()`` and then do it your cleaner way....

include/viz/sfml/robot_control.h
alexdewar: I actually removed the value-caching from ``TankBase`` as I felt that having the extra state in there was making things more complicated (i.e. all subclasses then had to make sure that they told ``TankBase`` when they were updating the motor commands).

The Ackermann robot *will* carry on doing whatever the last command was (``velocity`` and ``turn`` are only zero-initialised in the first call to ``drive()``). I admit that it is a bit gross to have this hidden state lurking in ``drive()`` *however* I don't really want to make an ``AckermannBase`` class which carries around this state either :shrug:. (I suppose the other option would be to make a templated class purely for the robot control.) The "multiple robots" problem is *really* unlikely to be an issue because we're never going to be controlling multiple robots with this interface anyway.

projects/dataset_recorder/auto_dataset_recorder.cc
alexdewar: Partly. But it does also make the logic a bit clearer because we can do a ``return`` in that loop body. An alternative would be to turn this into a separate member function.

projects/dataset_recorder/robot_recorder.h
alexdewar: It occurs to me that another way around the ``goto`` ugliness is with the immediately invoked lambda idiom again :wink:

projects/dataset_recorder/auto_dataset_recorder_simulated.cc
alexdewar: True true

projects/dataset_recorder/dataset_recorder.cc
alexdewar: Good point. That would make the logic cleaner too

examples/obstacle_circumnavigation/simulated_obstacle_avoiding_positioner.cc
neworderofjamie: makes sense, might be worth a comment though

projects/dataset_recorder/robot_recorder.h
neworderofjamie: Yey! I actually spotted a bug!

projects/dataset_recorder/robot_recorder.h
neworderofjamie: your choice! I feel especially with the option of the immediately invoked lambda idiom and this being an exception-allowing codebase, there's not many reasons for a goto

include/video/randominput.h
```diff
@@ -1,40 +1,56 @@
...
+    {
+        m_FrameTimer.start();
+    }
 
     virtual bool readFrame(cv::Mat &outFrame) override
     {
         outFrame.create(m_Size, CV_8UC3);
         fillRandom(outFrame.data, outFrame.data + m_Size.area() * 3);
+
+        waitForFrameDelay();
```
neworderofjamie, line 44: shouldn't these non-sync methods just return false if the stopwatch hasn't reached the delay time? Then the generic method will deal with the waiting for you.

alexdewar: True! I wasn't thinking straight

PR #266, opened 2021-09-14
Fix building BoB robotics projects that aren't in git repos

### LINKED ISSUE #264: Build fails if you create a BoB robotics project not in git

alexdewar:
Fixes #264.

Calling return() in a macro returns from the *parent* scope, which in this case has the effect of aborting BoB_build() before modules etc. have been added, leading to linking errors in the case of projects not in git repos.


neworderofjamie (approved):
Damn, that's a subtle one...It _looks_ like cmake does also support functions so might it better to implement this sorta stuff as functions? It doesn't really look like there's any difference between the two aside from scope

alexdewar:
Yeah, I don't think there is. I've got a fairly random mix of ``macro``s and ``function``s in there atm... The reason I was using macros was to set variables in the parent scope (you don't have return values from CMake functions), but it seems like you can do that anyway. Maybe in future I should port all the macros to being functions...

PR #267, opened 2021-09-14
Fix builds using OpenCV 4.3+ on CMake <= 3.10

### LINKED ISSUE #265: Build issues with OpenCV > 4.2.0 on Ubuntu 18
4.2.0 works fine but, something deep in the build system for subsequent versions overrides the C++ standard and breaks builds with cmake <= 3.10.2. Based on a diff of the cmake directories of cmake 4.2.0 and 4.30 I would be particularly suspicious of this change in OpenCVConfig.cmake.in:

```cmake
# Require C++11 features for OpenCV modules
if(CMAKE_VERSION VERSION_LESS "3.1")
  if(NOT OpenCV_FIND_QUIETLY AND NOT OPENCV_HIDE_WARNING_COMPILE_FEATURES)
    message(STATUS "OpenCV: CMake version is low (${CMAKE_VERSION}, required 3.1+). Can't enable C++11 features: https://github.com/opencv/opencv/issues/13000")
  endif()
else()
  set(__target opencv_core)
  if(TARGET opencv_world)
    set(__target opencv_world)
  endif()
  set(__compile_features cxx_std_11)  # CMake 3.8+
  if(DEFINED OPENCV_COMPILE_FEATURES)
    set(__compile_features ${OPENCV_COMPILE_FEATURES})  # custom override
  elseif(CMAKE_VERSION VERSION_LESS "3.8")
    set(__compile_features cxx_auto_type cxx_rvalue_references cxx_lambdas)
  endif()
  if(__compile_features)
    # Simulate exported result of target_compile_features(opencv_core PUBLIC ...)
    set_target_properties(${__target} PROPERTIES
        INTERFACE_COMPILE_FEATURES "${__compile_features}"
    )
  endif()
  unset(__target)
  unset(__compile_features)
endif()
```
which was added just after the 4.2.0 release (https://github.com/opencv/opencv/commit/01c1b341a881ba49d538acdf29d49ca7fd8ada57#diff-452ddb3443d10f7a9d257cfa21a76135a63c36ebcf21ed8355531dd627ff2329). I can also confirm, if I insert that code into a working OpenCV 4.2.0 installation's OpenCVConfig.cmake, it breaks.

alexdewar:
Fixes #265.

Without this, the C++ standard is forced to C++11 on CMake <= v3.10, which causes build errors for BoB robotics code.

### DIFF
--- a/cmake/external_libs/opencv.cmake
+++ b/cmake/external_libs/opencv.cmake
@@ -13,6 +13,12 @@
 #
 # We also provide an extra option to explicitly use pkg-config.
 if(NOT OpenCV_FOUND)
+    # The CMake package file for OpenCV v4.3+ appears to have a bug in which
+    # means that the C++ standard is forced to C++11, which will break BoB
+    # robotics code as we use C++14 features all over the place. Setting this
+    # variable seems to be a good workaround.
+    set(OPENCV_COMPILE_FEATURES "")
+
     if(NOT FIND_OPENCV_WITH_PKGCONFIG)
         find_package(OpenCV QUIET)
     endif()


neworderofjamie:
Have you verified this works on a VM or whatever or do you want me to have a go? I'm in the joyous position of having multiple, freshly-build OpenCVs on my machine...

alexdewar:
Yes, I checked that it worked in an Ubuntu 18.04 Docker image with OpenCV v4.5.3 built from source. Sorry, should have mentioned that!

alexdewar:
Side note: I'm thinking it probably makes sense to publish these docker images so that users can just do a ``docker pull`` to get the correct build tools etc. I've published this image to my docker hub: https://hub.docker.com/repository/docker/alexdewar/ubuntu1804-bob

I guess we could make a BoB docker hub repository if we really wanted. We could also push images for Jetsons so that we can build test stuff without having a robot there...

neworderofjamie (approved):

neworderofjamie:
Actually, have you checked this doesn't upset older versions of OpenCV? Might be worth only applying if version > 4.2.0 (if CMake provides you that info)

alexdewar:
I do actually have a working image that does this! You just stick a statically compiled version of qemu in the root of your image and execute it that way.

alexdewar:
> Actually, have you checked this doesn't upset older versions of OpenCV? Might be worth only applying if version > 4.2.0 (if CMake provides you that info)

There's a bit of a chicken-and-egg problem here: you can only figure out which version of OpenCV the user had once you've included the package, by which time things will already be broken :grimacing:

neworderofjamie:
Ahh of course - just checked and it works fine on OpenCV 4.2.0 and I did a quick search of the OpenCV github at the last 3.X tag and ``OPENCV_COMPILE_FEATURES`` isn't used anywhere so I'm sure it'll be fine

alexdewar:
Great!

PR #269, opened 2021-09-27
Fix image path in examples/perfect_memory

alexdewar:
Commit 530e873b changed the code to use ``Path::getRepoPath()`` to get to the root of the BoB robotics repo, but in doing so accidentally changed the name of the file, breaking the HOG example.


neworderofjamie (approved):
doh

PR #270, opened 2021-09-30
Windows fixes

neworderofjamie:
My recurring PR of random Windows fixes 😄 This time:

- Added some missing ``#ifdef __linux__`` guards
- Added magic option so ``<cmake>`` defines ``M_PI``
- Not a windows fix but, I'm sick of building opencv_contrib just for optical flow so I've added an ifdef around this module so it only gets built if you have built the requisite opencv_contrib module


alexdewar (approved):
This all looks sensible!

alexdewar:
Test comment

PR #275, opened 2021-11-08
Use a standard CMake package format for BoB robotics

### LINKED ISSUE #273: Fix build process for Python AntWorld on Windows
Currently the various DLLs need to be bundled manually otherwise it won't work. A related problem is that the readme file isn't particularly descriptive, making this even more of a faff for other users to build.

### LINKED ISSUE #276: Phantom duplicate OpenCV libraries linked to Visual Studio project
The project generated by cmake has two sets of Open CV libraries:
```
opencv_videoio.lib
opencv_video.lib
opencv_stitching.lib
opencv_photo.lib
opencv_objdetect.lib
opencv_ml.lib
opencv_imgproc.lib
opencv_imgcodecs.lib
opencv_highgui.lib
opencv_flann.lib
opencv_features2d.lib
opencv_dnn.lib
opencv_core.lib
opencv_calib3d.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_video.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_stitching.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_photo.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_objdetect.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_ml.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_highgui.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_videoio.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_imgcodecs.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_dnn.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_calib3d.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_features2d.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_imgproc.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_flann.lib
C:\Users\jamie\Documents\programming\vcpkg\installed\x64-windows\lib\opencv_core.lib
```
In order to build the project, you need to manually remove the first set. As a Visual Studio user, you wouldn't really expect full paths here. Typically you'd add the libraries by name and add the path they're in to the project's list of library directories...

alexdewar:
Currently we're doing a whole host of nasty hacks in macros like ``BoB_project()`` to try to pass dependencies between projects and modules etc. around. It seems like a better solution to work with CMake rather than fighting it :smile: 

It turns out we can get all the functionality we need by setting the compile properties of modules appropriately (e.g. setting include dirs and extra link flags to OpenCV etc.) and those properties are inherited by targets we create (e.g. executables).

The bulk of the work is now done, but there's no way this will pass CI checks yet. I'm just pushing it here so I can run the CI over it.

Fixes #276 
Fixes #273


alexdewar:
Fixes #273.

alexdewar:
Fixes #276.

cmake/BoBRobotics.cmake
```diff
@@ -0,0 +1,102 @@
...
+        include(CheckCXXCompilerFlag)
+        foreach(FLAG IN LISTS DEFAULT_COMPILER_FLAGS)
+            check_cxx_compiler_flag(${FLAG} COMPILER_SUPPORTS_${FLAG})
+            if(${VARNAME})
+                target_compile_options(bob_base INTERFACE -${FLAG})
+            endif()
+        endforeach(FLAG IN LISTS DEFAULT_COMPILER_FLAGS)
+    endif(NOT MSVC)
+
+    if(WIN32)
```
neworderofjamie, line 64: it might be more consistent to use one or other of ``WIN32`` and ``MSVC``

alexdewar: This was actually intentional -- some of those macros are Windows-specific rather than MSVC-specific -- although now you mention it I realise there is a mix of the two.

cmake/FindGeNN.cmake
```diff
@@ -0,0 +1,160 @@
...
+    endif()
+endmacro()
+
+# Let users give an explicit path to GeNN
+if(GENN_PATH)
+    set(GeNN_ROOT_DIR "${GENN_PATH}")
+    set(GeNN_BUILDMODEL "${GENN_PATH}/bin/genn-buildmodel.sh")
+else()
+    # Find genn-buildmodel (which should be in the path)
+    find_program(GeNN_BUILDMODEL genn-buildmodel.sh)
```
neworderofjamie, line 17: it's ``genn-buildmodel.bat`` on windows

neworderofjamie (commented):
my CMakeFU isn't up to really reviewing this in any depth but it works nicely for me. My only question is, how come it now produces executables in the build directories?

cmake/FindGeNN.cmake
alexdewar: Good point!

alexdewar:
Thanks for the review! I forgot to mention about the binaries-in-build-folder thing: essentially that's the normal thing to do in CMake and the way we were doing it before was odd. With CMake, in theory all your build artefacts including executables etc. should end up in your build folder and it kind of makes sense really (e.g. it means you can have separate ``build_debug`` and ``build_release`` folders).

neworderofjamie:
Also, are FindSFML.cmake, FindTBB.cmake etc provided by a third party? I don't quite understand who has responsibility for providing these e.g. CMake itself ships with some, OpenCV generates its own etc

alexdewar:
There are two ways you can provide a package that CMake can find with ``find_package``: https://cmake.org/cmake/help/latest/command/find_package.html

Programs can either ship their own, e.g. OpenCV provides an OpenCVConfig.cmake file, which essentially just return paths to include directories etc. (they're generated by CMake itself at build time). Alternatively you can have a bunch of hacks to try to figure out these things and put them in a ``FindMyLibrary.cmake``-type file (and CMake comes with a bunch of these for finding Python libraries etc.).

I pinched some of the ``Find*.cmake`` files from the web and wrote a couple myself. (Incidentally the GeNN one now works with ``SharedLibraryModel`` projects :smile: )

PR #277, opened 2021-11-16
CMake: Re-add OpenMP support when building with Eigen

alexdewar:
After the big CMake tidy-up, I accidentally removed this, meaning that Eigen was only running single threaded.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.10)
 include(../../cmake/BoBRobotics.cmake)
-project(common)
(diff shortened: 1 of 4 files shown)


alexdewar:
I'm not entirely sure why Jenkins is complaining but I'll dig into this. I'm assuming the failure is a spurious one as this is only a small simple change.

neworderofjamie (approved):
Jenkins notwithstanding, this seems entirely sensible - I really like the more explicit, less magical new CMake....

alexdewar:
Cool, thanks.

Re Jenkins: Did you give my PC a prod in the end? It's still not responding :cry:

PR #278, opened 2021-11-17
Convert snapshot_bot to use ``ImageDatabase``s

### LINKED ISSUE #274: Use standard image database format for snapshot_bot
Currently snapshot_bot uses its own bespoke format for saving image databases. It would be better to use the standard ``ImageDatabase`` method for saving the training data so that we can use the standard tools (e.g. navbench) to analyse them after the fact.

It turns out @efkag has been using this tool to record data, so it would be convenient if we could standardise things.

alexdewar:
Closes #274.

The main reason for this change is that @efkag has been using ``snapshot_bot`` to record image databases and it would be convenient if any future datasets were in the standard ``ImageDatabase`` format to make it easier to analyse with our standard tools (e.g. navbench). An added bonus, though, is that it should also make analysing any navigation-related issues easier (e.g. by using a quiver plot from navbench).

### DIFF
--- a/projects/snapshot_bot/config.h
+++ b/projects/snapshot_bot/config.h
@@ -12,7 +12,6 @@
 #include <opencv2/opencv.hpp>
 
 // Standard C++ includes
-#include <chrono>
 #include <limits>
 #include <map>
 #include <string>
@@ -22,7 +21,7 @@
 //------------------------------------------------------------------------
 class Config
 {
-    using Milliseconds = std::chrono::duration<double, std::milli>;
+    using millisecond_t = units::time::millisecond_t;
     using WindowConfig = BoBRobotics::Navigation::PerfectMemoryWindow::DynamicBestMatchGradient::WindowConfig;
 
 public:
@@ -34,7 +33,7 @@ public:
       , m_SaveTestingDiagnostic(false)
       , m_StreamOutput(false)
       , m_ODK2(false)
-      , m_MaxSnapshotRotateDegrees(180.0)
+      , m_MaxSnapshotRotate(180.0)
       , m_PMFwdLASize(std::numeric_limits<size_t>::max())
       , m_PMFwdConfig{ 0, 0, 0, 0 }
       , m_UnwrapRes(180, 50)
@@ -49,7 +48,7 @@ public:
       , m_SnapshotServerListenPort(BoBRobotics::Net::Connection::DefaultListenPort + 1)
       , m_BestSnapshotServerListenPort(BoBRobotics::Net::Connection::DefaultListenPort + 2)
       , m_MoveSpeed(0.25)
-      , m_TurnThresholds{ { units::angle::degree_t(5.0), { 0.5f, Milliseconds(500.0) } }, { units::angle::degree_t(10.0), { 1.0f, Milliseconds(500.0) } } }
+      , m_TurnThresholds{ { units::angle::degree_t(5.0), { 0.5f, millisecond_t(500.0) } }, { units::angle::degree_t(10.0), { 1.0f, millisecond_t(500.0) } } }
       , m_UseViconTracking(false)
       , m_ViconTrackingPort(0)
       , m_ViconTrackingObjectName("norbot")
@@ -69,7 +68,7 @@ public:
     bool shouldStreamOutput() const{ return m_StreamOutput; }
     bool shouldUseODK2() const{ return m_ODK2; }
 
-    units::angle::degree_t getMaxSnapshotRotateAngle() const{ return units::angle::degree_t(m_MaxSnapshotRotateDegrees); }
+    units::angle::degree_t getMaxSnapshotRotateAngle() const{ return units::angle::degree_t(m_MaxSnapshotRotate); }
 
     size_t getPMFwdLASize() const{ return m_PMFwdLASize; }
     const WindowConfig &getPMFwdConfig() const{ return m_PMFwdConfig; }
@@ -85,8 +84,8 @@ public:
     float getJoystickDeadzone() const{ return m_JoystickDeadzone; }
 
     bool shouldAutoTrain() const{ return m_AutoTrain; }
-    Milliseconds getTrainInterval() const{ return m_TrainInterval; }
-    Milliseconds getMotorCommandInterval() const{ return m_MotorCommandInterval; }
+    millisecond_t getTrainInterval() const{ return m_TrainInterval; }
+    millisecond_t getMotorCommandInterval() const{ return m_MotorCommandInterval; }
 
     bool shouldUseViconTracking() const{ return m_UseViconTracking; }
     int getViconTrackingPort() const{ return m_ViconTrackingPort; }
@@ -104,7 +103,7 @@ public:
 
     float getMoveSpeed() const{ return m_MoveSpeed; }
 
-    std::pair<float, Milliseconds> getTurnSpeed(units::angle::degree_t angleDifference) const
+    std::pair<float, millisecond_t> getTurnSpeed(units::angle::degree_t angleDifference) const
     {
         const auto absoluteAngleDifference = units::math::fabs(angleDifference);
 
@@ -117,7 +116,7 @@ public:
         }
 
         // No turning required!
-        return std::make_pair(0.0f, Milliseconds(0.0));
+        return std::make_pair(0.0f, millisecond_t(0.0));
     }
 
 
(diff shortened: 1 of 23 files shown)


projects/snapshot_bot/memory.cc
```diff
@@ -1,26 +1,31 @@
...
 #include "config.h"
+#include "image_input.h"
 
 using namespace BoBRobotics;
 using namespace units::angle;
 using namespace units::length;
 using namespace units::literals;
 using namespace units::math;
 
+constexpr const char *CSVBestHeading = "Best heading [degrees]";
```
neworderofjamie, line 23: I'm not 100% on the rules for ``constexpr`` objects but I think these should probably be in the anonymous namespace so they're only module scope

projects/snapshot_bot/memory.h
```diff
@@ -10,19 +10,31 @@
...
+
+namespace Navigation
+{
+class ImageDatabase;
+}
 }
 
+// Bounds used for extracting masks from ODK2 images
+const cv::Scalar odk2MaskLowerBound(1, 1, 1);
+const cv::Scalar odk2MaskUpperBound(255, 255, 255);
```
neworderofjamie, line 36: won't these be instantiated in every compilation unit and cause multiple definition linker errors?

third_party/path.h
```diff
@@ -350,6 +350,13 @@ inline void copy_file(const filesystem::path &from, const filesystem::path &to)
     ofs << ifs.rdbuf();
 }
 
+inline filesystem::path
+current_path()
+{
+    char cwd[PATH_MAX];
+    return getcwd(cwd, sizeof(cwd));
```
neworderofjamie, line 357: Pretty sure this is POSIX only

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -474,7 +379,7 @@ class RobotFSM : FSM<State>::StateHandler
                         m_BestSnapshotNetSink->sendFrame(perfectMemory->getBestSnapshot());
                     }
                     else {
-                        LOGW << "WARNING: Can only stream output from a perfect memory";
+                        LOGW << "Can only stream output from a perfect memory";
```
neworderofjamie, line 382: It's like it was previously just printing to ``std::cout`` or something :rofl:

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -513,13 +416,13 @@ class RobotFSM : FSM<State>::StateHandler
                 // If A is pressed
                 if(m_Joystick.isPressed(HID::JButton::A)) {
                     // Subtract time we've already moved for from drive time
-                    m_DriveTime -= m_MoveStopwatch.elapsed();
+                    m_DriveTime -= static_cast<millisecond_t>(m_MoveStopwatch.elapsed());
```
neworderofjamie, line 419: how come this needs a cast? Doesn't the stopwatch also return a time unit?

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -513,13 +416,13 @@ class RobotFSM : FSM<State>::StateHandler
...
 
                     // Transition to correct paused state
                     m_StateMachine.transition((state == State::DrivingForward) ? State::PausedDrivingForward : State::PausedTurning);
                 }
                 // Otherwise, if drive time has passed
-                else if(m_MoveStopwatch.elapsed() > m_DriveTime) {
+                else if(static_cast<millisecond_t>(m_MoveStopwatch.elapsed()) > m_DriveTime) {
```
neworderofjamie, line 425: Likewise

neworderofjamie: If it doesn't and it's using ``std::chrono`` types, why not just stick with those - they're not used in complex units-based calculations or anything!

projects/snapshot_bot/config.h
```diff
@@ -20,16 +21,39 @@
...
+      , m_ServerListenPort(BoBRobotics::Net::Connection::DefaultListenPort)
+      , m_SnapshotServerListenPort(BoBRobotics::Net::Connection::DefaultListenPort + 1)
+      , m_BestSnapshotServerListenPort(BoBRobotics::Net::Connection::DefaultListenPort + 2)
+      , m_MoveSpeed(0.25)
+      , m_TurnThresholds{ { units::angle::degree_t(5.0), { 0.5f, millisecond_t(500.0) } }, { units::angle::degree_t(10.0), { 1.0f, millisecond_t(500.0) } } }
+      , m_UseViconTracking(false)
+      , m_ViconTrackingPort(0)
+      , m_ViconTrackingObjectName("norbot")
+      , m_UseViconCaptureControl(false)
+      , m_ViconCaptureControlPort(0)
```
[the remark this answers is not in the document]
alexdewar: It's what clang-format does, not that that means it's a good idea! I find having all the initialisers squished onto a couple of lines a bit annoying when it comes to adding values (like I did here), but I'm not precious about it so happy to change back.

projects/snapshot_bot/config.h
```diff
@@ -170,13 +192,13 @@ class Config
         // **YUCK** why does OpenCV (at least my version) not have a cv::read overload for std::string!?
         cv::String outputPath;
         cv::read(node["outputPath"], outputPath, m_OutputPath.str());
-        m_OutputPath = (std::string)outputPath;
-
-        cv::String testingSuffix;
-        cv::read(node["testingSuffix"], testingSuffix, m_TestingSuffix);
-        m_TestingSuffix = (std::string)testingSuffix;
```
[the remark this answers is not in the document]
alexdewar: Sorry, I should have mentioned this in the PR...

The reason for removing it was that I've changed the test data to also be in an image database format (though there may well not be actual images stored in there) and by default ``ImageDatabase``s are initialised with a timestamp so they're unique. So you'd have e.g. a folder ``training`` with your snapshots in and then your test data would be in ``training/testing/20211117_150400_001`` or whatever. (I think it makes sense to have the testing data as a subfolder in with the training data, but we could also change this.)

It would be sensible to at least the option have more explicit and readable names though. Would it be better to keep them unique (e.g. have a ``testingPrefix`` option) or just have the whole name be specified?

projects/snapshot_bot/memory.cc
```diff
@@ -29,14 +34,50 @@ MemoryBase::MemoryBase()
...
+void MemoryBase::trainRoute(Navigation::ImageDatabase &route,
+                            bool useODK2, ImageInput &imageInput)
 {
-    os << getBestHeading().value() << ", " << getLowestDifference();
+    LOGI << "Loading stored snapshots...";
+    std::vector<std::pair<cv::Mat, ImgProc::Mask>> snapshots(route.size());
+
+    // Load and process images in parallel
+    // **TODO**: Add support for static mask images
+    std::atomic_int loadCount{ 0 };
```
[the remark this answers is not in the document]
alexdewar: Nope! Good spot.

projects/snapshot_bot/memory.h
alexdewar: It does compile ok for me. But you're right that this is wrong! It might break only if they're not inlined...

third_party/path.h
alexdewar: It does work on Windows (though it's technically deprecated): https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getcwd?view=msvc-170

projects/snapshot_bot/snapshot_bot.cc
alexdewar: It doesn't, though maybe it should!

projects/snapshot_bot/snapshot_bot.cc
alexdewar: Good point. I sort of hoped that it would end up making the code tidier but that didn't pan out...

projects/snapshot_bot/config.h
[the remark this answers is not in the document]
neworderofjamie: fair enough - my desire to fight clang-format is probably not mega-helpful.....

projects/snapshot_bot/config.h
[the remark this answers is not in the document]
neworderofjamie: ahhh I see - realistically, that probably makes not over-writing your old data rather less error-prone :smile: it might be nice to have an optional prefix or suffix as well though

neworderofjamie:
One thing, before this is merged is that it'd be good to check it works on the actual robot. If you can point me in the direction of the cmake PPA for Ubuntu 16, I'm happy to have a go at this.

alexdewar:
I've pushed the final changes that I wanted to get done on this PR (to fix #281). Other than that, I think I've addressed your comments.

Totally agree that we need to test this on a real robot. We should also probably test it for a few different configurations (e.g. with infomax and using the ODK2). There's a PPA here: https://apt.kitware.com/

I'm also happy to have a go myself. I'll probably be in tomorrow or Fri.

neworderofjamie:
Awesome! I'm away on Friday but will be on the office tomorrow

neworderofjamie:
Just trying to install the ppa but the 16.04 repository doesn't support arm64

PR #280, opened 2021-11-17
Fix correlation coefficient differencer's performance

### LINKED ISSUE #268: CorrCoefficient is over 50x slower than AbsDiff
This is on my new workstation with a 16 thread Xeon with 24Mb cache so, I fear, this is going to be too slow for robot use even with windowing. Talking to Stathis and looking at the code that eventually gets called in our situation (https://github.com/opencv/opencv/blob/170bf6d7af9c8bb2b34ec143826594dc84251b89/modules/imgproc/src/templmatch.cpp#L566-L760), it appears that ``cv::templateMatch`` is, as the name suggests, more general and capable of handling different sized input images through some complex DFT based tiled thing. Whereas, Stathis's code does the following:
```python
def nan_correlation_dist(a, b):
    """
    Calculates the correlation coefficient distance
    between two vectors.
    Where a and b can contain nan values.
    :param a:
    :param b:
    :return:
    """
    amu = np.nanmean(a)
    bmu = np.nanmean(b)
    a = a - amu
    b = b - bmu
    ab = np.nanmean(a * b)
    avar = np.nanmean(np.square(a))
    bvar = np.nanmean(np.square(b))
    dist = 1.0 - ab / np.sqrt(avar * bvar)
    return dist
```
and, as you would guess from looking at it, it is only around 5x slower than AbsDiff. I think we could massively optimise ``CorrCoefficient`` by implementing it directly using lower-level integer OpenCV ops.

alexdewar:
Fixes #268.

This *should* be using the same maths as @efkag's python code, rather than the slower ``cv::templateMatch``.

Automated tests are currently failing though, so presumably I've made a mistake somewhere and this in fact isn't doing exactly what it did before.


alexdewar:
@efkag I've made this into a PR for your reference. If you have the chance (and inclination) to have a quick look at this before tomorrow, that's great, but no worries if not.

PR #286, opened 2021-11-18
CMake: Don't use newfangled Eigen3::Eigen target

alexdewar:
BoB robotics code is currently not building on Ubuntu 16.04 because of an Eigen-related issue. It turns out that the version of Eigen included with Ubuntu 16.04 doesn't create it. Just use the EIGEN3_INCLUDE_DIRS variable instead, which seems to work with both old and new versions of Eigen.

Let me know if this fixes things for you @neworderofjamie !


neworderofjamie (approved):
Quick, post Alergic test :+1:

alexdewar:
Cool. Thanks!

PR #291, opened 2021-11-26
Fix compile options

### LINKED ISSUE #290: Horribly slow performance for training InfoMax networks
Currently, training with an image of size 90x25 and 90x25 hidden units, it's taking around 10 secs *per image* to train. I've no idea when this problem started, but it seems to be present on both Arch and Ubuntu (running in docker). By contrast, the numpy-based implementation in navbench takes only around ~200 ms per image.

alexdewar:
Fixes #290.

It turns out that since the big CMake refactor, compiler options haven't been being set appropriately, so all targets were being built without optimisation.


cmake/BoBRobotics.cmake
```diff
@@ -55,8 +77,8 @@ function(make_base_target)
         include(CheckCXXCompilerFlag)
         foreach(FLAG IN LISTS DEFAULT_COMPILER_FLAGS)
             check_cxx_compiler_flag(${FLAG} COMPILER_SUPPORTS_${FLAG})
-            if(${VARNAME})
-                target_compile_options(bob_base INTERFACE -${FLAG})
+            if(COMPILER_SUPPORTS_${FLAG})
```
neworderofjamie, line 80: 1. This is a cool mechanism
2. Whoops :smile:

neworderofjamie (approved):

cmake/BoBRobotics.cmake
alexdewar: Yeah :laughing:. Although this part only caused the warnings to be turned off (which is also not ideal). The main problem here was that CMake didn't bother applying the default build flags to the targets for some reason.

PR #293, opened 2021-11-26
Fix Jenkins issues

alexdewar:


alexdewar:
@neworderofjamie Commit 614bfe2 (CMake: Fix path to genn-buildmodel when GENN_PATH specified) is passing the Jenkins checks but only because it's not actually running any tests on your machine...

I think this is the offending part of the Jenkinsfile:
```groovy
// Build dictionary of available nodes and their labels
def availableNodes = [:]
for(node in jenkins.model.Jenkins.instance.nodes) {
    if(node.getComputer().isOnline() && node.getComputer().countIdle() > 0) {
        availableNodes[node.name] = node.getLabelString().split() as Set
    }
}
```

I guess if your machine is already running a test then it's counted as not idle and the tests are skipped?

neworderofjamie:
So, currently, we don't queue jobs and, right now, both the executers on INF900801 are occupied so it won't get run there. You _could_ remove the ``node.getComputer().countIdle() > 0`` check and then it would queue the tests on any available machines. I really need to figure out why my new workstation isn't connecting to Jenkins so we can get some more executers online...

alexdewar:
Now this should be good to go. I followed your suggestion of removing that check, which I *think* has fixed things.

Besides that, Jenkins was failing to identify warnings because the log file name was wrong, so I've fixed that up. I also fixed a few build and compiler warnings.

alexdewar:
I also added the python antworld module to the CI checks, but it's not running at present as scikit-build doesn't seem to be installed on your machine.

neworderofjamie:
I try and avoid installing python modules globally on my machine so I think the best thing to do in the Jenkins is something like https://github.com/genn-team/genn/blob/master/Jenkinsfile#L226-L234 (without the double ``bdist_wheel`` cos antworld doesn't foolishly use SWIG)

alexdewar:
> I try and avoid installing python modules globally on my machine so I think the best thing to do in the Jenkins is something like https://github.com/genn-team/genn/blob/master/Jenkinsfile#L226-L234 (without the double `bdist_wheel` cos antworld doesn't foolishly use SWIG)

Good idea!

neworderofjamie (approved):

neworderofjamie:
``${env.PYTHON}`` on the Jenkins machine should contain a suitable python 3 executable on the Jenkins machine btw

bin/build_all_jenkins.sh
```diff
@@ -14,5 +14,6 @@ pip3 install numpy scikit-build
 # Configure and build source, continuing if there are errors
 rm -rf build
 mkdir build
-cmake -B build -S . -DCMAKE_EXPORT_COMPILE_COMMANDS=1 "$@"
-make -C build -k -j $(nproc)
+cd build
+cmake -DPYTHON_EXECUTABLE="$(which python3)" -DCMAKE_EXPORT_COMPILE_COMMANDS=1 "$@" ..
```
neworderofjamie, line 18: the virtualenv should symlink ``python`` to whichever python executable you created the environment with so I don't see why this would be necessary unless cmake follows some very sketch python-finding logic

neworderofjamie: I do remember in the previous cmake we used a python-based lookup to find numpy - _maybe_ that was because the cmake python-finding isn't virtualenv-aware (which would be kinda shit)

alexdewar: It finds the python binary correctly on my machine when in a virtualenv but I suspect that it's still preferentially looking for python 2. I've told cmake we want python v3+ everywhere and that seemed to fix it, so let's just deprecate support for python 2.

Issue #289, opened 2021-11-23
snapshot_bot: Add option to save snapshots as video files


alexdewar:
We have the infrastructure for this now and it would be useful, e.g. for cases where we have many thousands of frames.

Issue #287, opened 2021-11-19
snapshot_bot: Add option to record video data at maximum framerate


alexdewar:
At present, it currently trains online, which is useful for demos, but means that the frame rate for recording images can be rather slow due to the time taken to preprocess images and, in the case of Infomax, to train the network. It would be useful to at least have the *option* to record from the camera at 30 fps; then we can choose what to do with that data after the fact.

Suggested by @aphilippides.

PR #294, opened 2021-11-30
Snapshot bot enhancements

alexdewar:
This PR implements a few changes suggested by @aphilippides. It's based on #293, so we should merge that first.

The two main changes are:
1. Do training after recording images so we can record at the camera's maximum frame rate (#287)
2. Add option to record databases as video files (useful for larger datasets, #289)

I also ended up refactoring some ``ImageDatabase`` code so the image and video recording classes are now virtual, but no functional change was intended there.

### DIFF
--- a/include/navigation/image_database.h
+++ b/include/navigation/image_database.h
@@ -389,20 +389,23 @@ public:
         template<class... Ts>
         void record(const Vector3<millimeter_t> &position, degree_t heading, const cv::Mat &image, Ts &&...extraFieldValues)
         {
-            this->addEntry(image, getFileName, position, heading, { 0, 0, 0 }, std::forward<Ts>(extraFieldValues)...);
+            this->addEntry(image, std::bind(&RouteRecorder::getFileName, this),
+                           position, heading, { 0, 0, 0 },
+                           std::forward<Ts>(extraFieldValues)...);
         }
 
         void record(const cv::Mat &image, Entry entry)
         {
-            this->addEntry(image, entry, getFileName);
+            this->addEntry(image, entry, std::bind(&RouteRecorder::getFileName, this));
         }
 
     private:
-        const std::function<std::string()> getFileName = [this]() {
+        std::string getFileName()
+        {
             std::ostringstream ss;
             ss << "image" << std::setw(5) << std::setfill('0') << this->size() + 1;
             return (this->getImageDatabase().getPath() / ss.str()).str();
-        };
+        }
     };
 
     ImageDatabase();
--- a/include/navigation/image_database.h
+++ b/include/navigation/image_database.h
@@ -120,10 +120,8 @@ public:
     class VideoFileWriter
       : public FrameWriter {
     public:
-        VideoFileWriter(const ImageDatabase &,
-                        const std::pair<const std::string &, const std::string &> &format);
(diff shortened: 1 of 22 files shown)


alexdewar:
I've put this as a draft because I've still got to test that it works on the actual robot.

include/navigation/image_database.h
```diff
@@ -374,29 +387,24 @@ class ImageDatabase
...
         void record(const cv::Mat &image, Entry entry)
         {
-            this->addEntry(image, entry);
+            this->addEntry(image, entry, getFileName);
         }
 
-        std::string getCurrentFilenameRoot() const override
-        {
+    private:
+        const std::function<std::string()> getFileName = [this]() {
```
neworderofjamie, line 401: This seems a very odd use of ``std::function``. Why not make this a normal function and bind it when you assign it to a ``std::function`` with ``std::bind(&RouteRecorder::getFileName, this);``

projects/snapshot_bot/config.cc
```diff
@@ -0,0 +1,271 @@
+#include "config.h"
```
neworderofjamie, line 1: Splitting that awful header into a cc is an excellent move and must do wonders for the compile time...

projects/snapshot_bot/config.cc
```diff
@@ -0,0 +1,271 @@
...
+        cv::read(node["bestSnapshotServerListenPort"], m_BestSnapshotServerListenPort, m_BestSnapshotServerListenPort);
+        cv::read(node["autoTrain"], m_AutoTrain, m_AutoTrain);
+        cv::read(node["moveSpeed"], m_MoveSpeed, m_MoveSpeed);
+
+        double trainInterval;
+        cv::read(node["trainInterval"], trainInterval, m_TrainInterval.count());
+        m_TrainInterval = (Milliseconds)trainInterval;
+
+        int frameSkip = m_SkipFrames;
+        cv::read(node["frameSkip"], frameSkip, frameSkip);
```
neworderofjamie, line 154: might be worth making clear exactly when these frames are skipped i.e ``m_TestSkipFrames``

projects/snapshot_bot/config.cc
```diff
@@ -0,0 +1,271 @@
...
+
+// Standard C++ includes
+#include <limits>
+
+using namespace BoBRobotics;
+using Milliseconds = std::chrono::duration<double, std::milli>;
+
+namespace {
+// **YUCK** as well as std::strings, it also seems like OpenCV can't serialise size_t
+int getIntegerSize(size_t value)
```
neworderofjamie, line 15: you could just accept that the OpenCV serialization sucks somewhat, use ``int``/``unsigned int`` and ``static_cast`` when required..

neworderofjamie (commented):
This generally all looks very sensible. Just because of the large number of changes, personally I would be _very_ reticent about using this version for the showcase but up to you!

include/navigation/image_database.h
alexdewar: Ohh good idea. I didn't know about ``std::bind``.

projects/snapshot_bot/config.cc
alexdewar: :+1:

PR #314, opened 2022-02-17
added imu data recording to snapshot bot

FullAPVayne:
It compiles but I didnt run it on the robot yet.

### DIFF
--- a/projects/snapshot_bot/snapshot_bot.cc
+++ b/projects/snapshot_bot/snapshot_bot.cc
@@ -266,7 +266,7 @@ private:
 #ifdef USE_VICON
                         // Get tracking data
                         const auto objectData = m_ViconTracking.getObjectData(m_Config.getViconTrackingObjectName());
-                        const std::array<degree_t, 3> imuData = m_IMU.getEulerAngles();
+                        const std::array<degree_t, 3> imuData = imu->getEulerAngles();
                         m_Recorder->record(objectData.getPose(),
                                            m_Output,
                                            elapsed,
@@ -552,12 +552,12 @@ private:
 #ifdef USE_VICON
     // Vicon tracking interface
     Vicon::UDPClient<Vicon::ObjectData> m_ViconTracking;
-    
+    std::unique_ptr<BN055> imu;
 
     // Vicon capture control interface
     Vicon::CaptureControl m_ViconCaptureControl;
 #endif
-    m_IMU = std::make_unique<BN055>();
+    
     
     // For training data
     ImageDatabase m_TrainDatabase;
--- a/projects/snapshot_bot/snapshot_bot.cc
+++ b/projects/snapshot_bot/snapshot_bot.cc
@@ -68,7 +68,7 @@ class RobotFSM : FSM<State>::StateHandler
 {
     using Milliseconds = std::chrono::duration<double, std::milli>;
     using ImageDatabase = Navigation::ImageDatabase;
-    
+
 public:
     RobotFSM(const Config &config, BackgroundExceptionCatcher &backgroundEx)
       : m_Config(config)
@@ -152,6 +152,9 @@ public:
             throw std::runtime_error("You must rebuild with -DUSE_VICON=ON");
 #endif
         }
+if (m_Config.shouldUseIMU()){
+    m_IMU = std::make_unique<BN055>();
+}
 
         // If we should train
         if(m_Config.shouldTrain()) {
@@ -165,7 +168,7 @@ public:
             // Start directly in testing state
             m_StateMachine.transition(State::WaitToTest);
         }
-    
+
     }
 
     //------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -67,6 +68,7 @@ class RobotFSM : FSM<State>::StateHandler
 {
     using Milliseconds = std::chrono::duration<double, std::milli>;
     using ImageDatabase = Navigation::ImageDatabase;
+    using AngleUnit = units::angle::degree_t;
```
alexdewar, line 71: You're not actually using ``AngleUnit`` anywhere, right? So you don't need this line. This just means "make a new type called ``AngleUnit`` which is an alias for ``units::angle::degree_t``".

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -259,16 +261,18 @@ class RobotFSM : FSM<State>::StateHandler
...
 #ifdef USE_VICON
                         // Get tracking data
                         const auto objectData = m_ViconTracking.getObjectData(m_Config.getViconTrackingObjectName());
+                        const auto imuData = imu.getEulerAngles();
                         m_Recorder->record(objectData.getPose(),
                                            m_Output,
                                            elapsed,
-                                           objectData.getFrameNumber());
+                                           objectData.getFrameNumber(),
+                                           imuData);
```
alexdewar, line 274: This is a bug. The reason this compiles is because the ``USE_VICON`` option is not enabled, so the compiler never actually runs into this code. What you need to do is build like so:
```
cmake -DUSE_VICON=ON ..
# ... etc.
```

The problem is that you're passing a ``std::array<degree_t, 3>`` and the compiler needs to be able to convert this to a string with ``std::to_string``, which it won't be able to. This should be:
```c++
imuData[0].value(), imuData[1].value(), imuData[2].value()
```

(Don't forget the ``.value()`` or you'll end up with something like ``1 deg, 2 deg, 0.5 deg`` written into your CSV file.)

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -259,16 +261,18 @@ class RobotFSM : FSM<State>::StateHandler
                     }
 
                     const double elapsed = static_cast<millisecond_t>(m_RecordingStopwatch.elapsed()).value();
-
                     // If Vicon tracking is available
                     if(m_Config.shouldUseViconTracking()) {
 #ifdef USE_VICON
                         // Get tracking data
                         const auto objectData = m_ViconTracking.getObjectData(m_Config.getViconTrackingObjectName());
+                        const auto imuData = imu.getEulerAngles();
```
alexdewar, line 269: It would be clearer if this was ``const std::array<degree_t, 3> imuData = ...``. You want to be explicit about the fact that these values are degrees so we know that we're not accidentally saving values in radians.

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -459,6 +463,9 @@ class RobotFSM : FSM<State>::StateHandler
         // Also log Vicon frame number, if present
         fieldNames.emplace_back("Frame");
 
+        // Also log the imu data
+        fieldNames.emplace_back("IMU [°]");
```
alexdewar, line 467: This needs to be broken into three separate fields (see above). I'd go with something like ``IMU yaw [degrees]..`` etc. without the unicode.

projects/snapshot_bot/snapshot_bot.cc
```diff
@@ -541,6 +548,7 @@ class RobotFSM : FSM<State>::StateHandler
 #ifdef USE_VICON
     // Vicon tracking interface
     Vicon::UDPClient<Vicon::ObjectData> m_ViconTracking;
+    BN005 imu;
```
alexdewar, line 551: Atm it'll only record IMU data if ``USE_VICON`` is enabled, which is not a biggie, but I don't think we should have this restriction.

More importantly though, we now won't be able to use ``snapshot_bot`` *without* the IMU -- if it's not available then we'll just get an error. The way to get round this is to make it a (smart) pointer, e.g. ``std::unique_ptr<BN003> m_IMU;``. Then if the user *actually* wants to use the IMU (i.e. if they've set a value in the config file), you can create the object later, with something like:
```c++
if (m_Config.shouldUseIMU()) {
    m_IMU = std::make_unique<BN055>();
}
```

You should definitely fix the other (compile-time) issues first though, then you can make this change after.

alexdewar (changes requested):
This actually has some compile-time errors in, but you need to build the code with ``-DUSE_VICON=ON`` to see them.

projects/snapshot_bot/snapshot_bot.cc
FullAPVayne: Hey Alex, 
I adressed all the other things, but this part does not compile. 
No matter if I do 

`BN005 imu;`
or
`m_IMU = std::make_unique<BN055>();` 

I get a "does not name a type error". Is there some header file in which I should forward declare?
Did this compile for you?

alexdewar: It should be ``BN055 imu;`` I think (you've got the class name wrong).

Also, you shouldn't do the ``make_unique`` stuff where you're declaring the member variables! The whole point is that you don't necessarily want to assign a value to it, in the case that you're not using the IMU. (It's basically the equivalent of assigning ``None`` to a variable in Python.) You'd declare it as something like ``std::unique_ptr<BN055> m_IMU;`` and you later (optionally) assign to it (with something like the snippet I sent you) in the constructor depending on whether or not it's needed.

PR #333, opened 2022-06-06
Roboclaw

neworderofjamie:
Fancy a last review of some code? I promise I made it extra fun by gratuitous use of variadic templates....

This is the code to drive the big new outdoor robot as a BoB robotics tank. I have only tested with a single motor connected to the motor controller so may push some fixes once we have the whole thing assembled. Also, as the ``Command`` enum indicates, you _can_ do a lot more fancy stuff with this motor controller than I have currently implemented but the variadic template glory of ``readCommand`` and ``writeCommand`` should make it very easy.

### DIFF
--- a/include/common/robo_claw.h
+++ b/include/common/robo_claw.h
@@ -14,10 +14,22 @@ namespace BoBRobotics {
 class RoboClaw
 {
 public:
-    Roboclaw(const char *path = SerialInterface::DefaultLinuxDevicePath, uint8_t address = 0x80, int maxRetry = 2);
+    RoboClaw(const char *path = SerialInterface::DefaultLinuxDevicePath, uint8_t address = 0x80, int maxRetry = 2);
 
+    //! Set motor speed
     void setMotor1Speed(float throttle);
     void setMotor2Speed(float throttle);
+
+    //! Get encoder values for motors
+    uint32_t getMotor1Encoder();
+    uint32_t getMotor2Encoder();
+
+    //! Get motor speeds
+    uint32_t getMotor1Speed();
+    uint32_t getMotor2Speed();
+
+    //! Get RoboClaw version string
+    std::string getVersion();
 
 private:
     //------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


include/common/robo_claw.h
```diff
@@ -0,0 +1,383 @@
...
+
+    // Perform a read command consisting of:
+    // write address [1 byte]
+    // write command[1 byte]
+    // read variadic parameters
+    // read CRC [2 bytes]
+    template<typename ...Data>
+    void readCommand(Command command, Data... data)
+    {
+        for(int r = 0; r < m_MaxRetry; r++) {
```
alexdewar, line 349: It looks like this loop is redundant: either you ``return`` or an exception is thrown.

alexdewar: I suppose you could try again if the CRC check fails... Not sure if that would make sense though

include/robots/tank/atv.h
```diff
@@ -0,0 +1,58 @@
...
+namespace Tank {
+//----------------------------------------------------------------------------
+// BoBRobotics::Robots::Tank::ATV
+//----------------------------------------------------------------------------
+//! An interface for large wheeled, all-terrain robot with two RoboClaw motor controllers
+class ATV
+  : public TankBase<ATV>
+{
+public:
+    ATV(const char *frontPath = "/dev/ttyACM0", const char *rearPath = "/dev/ttyACM1",
```
alexdewar, line 22: You could use persistent device names for these values, in case they get enumerated in a different order (or there's another serial device plugged in). You should be able to find them under ``/dev/serial/``.

src/common/robo_claw.cc
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    throw std::runtime_error("Unable to read to serial port");
+}
+//----------------------------------------------------------------------------
+float RoboClaw::getBatteryVoltage()
+{
+    uint16_t voltage;
+    readCommand(Command::GETMBATT, &voltage);
+    return (voltage * 0.1f);
```
alexdewar, line 126: There is the ``volt_t`` type, but that's probably not necessary

src/robots/tank/atv.cc
```diff
@@ -0,0 +1,37 @@
...
+#include "robots/tank/atv.h"
+
+namespace BoBRobotics {
+namespace Robots {
+namespace Tank {
+
+ATV::ATV(const char *frontPath, const char *rearPath, uint8_t frontAddress, uint8_t rearAddress)
+:   m_FrontController(frontPath, frontAddress), m_RearController(rearPath, rearAddress)
+{
+    // Sometimes Norbots get stuck driving, so let's stop it if we need to
```
alexdewar, line 13: This isn't a Norbot, right? :wink:

alexdewar (commented):
A few queries but otherwise looks good! Can't wait to see it in action :smile:

Shall we leave the PR open until we've tested it a bit more thoroughly on the finished robot?

include/robots/tank/atv.h
neworderofjamie: ooh, I never knew about that mechanism. Once I get both motor controllers plugged into something I'll investigate - **hopefully** they have different serial numbers or some way to distinguish them.

include/common/robo_claw.h
neworderofjamie: This is an excellent point! The retry logic was copied from the original arduino code but clearly doesn't currently quite make sense - with USB serial (rather than two wires jammed into a UART) I suspect it's entirely unnecessary but will check exactly when it was supposed to retry

neworderofjamie:
The code definitely works on the robot - will fix the dubious retry logic and maybe we can merge tomorrow?

alexdewar:
Tagging @shepai so he's in the loop.

src/common/robo_claw.cc
neworderofjamie: also those units are disabled by default so it would be kinda awkward

Issue #332, opened 2022-05-24
Fix build on Windows


alexdewar:
I fear the range library may not work properly on MSVC :cry:

neworderofjamie:
Do you want me to test? I have VS2019 on my machine at home which is supposed to be C++17 compliant

alexdewar:
Thanks, but I have been trying it myself with VS2019 and it's not working :cry:. Even setting the C++ standard to C++17 doesn't fix it (although it does actually work with C++11 on gcc).

I should probably open an issue. I don't think I'm the only one having problems though: https://github.com/ericniebler/range-v3/issues?q=is%3Aissue+is%3Aopen+msvc

alexdewar:
It's kinda weird really. I had it in mind the ``ranges-v3`` library was a pretty standard one.

neworderofjamie:
The docs say it needs:
> /permissive- and either /std:c++latest or /std:c++17

alexdewar:
Sorted!

PR #345, opened 2022-06-20
Fixes for various Windows/MSVC issues

alexdewar:
The build should now be working again. Regrettably I've had to bump the C++ standard to C++20 on MSVC (which is experimental and incomplete) so that the ranges library actually works.


alexdewar:
Fixes #332

Issue #349, opened 2022-06-23
Remove ranges-v3 library


alexdewar:
Unfortunately this has been causing build issues on Windows and Ubuntu 16.04 (which alas many of the Jetsons are running). While in theory it should work with any C++11 compliant compiler that seems not to be the case and for MSVC I had to enable experimental C++20 features just to get it to work. With gcc v5 (the default compiler on Ubuntu 16.04) it doesn't work and I haven't figured out why.

I'm not using it in many places and it seems like it isn't worth the hassle.

PR #350, opened 2022-06-24
Remove ranges from core code

alexdewar:
It turns out that the ranges library is only *really* being used in the Python module. There was one trivial usage in the ``navigation`` module but it was easy to factor out. This means that ``snapshot_bot`` will once again happily build on Ubuntu 16.04!

In related news, I summarily failed in my efforts to reflash the Jetson so it's still on Ubuntu 16.04. It seems less of a priority to sort that now. Once we connect up the IMU again we should be good to go!

Tagging @neworderofjamie


alexdewar:
Let's get this tested and merged along with #351 and #339 today!

alexdewar:
Relates to #349

PR #355, opened 2023-08-29
Infomax tweaks curtesy of Amany

neworderofjamie:
* Added comments to help future-self
* Now scale weight update by ``N*M`` not just N (pretty sure previously the code actually used ``M``)
* Allow number of hidden units to be specified separately
* Removed unnecessary ``array()`` and ``matrix()`` calls: + and - are always element-wise operations - @alexdewar is there a reason for these I'm not just getting?
* Removed tanh scaling - @amanyazevedoamin tells me it was a detour enroute to discovering the actual problem (updates should scaled as described above)

@amanyazevedoamin  - is there a more sensible default learning rate we should now use?
@alexdewar - after counteracting the learning rate scale, output of ``infomax_matlab_comparison`` is the same as before but differences are non-zero - what's going on here?

### DIFF
--- a/src/navigation/image_database.cc
+++ b/src/navigation/image_database.cc
@@ -351,7 +351,7 @@ ImageDatabase::loadCSV()
             extraFieldNames.emplace(i, std::move(fields[i]));
         }
     }
-
+    
     // Sanity check the file: we need the first four columns
     const auto validIdx = [](int idx) {
         return idx != -1;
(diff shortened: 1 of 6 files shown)


src/navigation/image_database.cc
```diff
@@ -363,15 +363,15 @@ ImageDatabase::loadCSV()
      * make sense for video-type ones.
      */
     BOB_ASSERT(validIdx(fieldNameIdx[4]) == !isVideoType());
-
+    
     // Assume it's a grid if we have any of the Grid fields...
     if (!hasMetadata()) {
-        m_IsRoute = !std::any_of(fieldNameIdx.cbegin() + 5, fieldNameIdx.cend(), validIdx);
+        m_IsRoute = !std::any_of(fieldNameIdx.cbegin() + 5, fieldNameIdx.cbegin() + 8, validIdx);
```
alexdewar, line 369: What's this change needed for?

alexdewar (approved):
Again, this looks fine to me and I'm not sure if I'm going to be much use :cry:.

So I think the calls to `array()` were just a mistake and you're safe to drop them.

I'm not sure why the output from the Matlab code is now different. With your tweak to the learning rate, there shouldn't be any semantic changes introduced (right?). How big are the differences? I seem to remember I had to set the tolerance of the floating-point comparisons to be fairly large for it to work. The reason for having the comparison to the Matlab code was because we were all confident that it had been done correctly (by Bart Baddeley, back in the day). That said, I don't think we need to keep it around forever if we're confident of the C++ implementation, though obviously the fact that the test has suddenly started failing is a bit concerning! Maybe ask Andy to look through it if you're concerned?

src/navigation/image_database.cc
neworderofjamie: It's a bit arbitrary really but pitch and roll data isn't really grid-specific

PR #356, opened 2023-08-31
Robot dead reckoning

neworderofjamie:
In order to implement robot-agnostic code to e.g. determine if the robot is moving or not, we want access to forward speed and turn speed. This should be pretty simple:

- Track the raw motor values in ``Omni2DBase`` and ``TankBase`` in the ``tank()`` and ``omni2D()`` methods
- Provide functionality to turn this into standard form of forward and turn speed (for omni this is trivial!) in same way that ``moveForward``, ``turnOnTheSpot`` turn standard commands into robot-specific ones

Some of this functionality existed in the past but was removed (@alexdewar any idea why? Tracking raw motor values  seems useful and exactly the type of thing the robot abstraction should provide) and, additionally, switching from a standard class hierarchy of robot classes to curiously-recurring template pattern makes it irritating as I have needed to liberally scatter ``friend`` declarations around (again, can't really remember why this change was made so worried I'm missing something here @alexdewar).

### DIFF
--- a/include/robots/omni2d/mecanum.h
+++ b/include/robots/omni2d/mecanum.h
@@ -19,17 +19,18 @@ namespace Omni2D {
 // BoBRobotics::Robots::Omni2D::Mecanum
 //----------------------------------------------------------------------------
 class Mecanum
-  : public Omni2DBase<Mecanum>
-{
+  : public Omni2DBase<Mecanum>{
+    friend Omni2DBase<Mecanum>;
 public:
     Mecanum(bool alternativeWiring = true,
             const char *path = SerialInterface::DefaultLinuxDevicePath);
     ~Mecanum();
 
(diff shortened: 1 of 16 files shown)


alexdewar (approved):
It all looks sensible to me.

Honestly, I can't really remember why I removed some of this functionality, but I suspect I was just having a clean-up and got rid of some unused functions. (It seems like a sensible thing to have though!)

Re the CRTP, I also can't really remember what the thinking was :wink:, but I have a feeling that it was nothing particularly profound. One advantage of using the CRTP cf. `virtual` methods is that you can have functions with the same name, but subtly different signatures, and I suspect I wanted to use that for something at some point. It's doubtful whether it makes the code better though and I think it's fine to remove it.

PR #357, opened 2023-09-25
Parallelism issues in ``MemoryBase::trainRoute`` 

neworderofjamie:
The snapshot bot encapsulates a bunch of image preprocessing logic in ``ImageInput`` classes. These are **not** thread-safe e.g. ``m_Unwrapped`` at https://github.com/BrainsOnBoard/bob_robotics/blob/snapshot_bot_load_fix/projects/snapshot_bot/image_input.cc#L71 is reused to reduce runtime memory allocations. However, they were being applied during parallelized loading which caused very weird image corruption issues! Making these thread-safe would make this code very messy so I'm just running these serially.

### DIFF
--- a/projects/snapshot_bot/memory.cc
+++ b/projects/snapshot_bot/memory.cc
@@ -57,31 +57,31 @@ void MemoryBase::setCSVFieldValues(std::unordered_map<std::string, std::string> 
     fields[CSVBestHeading] = std::to_string(getBestHeading().value());
     fields[CSVLowestDifference] = std::to_string(getLowestDifference());
 }
-
+//------------------------------------------------------------------------
 void MemoryBase::trainRoute(const Navigation::ImageDatabase &route,
                             ImageInput &imageInput,
                             size_t testFrameSkip,
                             BackgroundExceptionCatcher *backgroundEx)
 {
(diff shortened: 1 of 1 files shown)


alexdewar (approved):
Presumably, this was my doing. Sorry!

This looks sensible to me. An alternative would be to declare the relevant scratch variables as `static thread_local`, but as you say it would make things uglier (as well as harder to debug).

projects/snapshot_bot/memory.cc
```diff
@@ -92,7 +92,11 @@ void MemoryBase::trainRoute(const Navigation::ImageDatabase &route,
                 backgroundEx->check();
             }
 
-            train(snapshot.first, snapshot.second);
+            // Process snapshot
+            // **NOTE** ImageInput classes are NOT THREAD-SAFE so this needs to be done in serial
+            const auto processedSnapshot = imageInput.processSnapshot(snapshot);
+
+            train(processedSnapshot.first, processedSnapshot.second);
```
alexdewar, line 99: If you're using C++17 now, this could be:

```suggestion
            const auto [snapshot, mask] = imageInput.processSnapshot(snapshot);

            train(snapshot, mask);
```

Issue #353, opened 2022-07-04
Fix Python packaging issues


alexdewar:
As pointed out by @neworderofjamie, the folder layout for the Python module is
rather unorthodox and probably broken. Let's fix it up so e.g. "python setup.py
develop" etc. work properly.

PR #358, opened 2023-11-16
Fixed typo in algos and corrected name of dependency

neworderofjamie:
I'm afraid I still can't actually get this to work but these two issues definitely didn't help! @alexdewar when I build the old fashioned way e.g.
```
python antworld/setup.py develop
```
from ``python/`` it _appears_ to work but it looks like the egg-link is screwed and Python can't figure out how to map the ``bob_robotics`` python module to the  ``python`` directory and thus the two packages within  (I'm not really sure how the bob_robotics "namespace" thing is supposed to work really - I've never encountered a python package that works like this). The modern way:
```
pip install --editable antworld/
```
Fails with ``ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/its/home/jk421/bob_robotics/python/bob_robotics/antworld'`` which also suggest the installation is doing something really screwy. How **ARE** you supposed to install this thing? Clearly this is related to #353 but, unless you've got a fix knocking around, the only fix I would be confident with would be to turn this into a single standard ``bob_robotics`` package containing ``navigation`` and ``antworld``

### DIFF
--- a/python/navigation/src/algos.cc
+++ b/python/navigation/src/algos.cc
@@ -395,7 +395,6 @@ addAlgorithmClasses(py::module &m)
                         return PyAlgoWrapper<InfoMaxType>{
                             state[0].cast<cv::Size>(),
                             state[1].cast<float>(),
-                            state[2].cast<float>(),
                             state[3].cast<Normalisation>(),
                             nullopt,
                             state[4].cast<InfoMaxType::MatrixType>()
--- a/python/antworld/pyproject.toml
+++ b/python/antworld/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel", "scikit-build", "cmake", "cv2", "numpy"]
+requires = ["setuptools", "wheel", "scikit-build", "cmake", "opencv-python", "numpy"]


python/navigation/src/algos.cc
```diff
@@ -395,7 +395,6 @@ addAlgorithmClasses(py::module &m)
                         return PyAlgoWrapper<InfoMaxType>{
                             state[0].cast<cv::Size>(),
                             state[1].cast<float>(),
-                            state[2].cast<float>(),
```
alexdewar, line 398: What's this fix for?

alexdewar (approved):
What's the fix in `algos.cc` about? Assuming it makes sense and doesn't break anything, this is fine by me.

I was able to build and install the `antworld` package ok (after adding a missing `#include`) with `python setup.py bdist_wheel`. Honestly though I'm still not particularly au fait with the whole Python packaging thing (despite it sort of being part of my job :stuck_out_tongue:)... What does `pip install --editable` do?

PS -- I can't actually get the `navigation` module to build at all atm because of various minor issues :cry:, but I'll have a look at this and open up a PR.

PPS -- Are you still running CI for this on Jenkins btw? Would it be helpful if I wrote a GitHub Actions workflow for it instead? Assuming this code will still need minor fixes over the next few years, it might make things easier.

python/navigation/src/algos.cc
neworderofjamie: The algos fix is just a place I forgot to remove the tanh scaling from before. Editable mode is maybe not super helpful in this context as there's not much/no python code but I generally use it for packages I'm developing/have checked out of git and want to version that way - Like ``python setup.py develop`` before it, it just sticks a link into your python site-packages folder rather than making a copy.

neworderofjamie: Theoretically Jenkins is still running on BoB robotics - it doesn't look to be producing wheels though! I haven't dug deeper for a while but replacing Jenkins with github actions and local runners is on the near-future todo list after I finish the current round of GeNN development.

alexdewar: I'm happy to have a go -- we've got similar workflows to copy from. For non-CUDA-dependent bits, we could just run them on the default runners. They'll be free because BoB robotics is open source.

PR #360, opened 2023-12-11
Various build fixes

alexdewar:
This is based on #358, so that should be merged first.

Here are a smattering of random build fixes:

- Add some missing `#include`s
- Remove use of `tanhScalingFactor` in a few more places
- Bump versions of some submodules that weren't building for me

Various unit tests for infomax things are failing for me now, but I'm guessing that's because the data hasn't been regenerated since the algorithm was modified.


neworderofjamie:
Thank you so much for this! I'd like to try out building the antworld on a couple of machine myself so will get back to you next week as the next few days are going to be a nightmare

alexdewar:
Nw -- there's no rush.

neworderofjamie:
I'm now having a go an building Python antworld and your fixes seem to have basically worked on both Windows and Linux. Only outstanding issue is:

- Following the instructions and running ``python setup.py install`` in ``python\antworld`` fails because build_common can't be found (not clear to me how this could ever work without ``PYTHONPATH`` hacking)
- If I try and run ``python antworld\setup.py install`` in ``python``, it succeeds on Linux but you end up with a broken module and on Windows it fails with an error about Eigen which I suspect is merely a symptom of something else
- If I _copy_ ``build_common.py`` into antworld it works (

Any thoughts on the best solution to this? As adding the git hash to the version does nothing to actually help with semantically versioning the Python module, personally I'd just get rid of it but hey 😄

neworderofjamie (approved):

alexdewar:
Honestly, I'm not sure the best way to approach this. I basically hacked this in the first place so it's not surprising that it doesn't entirely work... I think I had some trouble trying to incorporate both Python and C++ code together using `scikit-build`, but that seems like a v common use case, so I'm sure there's a better way to do it than I did.

I think that realistically including the git hash in the version number is pretty pointless at this stage, given that the code base isn't changing a whole lot, so I'm def in favour of removing it if it makes things actually build properly.

neworderofjamie:
Hah, I thought that was the whole purpose of scikit-build 😄 I think the working directory thing may be a wider setuptools issue as python-only packages also seem to get broken if you run the install from outside of the package directory. I'll get rid of build_common and the versioning and merge this though - thank you again for your help with this!

neworderofjamie:
a python antworld wheel has been successfully installed on a windows machine not belonging to me! I think this can be counted as a success

PR #363, opened 2024-06-28
Remote joystick

neworderofjamie:
We have been previously using remote tank and friends quite happily for sending pure tank steering commands over the network but, because our ATV robot is encased in a big metal box and tends to outrun the xbox controller we need to be able to send button presses etc as well as do different things with input in different states.  So...in this PR I have implemented network sinks for joystick events (HID devices in BoB parlance).  I have made a ``remote_joystick`` 'tool' for running on a laptop and an example which polls the joystick and prints out events.

This is the deepest I've delved in BoB robotics for a while so any thoughts would be much appreciated! The only real gotcha is that, due to the way the joystick code is structured, the source and sink need to be running on the same OS (or at least their windowsness needs to match) but I don't think that's much of an issue. We could do something smarter with the initial state but I also think this is fine for now.

### DIFF
--- a/src/hid/net/source.cc
+++ b/src/hid/net/source.cc
@@ -2,6 +2,9 @@
 #include "hid/net/source.h"
 #include "net/connection.h"
 
+// Third party includes
+#include "plog/Log.h"
+
 namespace BoBRobotics {
 namespace HID {
 namespace Net {
@@ -12,6 +15,10 @@ namespace Net {
 Source::Source(BoBRobotics::Net::Connection &connection)
 :   m_Connection(connection)
 {
+#ifndef __linux__
+    LOGW << "Net joysticks require the same OS to be running the sink and source - be careful!";
+#endif
+
     // Set initial state of all axes to 0.0f
     for(size_t a = 0; a < static_cast<size_t>(JAxis::LENGTH); a++) {
         setState(static_cast<JAxis>(a), 0.0f, true);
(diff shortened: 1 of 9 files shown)


src/hid/net/source.cc
```diff
@@ -0,0 +1,94 @@
...
+
+                } else {
+                    setReleased(e.button.button, false);
+                }
+            }
+        }
+
+        m_Events.clear();
+        return true;
+    }
```
alexdewar, line 90: Just a stylistic thing, but personally I'd drop the `else`:

```suggestion

    for(const auto &e : m_Events) {
        if (e.axisNotButton) {
            setState(e.axis.axis, e.axis.value, false);
        } else {
            if (e.button.pressed) {
                setPressed(e.button.button, false);

            } else {
                setReleased(e.button.button, false);
            }
        }
    }

    m_Events.clear();
    return true;
```

alexdewar (approved):
I've suggested a few small stylistic things, but this looks sensible to me. Not sure why we never did this sooner actually... And it was quite nice to look at some BoB robotics code again and go "hey, these interfaces we came up with a few years ago are actually all right"!

Re the whole "host OS must match client OS" thing: that's obvs fine if it's not going to cause issues, but you *could* at least identify when they don't match (e.g. send a `JOY_OS` command, with the argument identifying the OS) so that you could give users an informative error message. Otherwise I can imagine that this might bite someone down the line. Up to you whether you think it's worth the effort or not.

neworderofjamie:
Thanks for the review, both suggestions are definitely stylistic improvements. I think I might just add a warning if not linux for now

neworderofjamie:
Also, clearly you are also more used to the sunlit uplands of C++17. Seems like, prior to that, ``std::vector::emplace_back`` returns void

alexdewar:
> Also, clearly you are also more used to the sunlit uplands of C++17. Seems like, prior to that, `std::vector::emplace_back` returns void

Haha. Oops!

